PR #105, opened 2015-01-14
Make devel vms install katello-devel-installer

eLobato:
Currently ```vagrant up centos7-devel``` installs katello-installer and fails when trying to provision by running katello-devel-installer


ehelms:
ACK

PR #109, opened 2015-01-19
attempt to use PV images

jlsherrill:
Which is actually what foreman-bats now does:

https://github.com/theforeman/foreman-bats/commit/bec1be654605ee7a3d8eb8908321802d22cc1777#diff-23b6f443c01ea2efcb4f36eedfea9089R35

Passing build when pointed to my git repo:  http://ci.theforeman.org/job/systest_katello/578/console

Will update the config once this is merged.


ehelms:
ACK

PR #111, opened 2015-01-21
Fix formatting of README

jmrodri:
The README was oddly formatted after the Plugins section. Adding a ``` seems to fix it.


mccun934:
ACK

PR #114, opened 2015-01-23
display the bash commands that we're running

komidore64:


dustints:
code looks good. :+1: 
going to run a quick test.

dustints:
ACK. tested looks good...log output: https://gist.github.com/1223f595a516ff50f397

setup.rb
```diff
@@ -12,6 +12,17 @@
   "2.1" => "releases/1.7"
 }
 
+# outputs system command as well as running the command
+module Kernel
+  def system_with_debug(*args)
+    puts "$ #{args.join(" ")}"
```
ehelms: What do you think about something more explicit (and less confusing with something manually from from the command line) such as "Running step: "

komidore64: `Running:` is cool. i'm down

komidore64: addressed

komidore64:
@ehelms ?

ehelms:
Looks good - only thing I might suggest is outputing a blank line before each command to make it easier to visually find where a new command is run. ACK either way.

komidore64:
easy enough. i'm not against that.

komidore64:
@ehelms updated

ehelms:
ACK

PR #115, opened 2015-01-28
Add parameter to allow testing module pull requests.

ehelms:
This adds the ability to specify any number of module/pull request
combinations that are then pulled into a local copy of the installer.
This can be used to more easily test someones change(s) to a given
module or set of modules without having to go through a clone or patch
workflow.


ehelms:
I debated supporting 'puppet-qpid' for example, and using the modules/ directory and Puppetfile to also parse that since, when looking at github, it might be natural to assume thats the name to use. And/or supporting full urls depending on what others did or thought might be the most useful.

PR #121, opened 2015-02-19
Remove branch parameter from bootstrap bats

eLobato:
There is no branch or tag bats in the repository so using this parameter in git clone would normally result in a `fatal: Remote branch bats not found in upstream origin`


jlsherrill:
ACK, thanks @eLobato

PR #125, opened 2015-02-26
Enable bats without Vagrant

eLobato:
katello-bats currently depends on Vagrant. If you run it on a bare-metal host, even though it tries to wget setup.rb, it will fail with the following error as it cannot load any of the requires:

` /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:53:in require': cannot load such file -- ./helper (LoadError) `

Instead I'm getting the whole repo and unzipping it, so that it's able to run in non-vagrant machines, at least without having to read the source and create a /vagrant directory with the contents of katello-deploy :smiley_cat:

### DIFF
--- a/bats/fb-install-katello-devel.bats
+++ b/bats/fb-install-katello-devel.bats
@@ -57,7 +57,9 @@ setup() {
   if [ -e "/vagrant/setup.rb" ]; then
     cd /vagrant
   else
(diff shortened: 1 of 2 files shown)


dustints:
ack!... +1 from non vagrant user

ehelms:
ACK, thanks @eLobato !

PR #128, opened 2015-03-12
[DO NOT REVIEW] Fixes #9689 - Adding BATS tests for docker

daviddavis:


ehelms:
I know it says DNR, but I'll save you me posting this later, how about adding  a 'docker' flag for these tests as well? `https://github.com/Katello/katello-deploy/blob/master/bats/katello-bats`

ehelms:
@daviddavis whats the status of this?

daviddavis:
I probably won't be able to address this in the next couple weeks. I'll probably be opening a PR with docker + atomic BATS tests.

PR #131, opened 2015-03-31
Eliminate dots in hostname that don't separate the domains

iNecas:
I had issues with hostnames like katello-centos7-2.2.example.com when using
dnsmasq to resolve them: my guess is the .2. being understood as additional
domain. I've replaces the dot's in the box name with '-' to fix this issue.


ehelms:
Thought I had ack'd this already...ACK pending rebase :)

iNecas:
Rebased

PR #136, opened 2015-04-04
fix jenkins build that does not support insert_key

jlsherrill:
and seems to validate the ssh config

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,7 @@
 require "yaml"
 
 VAGRANTFILE_API_VERSION = "2"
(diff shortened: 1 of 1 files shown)


ehelms:
Weird - I tested this exact case locally without issue.

jlsherrill:
@ehelms maybe its using an even older version?  Or maybe something related to the rackspace plugin?  I'm not entirely sure, was just going by the error

ehelms:
ACK

jlsherrill:
reading:  https://github.com/edx/configuration/issues/1008  it does seem to be a very old version of vagrant that causes it.

PR #137, opened 2015-04-06
use foreman-release-scl for scl repo files

jlsherrill:
also remove 2.0 installs from setup.rb as it is broken
and cannot use foreman-release-scl


ehelms:
Might as well remove the 2.0 boxes while you are at it.

jlsherrill:
@ehelms done

ehelms:
ACK

PR #139, opened 2015-04-09
typo our to out

jmrodri:
The Custom Boxes section stated 'could wipe our your local changes.' I changed our to out so it actually makes sense.


ehelms:
ACK, thanks @jmrodri

PR #140, opened 2015-04-13
Remove plugins from koji repos configuration option.

ehelms:
The plugins repository is special in that it houses multiple versions
of plugins. The Koji version of the repository only contains the latest
built package which can lead to issues if a major version is generated.


jlsherrill:
ack

PR #141, opened 2015-04-13
Configure foreman-plugins manually if using koji repos from published re...

ehelms:
...pos.

Prior commit removed configuring plugins from Koji, however, in a normal
situation the foreman release RPM would setup the plugins repository.
Since we are configuring Foreman repos from Koji manually, we need to
still setup the plugins repository pointing at the published repository.


jlsherrill:
ACK, merging!

PR #145, opened 2015-04-22
updating to use 7.1 centos box

jlsherrill:


Vagrantfile
```diff
@@ -18,12 +18,12 @@ module KatelloDeploy
       :libvirt    => 'http://m0dlx.com/files/foreman/boxes/centos64.box'
     },
     :centos7 => {
-      :box_name   => 'centos7',
+      :box_name   => 'centos7_1',
```
ehelms: We labeled the centos6 box as just '6', but I guess we have to do this for the new physical box?

jlsherrill: i wanted to change the name in order to force everyone to download it, especially anyone that may have gotten the 'bad' 7.0 box we had been pointing to.  We could provide steps to remove it i guess instead?

ehelms: I guess I am mostly using boxes.yaml these days anyway, so, I think the forcing people to download it is better than trying to keep the box_name consistent at all times.

ehelms:
ACK

PR #162, opened 2015-06-16
Exit 1 if installer fails

ehelms:


lib/katello_deploy/installer.rb
```diff
@@ -38,27 +38,29 @@ def install_packages(packages)
     def run_installer(command)
       if @skip_installer
         warn "WARNING: Skipping installer command: #{command}"
-        return
+        return false
```
jlsherrill: should this return false?  seems to be if i told it not to run the installer it should be successful?

ehelms: Good point

jlsherrill:
APJ

PR #165, opened 2015-07-06
Box to deploy base Pulp and Runcible for test running or editing.

ehelms:


ehelms:
This is sort of crude compared to our other methods, but its a start and accomplishes both allowing spinning up an easy Runcible environment and testing it automatically as well as testing of the puppet-pulp module. This currently relies on master of puppet-pulp which has also found a few issues that need to be fixed there. Additions that crossed my mind:

1) Add this scenario to puppet-katello_devel instead
2) Allow passing the pulp version to this script

beav:
I tried this out and it worked, but I had to run `gem install bundler; bundle install` from the runcible dir to run the tests, even after the tests ran via the vagrant setup script. I am not sure if that is expected or not.

ehelms:
You likely just needed to activate and use the right ruby version and gemset with rvm. Other things I considered with this were letting the script handle pointing at a PR, as well as ways to configure which Pulp version it pointed at (for testing new Pulp versions). And potentially simplifying the script to make use of other puppet modules for some of the configuration. If all of that seems pre-mature, I am fine with that as well but I originally did the bare minimum to get some developer feedback :) I also considered doing this via Docker instead.

beav:
that was indeed the issue, I just had to do a `rvm gemset use runcible` to make it work.

looks good to me after rebase

ehelms:
rebased

beav:
@ehelms does it run ok for you? I get a strange error on qpid startup and was wondering if you got the same:

```
Oct 27 10:47:32 katello-runcible.example.com qpidd[30909]: /usr/sbin/qpidd: symbol lookup error: /lib64/libqpidbroker.so.2: undefined symbol: _ZN4qpid3sys4Path9separatorE
```

`rpm -V qpid-cpp-server-0.34-4.el7.x86_64` says that `libqpidbroker.so.2` looks ok.

ehelms:
That appears to be a qpid error coming from the epel qpid which is newer than what we have in our repositories. I am going to test enabling our Pulp repository to get that version of qpid.

ehelms:
Updated -- give it another try if you can. You will likely find that six tests are failing which are ostree related and require -- https://github.com/Katello/puppet-pulp/pull/85

beav:
```
==> runcible: 227 tests, 545 assertions, 0 failures, 6 errors, 3 skips
```

total runtime 18 min 42s

ACK, looks good :smile_cat:

beav:
also good catch on the qpid package :tropical_fish:

ehelms:
This is the most emoji's I have seen in a single PR! Thanks for the reviews @beav

PR #183, opened 2015-10-26
Refs #7990 - install client repo rpm in addition to server

jlsherrill:


jlsherrill:
Fixes Katello XL Pipeline leak

ehelms:
What caused this sudden breakage?

jlsherrill:
https://github.com/Katello/katello-deploy/pull/119 i assume.   the pipeline doesn't run over the weekend I don't think? so last night was the first run to catch it.

ehelms:
Ohhh, I was thinking something broke it vs something new being added that requires it: makes perfect sense. I would have thought https://github.com/Katello/katello-packaging/blob/master/repos/katello-repos.spec#L61 implies that both repo files are deployed when installing katello-repos ?

jlsherrill:
@ehelms hrmmmmm it should..... http://koji.katello.org/koji/rpminfo?rpmID=142267 

I'll investigate further

jlsherrill:
@ehelms updated after further investigation

ehelms:
ACK

PR #187, opened 2015-11-03
Refs #7990 - use longer timeout for package tests

jlsherrill:


stbenjam:
Is 120 enough? If its pulling all the centos and katello metadata too maybe we should disable all the repos except the ones in redhat.repo?

jlsherrill:
It should have that all downloaded and cached, but no harm in bumping to 5
minutes?

stbenjam:
Hmm, ok, I thought maybe the 120 was needed because the cache was cleared, 30 seconds isn't enough for the local repo? Slow box?

Anyway, ACK from me

mccun934:
ACK from me as well

jlsherrill:
@stbenjam my guess is that http://projects.theforeman.org/issues/12375  may be partly to blame.  Last night one of the 2 bats tests (el6) did pass, previously neither were passing so my guess is that its taking slightly longer than 30 seconds, and looking at the logs that seems like what is happening.    We can always bump again if needed.

PR #188, opened 2015-11-20
use rpm to install epel release

jlsherrill:
as that is what foreman-bats does and they have not seen failures on installation

### DIFF
--- a/lib/katello_deploy/repositories.rb
+++ b/lib/katello_deploy/repositories.rb
@@ -57,7 +57,7 @@ module KatelloDeploy
     end
 
     def bootstrap_epel(release)
-      local_install("http://dl.fedoraproject.org/pub/epel/epel-release-latest-#{release}.noarch.rpm")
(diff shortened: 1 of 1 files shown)


ehelms:
Do you want to temporarily point systest_katello at your fork and branch of katello-deploy to see if this fixes the issue?

jlsherrill:
@ehelms yeah i had contemplated that but figured i'd open the PR first, doing that now

jlsherrill:
test run:  http://ci.theforeman.org/view/Katello%20Pipeline/job/systest_katello/1466/console

jlsherrill:
@ehelms that was successful, which makes no sense, want me to try a 2nd time?

ehelms:
Isn't the point to be successful since its using this PR?

jlsherrill:
Yes that is the point, but it makes no sense :)

It actually passed once in the past ~3 weeks, so lets go with 'this fixes it'

ehelms:
ACK

PR #194, opened 2016-01-22
Fix bats tests for activation key create and host package install

ehelms:


ekohl:
ACK

ehelms:
Opened a PR to fix the failures before merging this (even though this is unaffected by it) -- https://github.com/Katello/katello-deploy/pull/195

PR #197, opened 2016-02-05
sort scripts before running them

komidore64:
This may not apply to everyone, but the scripts I wrote to be executed
after Katello finishes installing are to be run in order. For the most
part they seemed to go in order, but I upgraded rubies and that seems to
be less consistent now. I added an explicit call to sort.

```
# directory structure
./scripts
  10-packages.bash
  20-dotfiles.bash
  30-foretello.bash
  40-hammer.bash
```

### DIFF
--- a/lib/katello_deploy/processors/scripts_processor.rb
+++ b/lib/katello_deploy/processors/scripts_processor.rb
@@ -14,7 +14,7 @@ module KatelloDeploy
       def self.run_scripts
         scripts = Dir.glob('*').select { |e| File.file? e }
 
-        scripts.each do |script|
+        scripts.sort.each do |script|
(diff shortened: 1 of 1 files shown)


ehelms:
ACK - the failure seems unrelated and due to a Rubocop update

komidore64:
[test]

ehelms:
Travis doesn't listen to [test] :) And without adding it to this PR or another PR, the rubocop failures won't go away. But, I am OK with this getting merged despite that.

komidore64:
oh right, haha

ehelms:
ACK - merging this and will follow up with Rubocop fix

PR #201, opened 2016-02-10
only use box_check_update on newer vagrants

jlsherrill:


ehelms:
ACK - thanks for fixing this

PR #204, opened 2016-02-25
Use prefixed params for setting devel_user

mbacovsky:
With https://github.com/Katello/katello-installer/pull/303 merged the params in katello-devel scenario will be prefixed


ehelms:
Thanks @mbacovsky !

PR #207, opened 2016-03-16
Update for move to dist based repo naming

ehelms:


docker/clients/fedora20/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fedora:20
+FROM fedora:22
```
jlsherrill: unrelated changes?

jlsherrill: or just updating to supported fedora client?

jlsherrill:
Haven't tested, but looks good to me if the dockerfile change is intended.

docker/clients/fedora20/Dockerfile
ehelms: I had to update the repo location in the Dockerfile and realized they were still set to f20 and f21 which won't work for nightly so I updated them to reflect the repos that will actually exist.

jlsherrill:
ACK

PR #212, opened 2016-03-29
change bats content test to new hosts options

jlsherrill:


jlsherrill:
@zjhuntin does this look correct based on your recent changes?

ekohl:
Does this only work on development versions or also on existing releases? Is that even a relevant question for this repo?

ehelms:
Right now we only run our bats regularly for nightly tests, we have used it for pipeline testing releases before, and, at least for 3.0, this should work still.

jlsherrill:
That said we did branch this repo at katello-2.4, maybe we should start doing that for every release.

zjhuntin:
Yes, that does look correct!

PR #216, opened 2016-04-13
ignore .ruby-version and .ruby-gemset

komidore64:
If you use RVM then you might potentially need to set a ruby version and
gemset for your katello-deploy directory.

If the ``system`` ruby is not RVM's default ruby, then you might run
into some issues. It can help to specify a desired ruby version and
gemset for RVM to switch to whenever using katello-deploy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ plugins
 scripts
 repo
 Gemfile.lock
+.ruby-version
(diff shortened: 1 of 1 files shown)


ehelms:
ACK

komidore64:
plz2merge?

PR #218, opened 2016-04-27
Enable ostree for the runcible box

daviddavis:


parthaa:
ack

parthaa:
I tested this and it ran ok

ehelms:
thanks @daviddavis

PR #232, opened 2016-05-24
Tools cli

ehelms:
Something I had been pondering for a bit, and am curious others thoughts given what it introduces dependency. Sometimes I find I want some of the functionality built here without having to run through all of the installer aspects. For example, I want to be able to take a Koji task and turn it into a repository or turn a local directory into a repository easily. This idea introduces a CLI for common development/debugging tasks and tools that are not tied to running the installer.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,3 +9,4 @@ end
 
 gem 'librarian-puppet'
 gem 'puppet', '< 4.0.0'
+gem 'clamp'
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,5 +1,5 @@
 require 'yaml'
-require './lib/forklift'
+require './lib/forklift/box_loader'
 
 VAGRANTFILE_API_VERSION = '2'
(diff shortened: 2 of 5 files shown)


ehelms:
@stbenjam @ekohl would love to get your thoughts on this :)

ekohl:
I think this could be a good idea, but I wonder if setup.rb should also be merged into this. That would make it a unified tool.

stbenjam:
Yea, I like this, I agree it should be merged with setup.rb if possible.

stbenjam:
Re: the specific koji functionality, one thing that would be useful is giving it a PR instead.  When I submitted my PR to update rex plugins for example, it is a ton of different koji tasks, and it's just as cumbersome to use this as it is just to copy and paste from the Koji UI.

Although maybe the Jenkins jobs for packaging PR's should really do this, if we had enough space, to have it create a yum repo as an artifact.

stbenjam:
@ehelms Do we still need this, or should this kind of functionality be in ansible playbooks now?

ehelms:
We can close now.

PR #237, opened 2016-06-03
update how to start rails server

thomasmckay:
Specified path to rails as well as allowing server to be reached via http on port 3000

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ The box can now be accessed via ssh and the Rails server started directly (this 
     vagrant ssh <deployment>
     cd /home/vagrant/foreman
     sudo service iptables stop
-    rails s
+    bin/rails s -b 0.0.0.0
 
 ### Adding Custom Boxes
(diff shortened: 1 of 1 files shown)


jlsherrill:
Not opposed to this change, but any reason to not just hit  https://hostname/ ?

jlsherrill:
Also is rails not in your bin path ?    bin/rails

[vagrant@robot ~]$ which rails
~/.rvm/gems/ruby-2.2.4/bin/rails

thomasmckay:
If you don't think this is generically useful, I'm alright closing it. I run remote debugger and normally do so against :3000. I don't know if there are issues w/ https but I know I had problems in the distant past. I do not use rvm but yes rails is in my path; I just thought it was best practice to use path to rails.

jlsherrill:
ACK

thomasmckay:
I don't have merge permissions on this repo

PR #246, opened 2016-07-07
add 3.1 support

jlsherrill:


ehelms:
Also here -- https://github.com/Katello/forklift/blob/master/playbooks/roles/common/vars/main.yml#L3

jlsherrill:
updated

ehelms:
ACK

PR #254, opened 2016-07-21
Added centos7 atomic vagrant box

parthaa:


ehelms:
ACK - thanks @parthaa !

PR #256, opened 2016-07-22
Support multiple playbooks for provisioning

ehelms:


.gitignore
```diff
@@ -14,3 +14,4 @@ Gemfile.lock
 .ruby-gemset
 *.retry
 *.pyc
+user_playbooks
```
akofink: We should go ahead and create this directory either with a `.gitkeep` file or with an example playbook.

ehelms: Does gitkeep prevent you from committing anything you add to it? I want to help prevent devs accidentally committing their custom scripts.

akofink: It's just a way to commit an empty directory to a git repo. You could also ignore `user_playbooks/*.yml` and add a `user_playbooks/playbook1.yml.example`. I just don't want our users to have to create the `user_playbooks` folder.

ehelms:
Updated @akofink

Vagrantfile
```diff
@@ -64,11 +64,25 @@ module Forklift
...
+          if playbooks.is_a?(String)
+            machine.vm.provision 'main', type: 'ansible' do |ansible|
+              ansible.playbook = playbooks
+
+              ansible.groups = @ansible_groups
+            end
+
+          elsif playbooks.is_a?(Array)
+
+            playbooks.each do |playbook, index|
```
zjhuntin: I think this needs to be playbooks.each_with_index

akofink:
Very nice. ACK

ehelms:
@zjhuntin updated -- I think it was just working because I kept redefining the main provisioner, using an acutal index is probably safer in the long run.

zjhuntin:
Works great! ACK

PR #259, opened 2016-07-26
Add hammer development box via ansible

akofink:
This will be a standalone hammer development environment, which you can point to any Satellite/Katello remote.

I thought I would push what I have - it's working in a Vagrant box, but the ultimate goal is to get it running in a container with ansible-container (thus the WIP). I made some decisions along the way which you may disagree with (i.e. rbenv or where I've placed certain roles or named certain variables). Because it's my first ansible playbook, I am all ears for suggestions on improvement. :hammer:

### DIFF
--- a/config/base_boxes.yaml
+++ b/config/base_boxes.yaml
@@ -92,3 +92,10 @@ boxes:
   pipeline-capsule-3.1-centos7:
     box: centos7
     memory: 3072
+
+  hammer-devel-centos7:
+    box: centos7
+    memory: 512
+    ansible:
+      playbook: 'playbooks/hammer_devel.yml'
+      group: 'hammer_devel'
--- a/playbooks/hammer_devel.yml
+++ b/playbooks/hammer_devel.yml
@@ -1,0 +1,4 @@
+---
+- hosts: all
+  roles:
+    - hammer_devel
--- a/playbooks/roles/common/tasks/main.yml
+++ b/playbooks/roles/common/tasks/main.yml
@@ -1,2 +1,3 @@
 ---
 - include: repositories.yml
+  become: true
--- a/playbooks/roles/hammer_devel/defaults/main.yml
+++ b/playbooks/roles/hammer_devel/defaults/main.yml
@@ -1,0 +1,10 @@
+---
+hammer_devel_ruby_version: 2.3.1
+hammer_devel_host: 'http://katello-devel.example.com:3000/'
+hammer_devel_username: admin
+hammer_devel_password: changeme
+hammer_devel_repositories:
+  - theforeman/hammer-cli
+  - theforeman/hammer-cli-foreman
+  - theforeman/hammer-cli-foreman-tasks
+  - katello/hammer-cli-katello
--- a/playbooks/roles/hammer_devel/tasks/git.yml
+++ b/playbooks/roles/hammer_devel/tasks/git.yml
@@ -1,0 +1,4 @@
+---
+- name: 'Install git'
+  become: true
+  yum: name=git-core state=present
--- a/playbooks/roles/hammer_devel/tasks/hammer.yml
+++ b/playbooks/roles/hammer_devel/tasks/hammer.yml
@@ -1,0 +1,71 @@
+---
+- name: 'Clone the hammer repositories'
+  git:
+    repo: https://github.com/{{ item }}.git
+    dest: ~/{{ item.split('/')[1] }}
+    update: no
+  with_items: "{{ hammer_devel_repositories }}"
+
+- name: 'Add local gems to Gemfile.local'
+  blockinfile:
+    dest: ~/hammer-cli-katello/Gemfile.local
+    create: yes
+    block: |
+      gem 'hammer_cli', :path => '../hammer-cli'
+      gem 'hammer_cli_foreman', :path => '../hammer-cli-foreman'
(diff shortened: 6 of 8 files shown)


playbooks/hammer-devel.yml
```diff
@@ -0,0 +1,4 @@
+---
+- hosts: hammer-devel
```
ehelms: If you add 'all' here, it would allow me to do this against my own box however I want, e.g.

```
ansible-playbook -l myhost playbooks/hammer-devel.yml

**Comment by ehelms** [comment]
I would not put anything in 'common', in fact, I need to remove it entirely. This is generally a bad practice that I started us with. I'd encourage small, well defined roles that do specific things and then combine them in playbooks.

File: playbooks/roles/common/tasks/rbenv.yml [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+- name: 'Initialize rbenv on login'
+  lineinfile: dest=~/.bash_profile line="eval \"$(rbenv init -)\""
+- name: 'Reload the environment'
+  shell: 'source ~/.bash_profile'
+- name: 'Clone ruby-build'
+  git: repo=https://github.com/rbenv/ruby-build.git dest=~/.rbenv/plugins/ruby-build
+- name: 'Build and install ruby'
+  command: ~/.rbenv/bin/rbenv install {{ hammer_ruby_version }} creates=~/.rbenv/versions/{{ hammer_ruby_version }}
+- name: 'Set global ruby version'
+  command: ~/.rbenv/bin/rbenv global {{ hammer_ruby_version }}
```
ehelms: Spacing between each task would be nice for readability,

```
- name: foo
  command: stuff

- name: bar
  command: more stuff
```

playbooks/roles/hammer-devel/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+hammer_ruby_version: 2.3.1
+hammer_host: 'http://katello-devel.example.com:3000/'
+hammer_username: admin
+hammer_password: changeme
```
ehelms: Best practices encourage prefix with the role namespace, in other words, hammer_devel vs hammer

playbooks/roles/hammer-devel/tasks/hammer_cli.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: 'Clone the hammer-cli repository'
+  git: repo=https://github.com/theforeman/hammer-cli.git dest=~/hammer-cli update=no
```
ehelms: Nitpick/personal preference for this style syntax for readability and ease of editing:

```
git:
  repo: https://
  dest: ~/hammer-cli
  update: no
```

playbooks/roles/hammer-devel/tasks/hammer_cli_foreman_tasks.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: 'Clone the hammer-cli-foreman-tasks repository'
+  git: repo=https://github.com/theforeman/hammer-cli-foreman-tasks.git dest=~/hammer-cli-foreman-tasks update=no
```
ehelms: For all of these you might consider iterating:

```
- name: Clone repositories
  git:
    repo: https://github.com/{{ item }}.git
    dest: "{{ item.split('/')[1] }}"
    update: no
  with_items: "{{ hammer_devel_repositories }}"
```

in defaults.yaml

```
hammer_devel_repositories:
  - theforeman/hammer-cli-foreman-tasks
  - katello/hammer-cli-katello
```

playbooks/roles/hammer-devel/tasks/hammer_cli_katello.yml
```diff
@@ -0,0 +1,56 @@
...
+    dest: ~/hammer-cli-katello/Gemfile
+    regexp: "gem 'hammer_cli_foreman'"
+    line: ''
+- name: 'Remove hammer_cli in Gemfile'
+  lineinfile:
+    dest: ~/hammer-cli-katello/Gemfile
+    regexp: "gem 'hammer_cli'"
+    line: ''
+- name: 'Install bundler'
+  command: bash -lc 'gem install bundler'
```
ehelms: Could you use the gem module here? http://docs.ansible.com/ansible/gem_module.html

playbooks/roles/hammer-devel/tasks/hammer_cli_foreman_tasks.yml
akofink: Awesome! Learned something new. :mortar_board:

playbooks/roles/hammer-devel/tasks/hammer_cli_katello.yml
akofink: So... this is a pain point with rbenv. I had the gem module, and I specified the executable as `~/.rbenv/shims/gem`, but bundler still got installed system-wide rather than inside of `~/.rbenv`. The reason for this (I think) is because rbenv is a shell function and not an executable. It needs a shell environment to be configured properly. There might be a way around it, but I need to do some more research.

akofink:
FYI it works great with 512MB of ram.

akofink:
I'm good with this as a Vagrant box for now (removes WIP). I plan on making another PR to allow it to run in a container.

ehelms:
ACK - thanks @akofink !

PR #261, opened 2016-07-26
Allow external roles to be installed to galaxy_roles

akofink:
FYI: `ansible-galaxy` installs roles to the first role in your `roles_path`.

To test, try something like this:

```sh
ansible-galaxy install bennojoy.mysql
```

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -2,7 +2,7 @@
 callback_plugins = $PWD/playbooks/callback_plugins/
 nocows = 1
 inventory = $PWD/playbooks/inventory/vagrant.py
-roles_path = $PWD/playbooks/roles
+roles_path = $PWD/playbooks/galaxy_roles:$PWD/playbooks/roles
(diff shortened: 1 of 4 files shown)


ansible.cfg
```diff
@@ -2,7 +2,7 @@
 callback_plugins = $PWD/playbooks/callback_plugins/
 nocows = 1
 inventory = $PWD/playbooks/inventory/vagrant.py
-roles_path = $PWD/playbooks/roles
+roles_path = $PWD/galaxy_roles:$PWD/playbooks/roles
```
ehelms: I was thinking of putting this side-by-side for parity: `playbooks/galaxy_roles`

ekohl:
Would it be useful if this was added to the readme?

akofink:
I updated the Readme and also added a `requirements.yml.example` and ignored `requirements.yml` for use with `ansible-galaxy install -r requirements.yml`.

ehelms:
Thanks @akofink

PR #264, opened 2016-07-28
More generic capsule_ip

iNecas:
We can't rely on the eth0 if the playbook used outside
of the vagrant environment


ehelms:
Truth! ACK - thanks @iNecas !

PR #266, opened 2016-07-28
Fix error with devel install from erroneuous Rubocop fix

ehelms:


lib/forklift/installer.rb
```diff
@@ -96,7 +96,7 @@ def run_installer(command)
 
     def syscall(command)
       system(command)
-      $CHILD_STATUS.zero?
+      $CHILD_STATUS.to_i.zero?
```
daviddavis: Do you want this to return true when $CHILD_STATUS is nil? I ask because previously (`$CHILD_STATUS == 0`) it would return false for nil.

stbenjam: Why not just `$?.success?`

ehelms:
Updated based on @stbenjam suggestion

stbenjam:
Lol sorry, rubocop doesn't like it, but I think rubocop's suggestion is bad and maybe we can just turn it off.

```
1.8.7-head :007 > system('ls')
[...snip...]
1.8.7-head :008 > $CHILD_STATUS.success?
NoMethodError: undefined method `success?' for nil:NilClass
	from (irb):8
	from /usr/bin/irb:11:in `<main>'
1.8.7-head :009 > $?.success?
 => true 
```

ehelms:
Let's see if its happy now...

stbenjam:
ACK

daviddavis:
It's because you're not requiring 'English'. See here:

https://github.com/bbatsov/rubocop/issues/1747

```
[6] pry(main)> system('ls')
[...]
[7] pry(main)> $CHILD_STATUS.success?
=> true
```

stbenjam:
I still prefer `$?`

daviddavis:
I think `$?` is clear because I see it so much in the terminal. However, some of the others like `$'`, `$;`, etc, [I would have to look up](http://ruby-doc.org/stdlib-2.0.0/libdoc/English/rdoc/English.html). I wish rubocop would let you whitelist particular ones though like `$?`. I wouldn't mind disabling this rubocop.

ehelms:
Good to merge this as is? I'll most likely be turning a lot of this into
roles and playbooks so it shouldn't be around for too long.

daviddavis:
ACK from me.

PR #269, opened 2016-08-03
Experimental Foreman deployment with separate DB server

ehelms:
In my free time, I like to play around with different deployment ideas looking ahead to the future. Rather than have them live off in some repository of mine, I thought it would be interesting to create a section for "experiments" to show what might be possible, to move forward with changes to make them first class citizens, or to show off neat ideas. To that end, I am throwing out this idea of an experiments directory for housing these types of things so that they can be more broadly viewed, consumed or contributed to. Some are things I hope to push forward towards production level versions of in the future. 

The first of these is a simple deploy Foreman with a Postgres database into two separate VMs. This shows what it takes to automate this today, and some of the work arounds that are needed to use our existing installation procedures to pull this off.


config/base_boxes.yaml
```diff
@@ -99,3 +99,11 @@ boxes:
...
       playbook: 'playbooks/hammer_devel.yml'
       group: 'hammer_devel'
+
+  postgres:
+    box: centos7
+    memory: 1024
+
+  foreman:
+    box: centos7
+    memory: 1024
```
ehelms: Not a big fan of this, and thinking about better ways to handle it. I might switch to using something like [1] to create ones like these on the fly more.

[1] https://github.com/robparrott/ansible-vagrant

ekohl:
There's a merge conflict. Other than that, is this still experimental?

ehelms:
@ekohl I suppose that depends on how we want to view experimental. I was reticent to clog up the normal playbooks folder with deployments that are not typical but have some eye towards the future or different configurations of deployment.

PR #270, opened 2016-08-08
Allow custom user roles

akofink:
This really helps users perfect local roles without changing the git repo, then when ready, move them into forklift by simply moving them into the `playbooks/roles` directory and submitting a PR.


ehelms:
Thanks @akofink !

PR #271, opened 2016-08-10
add per-box disk size field

komidore64:
You can now include `disk_size` in `boxes.yaml` for your custom boxes.
Disk size is specified in gigabytes. This only sets the virtual disk size, so you will still need to resize partitions and filesystems manually.

Example:

```yaml
# boxes.yaml

devbox6:
  box: centos6-devel
  bridged: br0
  disk_size: 250
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ memory -- set the amount of memory (in megabytes) this box will consume
 cpus -- set the number of cpus this box will use
 hostname -- hostname to set on the box
 networks -- custom networks to use in addition to the management network
+disk_size -- specify the size (in gigabytes) of the box's virtual disk. This only sets the virtual disk size, so you will still need to resize partitions and filesystems manually.
 ```
 
 Entirely new boxes can be created that do not orginate from a box defined within the Vagrantfile. For example, if you had access to a RHEL Vagrant box:
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -40,7 +40,7 @@ module Forklift
         network['options'] = network['options'].inject({}){ |memo,(k,v)| memo.update(k.to_sym => v) }
       end
 
-      if box.key?('shell') && !box['shell'].nil? 
+      if box.key?('shell') && !box['shell'].nil?
         machine.vm.provision :shell do |shell|
           shell.inline = box.fetch('shell')
           shell.privileged = false if box.key?('privileged')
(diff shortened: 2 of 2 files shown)


ehelms:
ACK - thanks @komidore64 !

I wonder if we could add a simple playbook that runs on boxes and handles the partition resizing and such automatically when the box is spun up.

komidore64:
@ehelms I've found that a machine restart is needed after running `fdisk` on a live mounted partition. For whatever reason `partprobe` fails to notify the kernel of the new partition table on these vagrant boxes that I've tried. So as long as we can get around that, we should be good.

ehelms:
I swear I've done it without needing to restart anything. I think I followed the bottom section of https://github.com/mccun934/wiki/wiki/AddDiskToVMHowTo

komidore64:
That is adding an entirely new virtual disk though, which is different. The kernel should have very little trouble accepting a new disk and partition table on something that hasn't been mounted, so there is no chance of current read/writes.

ehelms:
Oh - right, this is just changing the base VM disk size. That seems like we could add https://github.com/vagrant-libvirt/vagrant-libvirt#additional-disks as an option and automate those steps to get 'bigger' VMs on demand as well.

swapnilabnave:
@komidore64 
> "resize partitions and filesystems manually."
Can you please share the steps needed to do it ? I am newbie to libvirt and disk resize.

komidore64:
@swapnilabnave sure! Use ``fdisk`` or ``parted`` to resize the partition on the VM's disk, after that you should restart the VM to ensure the kernel sees the newly updated partition table (``partprobe`` errors on me more often than not here so I just reboot), then resize on the VM's file system with ``xfs_growfs`` or ``resize2fs``.

Something similar to the following:

I added a new box to my ``boxes.yaml``

```yaml
disktest:
  box: centos7
  disk_size: 250
```

Once the VM is up and running with us ssh'd in we're going to resize its root partition with ``fdisk``.

```sh
$ sudo fdisk /dev/vda <<EOF
d
3
n
p
3


w
EOF
```

Reboot. Since this file system is ``ext4``, we'll use ``resize2fs``.

```sh
$ sudo resize2fs /dev/vda3
```

Hope this helps!

swapnilabnave:
Sweet..!!! :heart_eyes: 
```
[root@centos7-devel ~]# df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda3        89G  6.7G   78G   8% /
```

Thanks! @komidore64

komidore64:
Glad it worked for you :+1:

PR #272, opened 2016-08-10
Improve the hammer devel environment

akofink:
This change removes rbenv in favor of an RPM installation, greatly
reducing the amount of time and CPU resources required to up this box.

In addition, it allows the user to use the `hammer` executable from any
directory once ssh'd into the box.

Finally, the etc_hosts role is added to this playbook to ensure
whichever katello/satellite server the user wishes to use as the
hammer host is already defined in /etc/hosts


playbooks/roles/hammer_devel/defaults/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 hammer_devel_ruby_version: 2.3.1
-hammer_devel_host: 'http://katello-devel.example.com:3000/'
+hammer_devel_host: https://centos7-katello-3-0.example.com
```
ehelms: Changing the default to a 3.0 production box vs a more standard dev box?

playbooks/roles/hammer_devel/tasks/ruby-2-3-1.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: 'Remove Old Ruby'
+  become: true
+  yum: name=ruby,ruby-libs state=absent
+- name: 'Install Ruby'
+  become: true
+  yum: name=https://github.com/feedforce/ruby-rpm/releases/download/2.3.1/ruby-2.3.1-1.el7.centos.x86_64.rpm state=present
```
ehelms: If this is applied to an existing environment this could have unintended consequences for the developer. I get how this makes things easier on a stand-alone deployment. I just wonder about invasive changes like this vs using RVM or even trying to use software collections if RPMs is the goal [1].

[1] https://www.softwarecollections.org/en/scls/rhscl/rh-ruby23/

akofink:
Updated, @ehelms

playbooks/roles/hammer_devel/tasks/hammer_install.yml
```diff
@@ -0,0 +1,38 @@
...
+- name: 'Add local gems to Gemfile.local'
+  blockinfile:
+    dest: ~/hammer-cli-katello/Gemfile.local
+    create: yes
+    block: |
+      gem 'hammer_cli', :path => '../hammer-cli'
+      gem 'hammer_cli_foreman', :path => '../hammer-cli-foreman'
+      gem 'hammer_cli_foreman_tasks', :path => '../hammer-cli-foreman-tasks'
+
+- name: 'Remove hammer_cli_foreman in Gemfile'
```
daviddavis: With https://github.com/Katello/hammer-cli-katello/pull/442, you can remove lines 18-28.

akofink: @daviddavis updated

ehelms:
ACK - thanks @akofink !

PR #278, opened 2016-09-01
fix missing `<my_github_username>`

tbrisker:


docs/development.md
```diff
@@ -13,7 +13,7 @@ This covers how to setup and configure a development environment using the Forkl
 A Katello development environment can be deployed on CentOS 6 or 7. Ensure that you have followed the steps to setup Vagrant and the libvirt plugin. There are a variety of useful development environment options that should or can be set when creating a development box. These options are designed to configure your environment ready to use your own fork, and create pull requests. To create a development box:
 
   1. Copy boxes.yaml.example to boxes.yaml. If you already have a boxes.yaml, you can copy the entries in boxes.yaml.example to your boxes.yaml.
-  2. Now, replace <my_github_username> with your github username
+  2. Now, replace `<my_github_username>` with your github username
```
ekohl: I wonder if the backticks are sufficient and we should remove the braces. @ehelms what do you think?

ekohl:
I like this, but I think there are more things in the file that need it as well. All the --options below it for example.

docs/development.md
ehelms: :+1:

tbrisker:
@ekohl thanks, added more backticks.

ekohl:
Thanks!

PR #287, opened 2016-09-09
Remove common role

ehelms:


ekohl:
ACK since it's unused now.

PR #290, opened 2016-09-12
Fix repositories cleanup for missing scl

### LINKED ISSUE #285: Clean-up prior installation missed to remove 'centos-release-scl'
```
$ ./setup.rb --skip-installer

Running shell command: yum clean all

Running shell command: yum -y update nss ca-certificates

Running shell command: rpm -e epel-release

Running shell command: rpm -e foreman-release

Running shell command: rpm -e katello-repos

Running shell command: rpm -e puppetlabs-release

Running shell command: yum -y  --disablerepo="*" --enablerepo=rhel-6-server-rpms install yum-utils wget
...
```
Re-running installation fails as  ```centos-release-scl``` repos are installed but disabled.
We disable all repos, then install centos-release-scl and it ends up with already installed package thus it doesnt not *enable* them.

The fix is to remove centos-release-scl* repos like we do with other repos at cleanup prior installation:
```rpm -e foreman-release-scl```

lpramuk:
Fix repositories cleanup for missing scl

(Fixes #285)

### DIFF
--- a/lib/forklift/repositories.rb
+++ b/lib/forklift/repositories.rb
@@ -45,6 +45,8 @@ module Forklift
       system('rpm -e katello-repos')
       system('rpm -e puppetlabs-release')
       system('rpm -e puppetlabs-release-pc1')
+      system('rpm -e foreman-release-scl')
+      system('rpm -e centos-release-scl centos-release-scl-rh')
     end
 
     def configure_rhel(os_version)


lpramuk:
Re-running installation fails as centos-release-scl repos are installed but disabled.
We disable all repos, then install centos-release-scl and it ends up with already installed package thus it doesnt not *enable* them.

ehelms:
ACK - thanks @lpramuk !

PR #294, opened 2016-09-13
remove branching path in playbook parsing

komidore64:


Vagrantfile
```diff
@@ -68,27 +68,12 @@ module Forklift
         @host_vars = {box['name'] => box['ansible']['variables']}
 
         if (playbooks = box['ansible']['playbook'])
-
-          if playbooks.is_a?(String)
-            args = SUPPORT_NAMED_PROVISIONERS ? ['main', type: 'ansible'] : [:ansible]
```
ehelms: I don't see this still being supported which was added to prevent breakage of bats on our CI system where an older version of vagrant is used.

I like the code consolidation though.

komidore64: oh, i assumed since it wasn't supported by both paths that it was maybe just something leftover? alright i'll see if i can squeeze it back in there.

komidore64: [added](https://github.com/Katello/forklift/pull/294/files#diff-23b6f443c01ea2efcb4f36eedfea9089R72)!

ehelms:
ACK - thanks @komidore64 ! If you wouldn't mind keeping an eye on the pipeline for a day or two to ensure this doesn't lead to breakages that would be appreciated.

PR #308, opened 2016-09-22
Add hammer-cli-foreman-admin to the centos7-hammer-devel box 

akofink:
Add hammer-cli-foreman-admin to the centos7-hammer-devel box to allow easy development on that repository.

Remove vestigial hammer.yml from a previous mail.yml role

Adjust config files to be copied from hammer repos and modified as
needed instead of built from scratch

Adjust Gemfile.local to use nested file path; include pry for debugging

CC: @lzap


playbooks/roles/hammer_devel/tasks/hammer_config.yml
```diff
@@ -19,6 +19,20 @@
         :username: {{ hammer_devel_username }}
         :password: {{ hammer_devel_password }}
 
+- name: 'Configure hammer-cli-foreman-admin'
+  blockinfile:
+    dest: ~/.hammer/cli.modules.d/foreman_admin.yml
+    create: true
+    block: |
+      :foreman_admin:
+        :enable_module: true
```
ehelms: A thought, we could store this information with the definition of the plugin and then make this generic, e.g.:

```
- name : "Configure {{ item.name }}"
  blockinfile:
    dest: "~/.hammer/cli.modules.d/{{ item.name }}.yml"
    create: true
    block: {{ item.config }}
  with_dict: "{{ hammer_plugins }}"
```

```
hammer_plugins:
  - name: foreman_admin
    repo: theforeman/foreman_admin
    config: |
      :foreman_admin:
         :enable_module: true
```

akofink: Yeah, that's a great idea - I was also thinking these files might just be copied from the hammer repository itself, since they all contain this simple config: https://github.com/Katello/hammer-cli-katello/tree/master/config That would allow future config changes to happen in the plugins rather than in Forklift.

Which would you prefer?

akofink: Yeah, we can use lineinfile to update config settings to be appropriate for this box. I'll get on it.

playbooks/roles/hammer_devel/tasks/hammer_config.yml
```diff
@@ -9,31 +9,43 @@
...
+    regexp: 'password'
+
+- name: 'Enable hammer-cli-foreman-admin'
+  command: cp ~/hammer-cli-foreman-admin/config/foreman_admin.yml ~/.hammer/cli.modules.d/
+
+- name: 'Configure hammer-cli-foreman-admin logging core'
+  command: cp ~/hammer-cli-foreman-admin/config/cli.modules.d/foreman_admin_logging_core.yml ~/.hammer/cli.modules.d/
+
+- name: 'Configure hammer-cli-foreman-admin logging for katello'
+  command: cp ~/hammer-cli-foreman-admin/config/cli.modules.d/foreman_admin_logging_katello.yml ~/.hammer/cli.modules.d/
```
ehelms: I'm ok if you prefer the explicit here, but you may want to consider a blanket * copy to get all the config files

akofink: Yeah, I was thinking about what happens if they change the file (i.e. from foreman.yml to foreman.yml.example, how it should be...) and I'd rather Forklift start failing from not finding the source file than from not configuring hammer properly (hammer wouldn't recognize example files, for instance) - hard fail vs. soft fail.

ehelms:
Thanks @akofink !

PR #309, opened 2016-09-22
Fix rubocop guard clause failure

ehelms:


ehelms:
@ekohl afaik, since the IF statement was found as the final statement, it could be a guard instead, but I thought that would look silly, e.g.:

```
 symlink(
        "#{curr_dir}/katello-installer/config/katello-devel.migrations",
        "#{foreman_config_root}/katello-devel.migrations"
      )

 return unless File.exist?("#{curr_dir}/katello-installer/modules")

 symlink(
   "#{curr_dir}/katello-installer/modules",
   '/usr/share/katello-installer-base/modules'
  )
end

```

ekohl:
I fully agree that's silly and I already noted somewhere else that I think the cop in this case doesn't make much sense to me. Then again, I'm not a big fan of guard clauses.

ehelms:
I like them for simple functions, like a function that just has an if statement in it, or if there are multiple conditions in the if I think it can help to make it clear what the function should do in each case to have multiple return guards.

PR #311, opened 2016-09-22
Updates version of Jenkins_job_builder in the container.

zjhuntin:
This makes it so we have the most up to date version of jenkins_job_builder in the docker container.


ehelms:
Thanks @zjhuntin !

PR #314, opened 2016-09-29
Only sort boxes if there is boxes.yaml

zjhuntin:


Vagrantfile
```diff
@@ -129,9 +129,11 @@ module Forklift
   plugin_vagrantfiles.each { |f| load f }
   plugin_base_boxes
   unsorted = @box_loader.add_boxes('boxes.yaml', 'config/versions.yaml') if File.exists?('boxes.yaml')
-  @boxes  = unsorted.keys.sort.inject({}) do |hash, key|
-    hash[key] = unsorted[key]
-    hash
+  if unsorted
+    @boxes  = unsorted.keys.sort.inject({}) do |hash, key|
```
ehelms: Could this just be `@boxes.keys.sort` ? I'm not clear why the original unsorted variable was ever needed.

Vagrantfile
```diff
@@ -129,9 +129,11 @@ module Forklift
   plugin_vagrantfiles.each { |f| load f }
   plugin_base_boxes
   unsorted = @box_loader.add_boxes('boxes.yaml', 'config/versions.yaml') if File.exists?('boxes.yaml')
-  @boxes  = unsorted.keys.sort.inject({}) do |hash, key|
-    hash[key] = unsorted[key]
-    hash
+  if unsorted
```
ehelms: This changes the behavior to essentially only sort if you have a `boxes.yaml`

Vagrantfile
zjhuntin: I think it's needed because it's adding and sorting the boxes that are taken from the boxes.yaml file, and then sorts them.

zjhuntin: Oh, I see what you mean.

ehelms: Right but `@boxes` contains everything, just as unsorted does, because add_boxes returns ALL boxes. @jlsherrill was there a reason you couldn't just store the result in `@boxes` and then sort that?

jlsherrill: I re-factored this code a few times while i was writing it, and I think i switched to inject so that i could store it all in @boxes but forgot to go back and get rid of the unsorted variable, so yes, you can use @boxes

zjhuntin:
@ehelms updated to just use @boxes.

ehelms:
Looks good -- @jlsherrill ?

jlsherrill:
ACKity ACK ACK

ehelms:
Thanks @zjhuntin !

zjhuntin:
Actually, let me change the title of the commit real fast.

PR #318, opened 2016-10-04
Refactor box loading to be more explicit

ekohl:
The foreman repositories no longer has an EL6 repository so this can't
work anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,14 @@ Forklift provides tools to create Foreman/Katello environments for development, 
 
 ## Using Forklift
 
+### Requirements
+
+* Vagrant - 1.8+ - Both the VirtualBox and Libvirt providers are tested
+* Ansible - 2.1+
+
 ### Quickstart
 
-This will walk through the simplest path of spinning up a production test environment of a bleeding edge nightly installation assuming vagrant and Libvirt are installed and configured.
+This will walk through the simplest path of spinning up a production test environment of a bleeding edge nightly installation assuming Vagrant and Libvirt are installed and configured.
 
 ```
 git clone https://github.com/katello/forklift.git
--- a/playbooks/roles/selinux/tasks/main.yml
+++ b/playbooks/roles/selinux/tasks/main.yml
@@ -1,0 +1,6 @@
+---
+- name: 'Ensure libselinux-python is installed'
+  yum:
+    name: 'libselinux-python'
+    state: installed
+- selinux: policy=targeted state={{ selinux_state }}
--- a/config/base_boxes.yaml
+++ b/config/base_boxes.yaml
@@ -8,20 +8,16 @@ shells:
 
 boxes:
   centos6:
-    box_name:   'centos6'
+    box_name:   'centos/6'
     image_name: !ruby/regexp '/CentOS 6.*PV/'
     default:    true
     pty:        true
-    virtualbox: 'http://opscode-vm-bento.s3.amazonaws.com/vagrant/vmware/opscode_centos-6.6_chef-provisionerless.box'
-    libvirt:    'http://m0dlx.com/files/foreman/boxes/centos64.box'
 
   centos7:
-    box_name:   'centos7_1'
+    box_name:   'centos/7'
     image_name: !ruby/regexp '/CentOS 7.*PV/'
     default:    true
     pty:        true
-    virtualbox: 'http://opscode-vm-bento.s3.amazonaws.com/vagrant/virtualbox/opscode_centos-7.1_chef-provisionerless.box'
-    libvirt:    'https://download.gluster.org/pub/gluster/purpleidea/vagrant/centos-7.1/centos-7.1.box'
 
   centos7-atomic:
     box_name: centos7-atomic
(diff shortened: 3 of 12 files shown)


ekohl:
Updated the loader and tests but now I realise this also removes all the older versions. We need to rethink this because I'm not sure what's the best solution here.

ekohl:
@stbenjam or @ehelms any idea what would be a good way to keep centos 6 with older foreman releases but not nightly? I think we may need some sort of support matrix.

ehelms:
@ekohl I meant to indicate I had seen this and was still pondering on it :)

ekohl:
I rewrote it. Please have a look if the general idea is correct. I know this is currently broken and also removes the puppet 4 configs but it's mostly about the versions.yaml rewrite.

ekohl:
Oh, and it will most likely go very nicely together with https://github.com/Katello/forklift/pull/322 and works towards https://github.com/Katello/forklift/issues/302

ehelms:
I *think* with your re-write we could remove this convolutedness I wrote -- https://github.com/Katello/forklift/blob/master/playbooks/roles/katello_repositories/defaults/main.yml#L2-L7

ehelms:
With your re-factor this moves us strictly on to Vagrant 1.8+ due to the use of host_vars (which is great). So I think this PR is a good candidate to add a 'Requirements' section to the top end of the Readme to spell out our minimums.

ehelms:
Overall I like it, the code itself looks simplified and the config is quite obvious. Keeps the generation dynamic but since its more simplified I think it will be more obvious when debugging.

ekohl:
Ok, I'm testing around a bit and will come up with a rebased PR when #322 has been merged since that makes supporting foreman boxes much easier.

ekohl:
I still need to update the README, fix the tests, verify the capsule boxes work. I also started with using the atlas boxes because the centos 6 box couldn't connect to EPEL anymore but the atlas box is lacking libselinux-python which ansible doesn't like. Haven't tried to fix that yet.

ekohl:
Updated. I tested:
* Provision centos6-foreman-1.12
* Provision centos7-katello-nightly
* Provision centos7-capsule-nightly

I'll try to test other boxes for the atlas box update though I may use a similar approach as in https://github.com/Katello/forklift/pull/316 with legacy boxes.

README.md
```diff
@@ -18,7 +18,7 @@ Forklift provides tools to create Foreman/Katello environments for development,
 
 ### Quickstart
 
-This will walk through the simplest path of spinning up a production test environment of a bleeding edge nightly installation assuming vagrant and Libvirt are installed and configured.
+This will walk through the simplest path of spinning up a production test environment of a bleeding edge nightly installation assuming Vagrant (version 1.8+) and Libvirt are installed and configured.
```
ehelms: What do you think about adding something like:

```
### Requirements

 * Vagrant - 1.8+
 * Ansible - 2.1+
```

playbooks/roles/selinux/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: 'Ensure libselinux-python is installed'
+  yum:
+    name: 'libselinux-python'
+    state: installed
```
ehelms: This would be a good role to also had enforce/permissive setting of SELinux which is sometimes needed for testing or devel setups.

ehelms: And could help ensure SELinux is enforcing by default in case the box someone uses has it disabled for any reason

README.md
ekohl: I like it, will change this.

playbooks/roles/selinux/tasks/main.yml
ekohl: I'll look into this.

ekohl:
Updated:
* Selinux is now set to enforcing but is a parameter (ready for the devel box)
* Documented the minimal version in README
* Fixed the Rubocop failures
* Fixed provisioning the freeipa server but I don't think it properly sets up the katello server

ehelms:
This inspired me to open up https://github.com/Katello/forklift/pull/326 for discussion

ekohl:
I think this is ready for merge though I'm still a bit unsure about the pipeline.

ehelms:
@ekohl Apologies, forgot about this one! Will prioritize re-review

ehelms:
Looks great - going to merge and see what happens :)

lpramuk:
Format change of ```config/versions.yaml``` has broken ```katello_version``` mapping func in ```forklift/lib/forklift/repositories.rb```:
```
    def katello_version
      @versions['mapping'][@version]
    end
```

ekohl:
Could you try replacing it with
```ruby
def katello_version
  @versions['installers'].select {|scenario| scenario['foreman'] == @version }['katello'] rescue nil
end
```

lpramuk:
Like almost, you missed ```[0]```
```
@versions['installers'].select {|scenario| scenario['foreman']=='nightly'}[0]['katello']
```

PR #319, opened 2016-10-04
Use the Ansible playbook instead of setup.rb for centos7-katello-nightly

ekohl:


ekohl:
Note this effectively drops ```--disable-selinux``` but it provisions correctly and I think it should just work.

PR #326, opened 2016-10-23
Add ansible-lint and starter rules

ehelms:


ehelms:
This adds ansible-lint (https://github.com/willthames/ansible-lint) and a small selection of rules to begin with.

.travis.yml
```diff
@@ -1,5 +1,11 @@
...
-  - "2.2.0"
+language: python
+python:
+  - 2.7
+install:
+  - pip install ansible-lint
+  - rvm install 2.2.4
+script:
+  - bundle install
+  - rake
```
ekohl: Shouldn't this be ```bundle exec rake```?

rules/TaskManyArgs.py
```diff
@@ -0,0 +1,18 @@
+from ansiblelint import AnsibleLintRule
+
+class TaskManyArgs(AnsibleLintRule):
+    id = 'E303'
+    shortdesc = 'Use ":" YAML format when arguments are over 3'
+    description = ''
+    tags = ['task']
+
+    def match(self, file, text):
+        count = 0
```
ekohl: Note sure if it's more readable, but this can be written as:

```python
count = len([part for part in text.split(" ") if "=" in part])
return count > 3
```

ehelms:
@ekohl updated. @stbenjam you may also want to review this

rules/TaskVariableHasSpace.py
```diff
@@ -0,0 +1,19 @@
...
+
+    compiled = re.compile(ur'{{(\w*)}}')
+
+    def match(self, file, text):
+        m = self.compiled.search(text)
+        if m:
+            return True
+        return False
+
+
```
ekohl: Nit: whitespace

rules/TaskVariableHasSpace.py
```diff
@@ -0,0 +1,19 @@
...
+    id = 'E305'
+    shortdesc = 'Variables should be enclosed by spaces "{{ foo }}"'
+    description = ''
+    tags = ['task']
+
+    compiled = re.compile(ur'{{(\w*)}}')
+
+    def match(self, file, text):
+        m = self.compiled.search(text)
+        if m:
```
ekohl: I'd prefer ```return bool(m)```

rules/TaskVariableHasSpace.py
```diff
@@ -0,0 +1,19 @@
...
+
+import re
+
+class TaskVariableHasSpace(AnsibleLintRule):
+    id = 'E305'
+    shortdesc = 'Variables should be enclosed by spaces "{{ foo }}"'
+    description = ''
+    tags = ['task']
+
+    compiled = re.compile(ur'{{(\w*)}}')
```
ekohl: Can you explain why this should do? Should it enforce ```{{ foo }}```? In that case I don't think this matches ```{{ foo}}``` or ```{{foo }}``` which you probably also don't want.

ehelms:
@ekohl for awareness, these rules come from a combination of https://github.com/willthames/ansible-lint and https://github.com/tsukinowasha/ansible-lint-rules which there is not a way to selectively choose which rules to enable other than adding them to your own project in a directory from my research thus far.

ekohl:
Ah, I'll submit the PRs there then :)

ehelms:
@ekohl we are having to copy the rules, and they aren't that complex. So I think our options are:

 1) Just carry our own rules that are based on or copied from other projects
 2) Try to add a configuration option to turn off and on different rules to ansible-lint and then git sub-module the various rules repos

Thoughts?

ekohl:
@stbenjam I noticed that the linked repository has the max argument count at 4 where we have 3. That's fine, but there's nothing like a parameter we can tune. IMHO it should be a python package we could install with some config file / command line parameters but if this is how the ansible community does it then for now I think we just have to copy them.

ekohl:
Oh, maybe it is a good idea to link the source of the file so others don't make the same mistake as I did ;)

ehelms:
@ekohl pointed out exactly what I was going to, that ansible-lint has built itself to be flexible by allowing you to specify your own rules easily but lacks configuration. You can't alter rule parameters (https://github.com/Katello/forklift/pull/326/files#diff-61f373c7d8471479380899beee8d5f9cR11) or turn things on and off currently like you can with Ruby and JS linters for example.

So we either just do it this way or try to get that functionality added to `ansible-lint`. If we just carry our rules in our own repo then I'd say we code them like we think they ought to be @ekohl

ekohl:
Is it license wise OK to just copy them?

ehelms:
@ekohl they are MIT, so I believe so.

ekohl:
Given there are 2 approvals I think we can merge this.

PR #330, opened 2016-11-14
Handle the new versions.yaml format

ekohl:
dd7e3bbacc9371523e8a010b8396483e51eaa7f0 introduced a regression when it
rewrote versions.yaml.

Reported in https://github.com/Katello/forklift/pull/318#issuecomment-260351002

### DIFF
--- a/lib/forklift/repositories.rb
+++ b/lib/forklift/repositories.rb
@@ -34,7 +34,7 @@ module Forklift
     end
 
     def katello_version
-      @versions['mapping'][@version]
+      @versions['installers'].select { |scenario| scenario['foreman'] == @version }[0]['katello']
     end
 
     def cleanup


lpramuk:
We need this: (I figure it out though I dont know ruby at all ;-)
```@versions['installers'].select { |scenario| scenario['foreman'] == @version }[0]['katello']```

ehelms:
APT

ehelms:
@ekohl we probably ought to put out a deprecation warning since I think we are nearing the end of setup.rb

ekohl:
Updated, not sure how to handle the rescue part so I'll just let it crash for now. It would most likely have with it anyway.

ekohl:
> @ekohl we probably ought to put out a deprecation warning since I think we are nearing the end of setup.rb

We have https://github.com/Katello/forklift/issues/305 for that. There's also some setup.rb left in bats tests, other than that we just need users to migrate.

PR #334, opened 2016-11-15
Directly use bats in the bats role

ekohl:


ekohl:
This ports the wrappers, but I don't know if this is the desired state. Do we want to use ```/vagrant/forklift``` instead of cloning again? Should the tests to run be a parameter?

ehelms:
Tests could be good as a parameter, but could be added after the fact when needed. I am fine with cloning again as that ensures we get the desired state.

ekohl:
I'd prefer to parametrize it later on.

ekohl:
Pushed a commit to parametrize it. Currently I'm not a machine with sufficient RAM to run the pipeline so it's untested.

ehelms:
@ekohl This branch needs a rebase

ekohl:
Rebased and changed. I'm also using it to run on local bats files as follows (left out the actual foreman provisioning for brevity):
```yaml
c7-fm-bind:
  box: 'centos7'
  ansible:
    playbook: "playbooks/bats.yml"
    group: "bind"
    variables:
      bats_forklift_dir: "/vagrant"
      bats_update_forklift: "no"
      bats_tests: "fb-proxy-dns.bats"
---

By default vagrant with libvirt only updates ```/vagrant``` on host boot or through ```vagrant rsync```, but I'm looking into the NFS sharing so it is live. The benefit is that you don't have to push everything to github and makes writing bats tests easier.

File: playbooks/roles/bats/defaults/main.yaml [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+bats_git_dir: "/root/bats"
+bats_forklift_dir: "/root/forklift"
+bats_update_forklift: "yes"
+bats_tests: "fb-content-katello.bats fb-capsule.bats fb-finish-katello.bats"
```
ehelms: Would be nice if we could use the "named" versions here the same as when you call `katello-bats` (e.g. `katello-bats content`)

ehelms: Although maybe the named ones from `katello-bats` get you disconnected from the bats files themselves.

ekohl: Personally I prefer the explicitness of filenames over the shorter named ```katello-bats``` versions. To me it's more obvious and allows for grepping where a file is used.

PR #335, opened 2016-11-17
Fixes capsule install playbook

johnpmitsch:
Capsule install playbook needs the katello version
and epel repos set up


cfouant:
APT thanks a lot @johnpmitsch !

PR #339, opened 2016-12-07
Fix Vagrantfile lookup

kahowell:
This allows usage of `VAGRANT_CWD`, and allows subdirectories of
forklift to detect the `Vagrantfile` properly. Required not referencing
anything via relative paths.


ehelms:
Thanks @kahowell !

ekohl:
Oh this is much cleaner than the ugly workaround I had locally (which I won't even share). Thanks!

PR #342, opened 2016-12-14
katello-certs-check not found on devel install

dLobatog:
Since katello-certs-check is not found on the -devel installer, and a
hook (pre/20-certs_update.rb) uses it, let's skip the certs check as it
just throws an error on -devel.

Alternatively (and I'd prefer it) we could provide this tool + katello-service on devel installs too, I'm not sure why it's not on the package.

### DIFF
--- a/playbooks/devel.yml
+++ b/playbooks/devel.yml
@@ -6,6 +6,7 @@
     foreman_installer_scenario: katello-devel
     foreman_installer_options_internal_use_only:
       - "--disable-system-checks"
+      - "--certs-skip-check"
       - "--katello-devel-github-username {{ katello_devel_github_username }}"
     foreman_installer_additional_packages:
       - foreman-installer-katello-devel


ekohl:
I'd have expected that foreman-installer-katello-devel (which is in the extra packages) depended on foreman-installer-katello.

NeilHanlon:
ack

ekohl:
I can't replicate this on my machine, how are you running into this? Could you paste your ```boxes.yaml```?

NeilHanlon:
@ekohl 

boxes.yaml is default from the distribution. Just running a "vagrant up $boxname" causes this to happen because the foreman-installer-katello-devel package does not depend on the katello-certs-check package. As @dLobatog said, the hook tries to check for certs but fails because the package isn't installed.

NeilHanlon:
Looking into this a bit more, this is what the problem is. It's not a package dependency issue, but rather foreman-installer-katello-devel is not packaged with the script, but foreman-installer-katello is.

    /usr/share/foreman-installer-katello/bin/katello-certs-check
    [vagrant@centos7-devel ~]$ repoquery -q -l --plugins foreman-installer-katello|grep certs-check
    /usr/sbin/katello-certs-check
    /usr/share/foreman-installer-katello/bin/katello-certs-check
    [vagrant@centos7-devel ~]$ repoquery -q -l --plugins foreman-installer-katello-devel|grep certs-check

ehelms:
I think the question is really does this cause an error or breakage when trying to spin up a devel install. If not, I think we can wait on the proper fix (https://github.com/Katello/katello-packaging/pull/332).

For some background, the idea in the past had been to make development environments quicker and lighter weight than a production install. One way to do that was to reduce the RPM dependencies which some of these things are a product of. We had found in the early days that if dev environments were too heavy or to slow people were less likely to re-spin them and wind up with old, outdated setups.

NeilHanlon:
@ehelms Yes. It really does cause an error spinning devel installs.

https://gist.github.com/NeilHanlon/812142caec1ba0594c84392504ff780a

I replicated it yesterday but lost my buffer somehow.

ekohl:
You're right, I hit it too now. Wonder why I didn't hit it yesterday.

ekohl:
With the packaging PR merged, should this be closed now?

dLobatog:
Yeah @ekohl - closing

PR #348, opened 2016-12-29
Bats testing updates

ehelms:


bats/fb-content-katello.bats
```diff
@@ -10,91 +10,97 @@ setup() {
   tSetOSVersion
 }
 
+@test "Create an Empty Organization" {
+  org=`hammer -u admin -p changeme organization list | grep -q "Empty Organization"`
+
+  if [ $org != 0 ]; then
```
ekohl: Does this work? Don't you need to test ```$?``` here?

bats/fb-content-katello.bats
```diff
@@ -10,91 +10,97 @@ setup() {
   tSetOSVersion
 }
 
+@test "Create an Empty Organization" {
+  run hammer -u admin -p changeme organization info --name "Empty Organization"
```
ehelms: @ekohl this was the  magic I was looking for so that this didn't cause a test failure

ehelms:
I'd agree with your assessment @ekohl helper would be nice but in a separate PR.

bats/fb-content-katello.bats
ekohl: Had to think for a moment, but you mean ```run```?

bats/fb-content-katello.bats
```diff
@@ -172,3 +196,8 @@ EOF
   skip # because of above
   puppet agent --test && grep -q Lorem /tmp/dummy
 }
+
+#@test "Delete an Organization" {
```
ekohl: Is this a local test leftover?

bats/fb-content-katello.bats
ehelms: Yea - otherwise this was failing the test when really I just want the output and status to use for the real test.

bats/fb-content-katello.bats
ehelms: Kinda - I'm not sure the deletion is actually working properly despite "passing" this test :/ 

@jlsherrill @parthaa any idea if organization delete actually works today?

ehelms:
@ekohl can you re-review this?

ekohl:
@ehelms I was waiting for an answer about the commented code. I'd prefer to avoid adding commented code, but if it's known to be broken then a FIXME with a link to redmine issue would be acceptable.

bats/fb-content-katello.bats
ehelms: ping @jlsherrill @parthaa

jlsherrill: org deletion 'works', however you may get errors/warnings that prevent deletion if certain things like hostgroups are associated to the organization.

bats/fb-content-katello.bats
jlsherrill: do you mind adding a comment here to explain why this org is being created? (so that we are assured a 2nd org exists when we try to delete Test Org)

ehelms:
Updated and uncommented out the deletion test.

ehelms:
@ekohl @jlsherrill Updated with comment

PR #349, opened 2016-12-29
Deploying Foreman and Katello on OpenShift Origin V3

ehelms:
This work is a work in progress to explore the ability to deploy Foreman and Katello on to an OpenShift V3 (container and Kubernetes based environment) environment to explore harnessing the power of the OpenShift platform to manage the application and scale out.

For more detailed information on this deployment, how to use, test and contribute see the README (https://github.com/ehelms/forklift/blob/openshift/openshift/README.md).

This work combines work built on top of the foreman-docker-compose repository (https://github.com/shlomizadok/foreman-docker-compose) along with previous work by the Pulp and Candlepin teams to build out images for use on OpenShift. This also pulls in a starter repo (https://github.com/ehelms/foreman-openshift) that was used to hold initial work on the OpenShift piece.

I have initially pulled the aforementioned repositories into a PR to Forklift (using git subtree to preserve the history) for a few reasons:

 * Forklift's goal is around putting Foreman into user and developer environments
 * Forklift has a broader audience than a new set of repositories and branches would have
 * To get more eyes and interest on this idea for deploying Foreman

If others feel this would be better maintained in a separate repository, I'm open to that discussion with the hope that the previous reasons will be thoughtfully considered with respect on how to position this.


openshift/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+Now that we have docker and the appropriate client tools setup we can spin up a docker based OpenShift cluster using the playbook to do this:
+
+    ansible-playbook cluster-up.yml
+
+## Creating the Foreman Deployment
+
+There are two aspects to the deployment: the default images and templates being loaded into OpenShift and creating the Foreman deployment. All of this is handled by a convenient playbook provided. You can view the configuration of the Foreman deployment itself by looking in the `templates/foreman.yaml` file. To create the application:
+
+    ansible-playbook ansible/playbooks/foreman.yaml
```
ekohl: File is called .yml

openshift/cluster-up.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: "Login back in as developer"
+      command: oc login -u developer -p a
+
+    - name: "Delete initial project"
+      command: oc delete project myproject
+    
+    - name: 'Create .tmp directory'
+      file:
+        path: .tmp
```
ekohl: I had to add ```state: directory```

ehelms:
Thanks for the feedback so far @ekohl . I made changes based on your comments. There was a top level gitignore that was blocking the puppet role file from being added - fixed now. I also removed sudo from the volume creation. Volume creation and management is one of the trickier aspects. At this point, I've taken the approach of creating a local directory with subpaths for each volume "store" needed by the various services.

openshift/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Setup OpenShift
+
+Now that we have docker and the appropriate client tools setup we can spin up a docker based OpenShift cluster using the playbook to do this:
+
+    ansible-playbook cluster-up.yml
+
+## Creating the Foreman Deployment
+
+There are two aspects to the deployment: the default images and templates being loaded into OpenShift and creating the Foreman deployment. All of this is handled by a convenient playbook provided. You can view the configuration of the Foreman deployment itself by looking in the `templates/foreman.yaml` file. To create the application:
```
ekohl: This should also be .yml

ehelms:
Thanks @ekohl - updated based on your comments.

A few things I wouldn't mind some feedback on:

 1) currently I am tracking TODOs in the README on this PR, I was thinking about either:
   a) having this live in its own repository and using github issues to manage the todos
   b) using this repo (since this is sorta Forklift's mission) and using github issues + milestone for openshift work to track

 2) creating an RFC either contained here with this PR and work or as part of the official RFC repo

ekohl:
I wouldn't mind if this lived in this repo, but we already have quite a few ways this repo can be used and we need to cut down on it and focus the documentation. As someone new to the repo it's hard to figure out what you can and can't do with it. Before we add more, we need to simplify it first. A few steps that I'd think of:

# Split the top directory into subdirs

Some candidates I see

* bats
* vagrant
* openshift
* docker

# Remove deprecated things

* Nuke setup.rb

ehelms:
That sounds reasonable as a first good step to re-organize the top level (unless you think it should be part of this change only). The top level docker repository I don't know if anyone outside of myself even uses at this point. There is also a docker folder inside this PR, where in theory generic docker build directories are living. They could perhaps be moved to the top to help emphasize keeping them multi-use; I had put them there originally cause I was having to do some hacks to get them to work inside openshift.

I am all for nuking setup.rb. I think the biggest hold back of it is the use in `bats` and thus the use in our CI system.

ekohl:
> That sounds reasonable as a first good step to re-organize the top level (unless you think it should be part of this change only).

No, I'd prefer to get that done first and then get this merged.

> The top level docker repository I don't know if anyone outside of myself even uses at this point.

That's a big problem with this repo IMHO. We have a lot of possibilities and they should be maintained. I'd like to go to a situation where each top level directory can be used independently.

> There is also a docker folder inside this PR, where in theory generic docker build directories are living. They could perhaps be moved to the top to help emphasize keeping them multi-use; I had put them there originally cause I was having to do some hacks to get them to work inside openshift.

That's fine, until they are generic they should live inside openshift.

> I am all for nuking setup.rb. I think the biggest hold back of it is the use in bats and thus the use in our CI system.

Agreed. Which is why I think we should get https://github.com/Katello/forklift/pull/331 merged and look at migrating the CI system.

ehelms:
@ekohl I hope you don't mind, since we begin to talk about more than just the Openshift aspect, I moved it to an issue (https://github.com/Katello/forklift/issues/367) with the previous points quoted.

ekohl:
Yes, let's keep this on topic.

ehelms:
Replaced by https://github.com/theforeman/forklift/pull/424

PR #350, opened 2016-12-30
Add bats tests for the proxy DNS

ekohl:
Only tested against Foreman 1.14 RC2 with PowerDNS and it assumes zones are already present. It should be generic and easily add any needed cache flushes for other nameservers.

### DIFF
--- a/bats/fb-proxy-dns.bats
+++ b/bats/fb-proxy-dns.bats
@@ -1,0 +1,142 @@
+#!/usr/bin/env bats
+# vim: ft=sh:sw=2:ts=2:et
+
+set -o pipefail
+
+load os_helper
+load foreman_helper
+
+setup() {
+  tSetOSVersion
+
+	tCommandExists curl || tPackageInstall curl
+	tCommandExists dig || tPackageInstall bind-utils || tPackageInstall dnsutils
+
+  export HOSTNAME=$(hostname -f)
+}
+
+tSpRequest() {
+  if [ -d /etc/puppetlabs/puppet/ssl ] ; then
+    SP_KEY=/etc/puppetlabs/puppet/ssl/private_keys/${HOSTNAME}.pem
+    SP_CERT=/etc/puppetlabs/puppet/ssl/certs/${HOSTNAME}.pem
+    SP_CACERT=/etc/puppetlabs/puppet/ssl/ca/ca_crt.pem
+  else
+    SP_KEY=/var/lib/puppet/ssl/private_keys/${HOSTNAME}.pem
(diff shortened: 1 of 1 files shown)


bats/fb-proxy-dns.bats
```diff
@@ -0,0 +1,140 @@
...
+  run tGetRecord ch txt version.bind
+  case $output in
+    *PowerDNS*)
+      pdns_control purge $1
+      ;;
+  esac
+}
+
+@test "verify dns feature is enabled" {
+  tSpRequest features | jq -e 'map(in({"dns": 42})) | any'
```
ekohl: ```grep '"dns"'``` removes the need for jq

bats/fb-proxy-dns.bats
```diff
@@ -0,0 +1,140 @@
...
+}
+
+@test "clean up any existing records" {
+  for type in A AAAA PTR CNAME ; do
```
ekohl: ```PTR``` is not needed since that's an invalid PTR here. We should drop the v4 and v6 PTRs though.

ekohl:
Verified to work on named as well. Replicate using:
```yaml
---
c7-fm-bind:
  box: 'centos7'
  ansible:
    playbook:
      - 'playbooks/foreman.yml'
    group: 'bind'
    variables:
      foreman_version: '1.14'
      puppet_repositories_version: 4
      foreman_installer_options: "'--foreman-proxy-dns=true
         --foreman-proxy-dns-managed=true
         --foreman-proxy-dns-reverse=121.168.192.in-addr.arpa'"
```

Then you also need to create the IPv6 reverse range:

```sh
cp /var/named/dynamic/db.121.168.192.in-addr.arpa /var/named/dynamic/db.8.b.d.0.1.0.0.2.ip6.arpa
cat >> /etc/named/zones.conf <<EOF
zone "8.b.d.0.1.0.0.2.ip6.arpa" {
    type master;
    file "/var/named/dynamic/db.8.b.d.0.1.0.0.2.ip6.arpa";
    update-policy {
            grant rndc-key zonesub ANY;
    };
};
EOF
systemctl reload named
```

Possibly you can also work around it by using ```--foreman-proxy-dns-reverse=arpa```

The CNAME test will fail until https://github.com/theforeman/smart-proxy/pull/489 is merged.

ehelms:
@ekohl If I read correctly this wont fully pass currently? If so do you want to get this merged and add it to the pipeline once the mentioned PR is merged? Or would you prefer to wait?

ekohl:
@ehelms good point. I think I'd prefer to get this merged now and only add it to the pipeline later.

PR #351, opened 2017-01-02
Instructions to run webpack-dev in HTTPS

dLobatog:
Without this, webpack-dev-server does noto run in Forklift due to Apache 
forcing HTTPS on every connection, and Foreman will get mixed content
warnings.

I intend to automate this with Ansible at some point, but for the time being, let's have some instructions so people can develop using this already.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -10,6 +10,7 @@ This covers how to setup and configure a development environment using the Forkl
  * [Hammer Development](#hammer-development)
  * [Capsule Development](#capsule-development)
  * [Client Development](#client-development)
+ * [Webpack](#webpack)
 
 ## Development Environment Deployment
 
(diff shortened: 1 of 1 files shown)


ekohl:
Oh this is very useful. I'm going to try and verify this, possibly take a stab at automating it.

README.md
```diff
@@ -193,3 +194,33 @@ module MyCustomPlugins
   end
 end
 ```
+
+### Webpack
+
+Forklift creates a reverse proxy between localhost:3000 (Rails development default server) and https://forkliftfqdn (e.g: https://centos7-devel.example.com). This will cause mixed-content errors if you try to enable the webpack server without any options, as it needs to use the right certificates for HTTPS and Foreman needs to allow it. If you do not need to hack on Webpack code, you can simply run `npm webpack:compile` from `~/foreman` and that will create the necessary assets to be served by Foreman. If you want to work on Webpack code, follow these instructions:
```
dLobatog: `bundle exec rake webpack:compile`, sorry

ehelms:
This only applies to development environments, yea? I'd think it would be better suited here -- https://github.com/Katello/forklift/blob/master/docs/development.md

dLobatog:
@ehelms Yup - only development environments. I'll move it there

ekohl:
It appears ```webpack_dev_server``` defaults to ```true``` in development so ```bundle exec rake webpack:compile``` is insufficient.

dLobatog:
Added a small note about ensuring `webpack_dev_server` is set to false before running `webpack:compile`

ehelms:
To be clear, this change is only required if you access the webui through the HTTPS entry point vs. directly through the development port (`:3000`) ?

dLobatog:
Righto - only required for https

ehelms:
Thanks @dLobatog !

PR #369, opened 2017-01-27
Remove setup.rb and replace with Ansible

ehelms:


playbooks/inventory/vagrant.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
```
ekohl: Why revert this change?

playbooks/inventory/vagrant.py
```diff
@@ -17,9 +17,18 @@ def parse_args():
 
 
 def list_running_hosts():
+    hosts = {}
+
+    try:
+        status = subprocess.call(["which","vagrant"])
```
ekohl: if you use ```check_call``` then you don't need to check the exit code since it'll raise an exception.

playbooks/inventory/vagrant.py
```diff
@@ -33,7 +42,14 @@ def list_running_hosts():
 
 def get_host_details(host):
     cmd = "vagrant ssh-config {}".format(host)
+    print(cmd)
```
ekohl: This looks like some debug code.

playbooks/inventory/vagrant.py
ehelms: Oops -- I was testing issues with this script and thought that was causing it. I will put that back.

ehelms:
Mind checking https://github.com/theforeman/foreman-infra/pull/273/files before I merge this so I can get the CI system updated shortly after I merge this?

ekohl:
Yay!

lpramuk:
@ehelms in nightly automation we used `./setup.rb`
Can you give me a hint for the replacement.
Ansible? What? How to run it? Where to grab it? ;-)

ehelms:
The replacement is to use Ansible. If you are wanting to spin up your own server, you'd clone Forklift, and run `ansible-playbook playbooks/katello.yml` for example.

PR #373, opened 2017-01-31
Name client koji repo properly

ehelms:
Installation was failing due to two repositories with the same name
being laid down and thus Katello main repository wasn't available.


ehelms:
Current failures on the nightly pipeline are a result of this.

PR #374, opened 2017-01-31
Default bats should be server tests only

ehelms:
The fb-proxy.bats tests are aimed at an external foreman proxy
and thus break when run on the server.


ehelms:
The next failure that Ansible bats run will hit.

PR #375, opened 2017-01-31
Ouput bats tests to tap output file

ehelms:
The idea is this will spit out an output file into `/root` per test that is run. Those results can then be scp'd off in Jenkins to feed into the TAP result plugin to trend over time and mark the build success or failure (very similar to what was done prior to the removal of setup.rb).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -69,14 +69,12 @@ module Forklift
           @ansible_groups["server-#{box.fetch('name')}"] = box['ansible']['server']
         end
 
-        @host_vars = {box['name'] => box['ansible']['variables']}
-
         if (playbooks = box['ansible']['playbook'])
           [playbooks].flatten.each_with_index do |playbook, index|
             args = SUPPORT_NAMED_PROVISIONERS ? ["main#{index}", type: 'ansible'] : [:ansible]
             machine.vm.provision(*args) do |ansible|
               ansible.playbook = playbook
-              ansible.host_vars = @host_vars if @host_vars && Gem::Version.new(Vagrant::VERSION) >= Gem::Version.new('1.8.0')
+              ansible.extra_vars = box['ansible']['variables'] if Gem::Version.new(Vagrant::VERSION) >= Gem::Version.new('1.8.0')
               ansible.groups = @ansible_groups
             end
           end
--- a/playbooks/roles/bats/tasks/main.yml
+++ b/playbooks/roles/bats/tasks/main.yml
@@ -21,7 +21,13 @@
     dest: "{{ bats_forklift_dir }}"
     update: "{{ bats_update_forklift }}"
 
+- name: "Bats output directory"
+  file:
+    state: "directory"
+    path: "{{ bats_output_dir }}"
+
 - name: "Run bats"
-  shell: "bats {{ bats_tests }}"
+  shell: "bats --tap {{ item }} > {{ bats_output_dir }}/{{ item }}.tap"
   args:
     chdir: "{{ bats_forklift_dir }}/bats"
+  with_items: "{{ bats_tests }}"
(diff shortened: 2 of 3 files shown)


ehelms:
This, or some form of this change is needed to fix the pipeline.

playbooks/roles/bats/tasks/main.yml
```diff
@@ -22,6 +22,7 @@
     update: "{{ bats_update_forklift }}"
 
 - name: "Run bats"
-  shell: "bats {{ bats_tests }}"
+  shell: "bats {{ item }} 2>&1 | tee /root/{{ item }}.out"
```
ekohl: Maybe make this directory a parameter as well?

ehelms:
I an look into whether vagrant Ansible provisioner fixed this. We are sadly butting up against best practice for use of the roles and the constraints of one user of them. If I have to, I can punt to parsing this by coma separated values like the merge request testing in the foreman_installer role. I still want to support bats running each test separately and producing a separate output file as I think that helps identify the test sets and failures therein.

playbooks/roles/bats/tasks/main.yml
ekohl: Also, what's the rationale for ```.out```? Since they're in tap output, would ```.tap``` be a logical choice?

ekohl:
And I forgot to mention, maybe we should just fix the path parameter, merge this now and unblock testing. We can look into it when we want to extend it again.

playbooks/roles/bats/tasks/main.yml
ehelms: Fair question, `.out` was what it was before and Foremans systest were using. But I tend to agree that would make more sense.

ehelms:
I didn't mean to imply that it was best practices per any document. I was more trying to say why use CSV when you can define a native array.

ehelms:
@ekohl take a look at this change for the variables and the output dir

@stbenjam I believe you implemented the first use of `host_vars`, mind looking this change over?

playbooks/roles/bats/tasks/main.yml
```diff
@@ -21,7 +21,13 @@
...
     update: "{{ bats_update_forklift }}"
 
+- name: "Bats output directory"
+  file:
+    state: "directory"
+    path: "{{ bats_output_dir }}"
+
 - name: "Run bats"
-  shell: "bats {{ bats_tests }}"
+  shell: "bats --tap {{ item }} > {{ bats_output_dir }}/{{ item }}.tap"
```
ehelms: Forgot this change -- specifically do tap output and spit out to `.tap` file type.

ehelms:
Accompanying infra change for this -- https://github.com/theforeman/foreman-infra/pull/277

Vagrantfile
```diff
@@ -69,14 +69,12 @@ module Forklift
...
 
-        @host_vars = {box['name'] => box['ansible']['variables']}
-
         if (playbooks = box['ansible']['playbook'])
           [playbooks].flatten.each_with_index do |playbook, index|
             args = SUPPORT_NAMED_PROVISIONERS ? ["main#{index}", type: 'ansible'] : [:ansible]
             machine.vm.provision(*args) do |ansible|
               ansible.playbook = playbook
-              ansible.host_vars = @host_vars if @host_vars && Gem::Version.new(Vagrant::VERSION) >= Gem::Version.new('1.8.0')
+              ansible.extra_vars = box['ansible']['variables'] if @extra_vars && Gem::Version.new(Vagrant::VERSION) >= Gem::Version.new('1.8.0')
```
ekohl: Where does ```@extra_vars``` come from? Is that something Vagrant provides?

ekohl:
The bats part looks good.

Vagrantfile
ehelms: This was a typo leftover from a previous implementation, fixed now.

ehelms:
When I tested this it allowed me to use a proper array (which I noticed in blog examples people posted that they were using more complex data structures).

PR #376, opened 2017-02-02
Fix vagrant inventory --host command

ehelms:
Otherwise this will error out on unknown hosts and prevent spinning up playbooks that target localhost.

### DIFF
--- a/playbooks/inventory/vagrant.py
+++ b/playbooks/inventory/vagrant.py
@@ -6,7 +6,7 @@ import json
 import paramiko
 import subprocess
 import sys
-
+import os
 
 def parse_args():
     parser = argparse.ArgumentParser(description="Vagrant inventory script")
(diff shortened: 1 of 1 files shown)


ehelms:
This doesn't appear to be working still in some cases, digging more in to it.

playbooks/inventory/vagrant.py
```diff
@@ -40,7 +40,7 @@ def list_running_hosts():
 def get_host_details(host):
     cmd = "vagrant ssh-config {}".format(host)
     try:
-        saubprocess.check_call(cmd.split()).rstrip()
+        status = subprocess.check_output(cmd.split(), stderr=subprocess.STDOUT).rstrip()
```
ekohl: You can keep using check_call, but there's a typo in subprocess. Maybe only ```subprocess.CalledProcessError``` should be caught.

ehelms:
Turns out it was the list command causing the issues.

PR #377, opened 2017-02-02
Vagrant inventory list returns proper list

ehelms:


ehelms:
@ekohl ok this should be the last one editing this file

PR #378, opened 2017-02-02
Print test results to stdout but do not throw an error on failures

ehelms:


ehelms:
This will also help with CI system, so that the output files can be the source of whether or not the run failed at the testing stage.

ekohl:
Will this still fail with an exit code if some test failed? I think that's a useful feature to have.

ehelms:
I don't believe so, given that would trigger the Ansible run to fail. Which would leave us in a place of either having to do some sort of catching and trapping the exit code to be able to do further work to pull out the tests results or something else. Given this is a "test run", is it typical for it to exit code? e.g. if I run `rake tests` and they fail do they exit or just report failure?

ehelms:
I guess I could have it error code at the end from the test results. And have Jenkins just trap the error from a vagrant up so that we can still collect the logs.

ehelms:
@ekohl I went ahead and worked on an update to handle the cases discussed. This should now check the output codes at the very end.

PR #379, opened 2017-02-03
fix #367: Restructure directories into concerning technologies

### LINKED ISSUE #367: Re-organizing Directory Structure
Taking discussion from (https://github.com/Katello/forklift/pull/349#issuecomment-274358709) to have it be more pointed:

From @ekohl:
> I wouldn't mind if this lived in this repo, but we already have quite a few ways this repo can be used and we need to cut down on it and focus the documentation. As someone new to the repo it's hard to figure out what you can and can't do with it. Before we add more, we need to simplify it first. A few steps that I'd think of:

> # Split the top directory into subdirs

> Some candidates I see

> * bats
> * vagrant
> * openshift
> * docker

> # Remove deprecated things

> * Nuke setup.rb

From @ehelms:

> That sounds reasonable as a first good step to re-organize the top level (unless you think it should be part of this change only). The top level docker repository I don't know if anyone outside of myself even uses at this point. There is also a docker folder inside this PR, where in theory generic docker build directories are living. They could perhaps be moved to the top to help emphasize keeping them multi-use; I had put them there originally cause I was having to do some hacks to get them to work inside openshift.

> I am all for nuking setup.rb. I think the biggest hold back of it is the use in bats and thus the use in our CI system.

From @ekohl:

>> That sounds reasonable as a first good step to re-organize the top level (unless you think it should be part of this change only).
>No, I'd prefer to get that done first and then get this merged.

>>The top level docker repository I don't know if anyone outside of myself even uses at this point.
That's a big problem with this repo IMHO. We have a lot of possibilities and they should be maintained. >I'd like to go to a situation where each top level directory can be used independently.

>>There is also a docker folder inside this PR, where in theory generic docker build directories are living. They could perhaps be moved to the top to help emphasize keeping them multi-use; I had put them there originally cause I was having to do some hacks to get them to work inside openshift.
>That's fine, until they are generic they should live inside openshift.

>>I am all for nuking setup.rb. I think the biggest hold back of it is the use in bats and thus the use in our CI system.
>Agreed. Which is why I think we should get #331 merged and look at migrating the CI system.

ehelms:


ekohl:
I think this is correct, but a couple of things:
* We should communicate this because git won't move the content of ```user_playbooks```
* Maybe we need to update the docs as well.
* I'd like some more eyes on this than just us 2.

After FOSDEM I might have some time to look at this as well.

ehelms:
This needs a rebase of course and will keep changing as the repository changes. As @ekohl points out, it would be nice to get more opinions on this from other users.

PR #382, opened 2017-02-05
Add foreman_chaos role and functionality to devel

ehelms:
This idea is inspired by the Netflix Chaos Monkey idea, whereby we deploy a cron job that runs a script at random intervals and randomly brings down a service. The goal being to force developers to have to deal with services going down more often and build better resiliency into the application to handle these situations.

### DIFF
--- a/playbooks/devel.yml
+++ b/playbooks/devel.yml
@@ -16,3 +16,4 @@
     - foreman_repositories
     - katello_repositories
     - foreman_installer
+    - foreman_chaos
--- a/playbooks/roles/foreman_chaos/defaults/main.yml
+++ b/playbooks/roles/foreman_chaos/defaults/main.yml
@@ -1,0 +1,4 @@
+---
+foreman_chaos_minute: 0
+foreman_chaos_hour: "*"
+foreman_chaos_weekday: "*"
--- a/playbooks/roles/foreman_chaos/files/foreman-chaos
+++ b/playbooks/roles/foreman_chaos/files/foreman-chaos
@@ -1,0 +1,15 @@
+#!/usr/bin/env ruby
+
+sleep_time = Random.new.rand(60)
+puts "Sleeping for #{sleep_time} minutes"
+sleep(sleep_time)
+
+service_list = `katello-service list`.split("\n")
+service_list = service_list.map { |service| service.split("enabled")[0].strip }
+service_list = service_list.map { |service| service.gsub(".service", "") }
+
+random_service = service_list[Random.new.rand(service_list.length - 1)]
+
+puts "Stopping #{random_service}"
+system("service #{random_service} stop")
+
--- a/playbooks/roles/foreman_chaos/tasks/main.yml
+++ b/playbooks/roles/foreman_chaos/tasks/main.yml
@@ -1,0 +1,17 @@
+---
+- name: 'Deploy chaos script'
+  copy:
+    src: "{{ role_path }}/files/foreman-chaos"
+    dest: "/usr/sbin/foreman-chaos"
+    mode: "u=rwx,g=rx,o=r"
+
+- name: 'Create cronjob'
+  cron:
+    state: "present"
+    user: "root"
+    name: "foreman chaos"
+    minute: "{{ foreman_chaos_minute }}"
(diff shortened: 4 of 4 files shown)


ekohl:
I like the idea. I do wonder how it'll be when you really use it. One might be very annoyed if it happens in a long test but on the other hand it's only once an hour so that might be OK.

ehelms:
A few things we could do:

 a) adjust how often it runs (currently it triggers every hour, but picks a random sleep before executing to help randomize it more)
 b) have it check if a 'rake test' run is occurring on the system and skip

ekohl:
I was considering option b as well or a more generic approach to 'lock' it, but people might use it to disable it and forget to turn it back on. Another thing to think about is explicitly to randomly disable services during a `rake test` (`rake chaostest`?).

ehelms:
Another interesting aspects would be if this could put a service under stress somehow vs. completely killing it. For example, if we could find a way to push dynflow or Pulp memory consumption or run something that starts eating the CPU.

jlsherrill:
Adding this by default to the devel box seems like a huge annoyance with the current way everything is set up considering we really have no redundancy, monitor, or failover of backend services or subsystems.  I think it would only lead to annoyed and angry devs. 

I don't think we can really build better resiliency (to me meaning that the application continues to function) into the application without those three things.  Maybe if you could give some conrete examples of what you think this might accomplish it would help.

As of now I imagine it would just drive developers to delete this cron job :)

thomasmckay:
-1 I'd be removing this from my devel setup if it lands. I have enough trouble debugging things already without having to guess which thing was purposefully broken.

ehelms:
How do we know what kind of resiliency, monitoring and failover we need without encountering it and being driven to fix the issue? If annoyed and angry developers try to fix the problems, then I'd consider that a win :) But in all seriousness, but goal is to find ways that we can encounter the same problems our users do before they do to make the project better. We do not have the luxury of running our project at scale in a production way that causes us to drive to address these kinds of issues, we leave it up to our users to hit these and then we back pedal to fix.

If we make this optional, like most things, I imagine it wouldn't be run by anyone. Just like if we didn't make Katello optional in the default development setup, less people would run the broader stack IMO. But I'd be up for starting without it being default.

@thomasmckay you mention having to GUESS which thing was broken. I see that as a red flag, if you are having to guess which thing is broken, then shouldn't that drive you/us to make it so you aren't having to guess?

I realize this is a radical and/or crazy idea, but I'd prefer the goal be the focus less so the implementation. If the goal is not sound then the implementation does not matter.

jlsherrill:
> If annoyed and angry developers try to fix the problems, then I'd consider that a win :) 

I agree 100%, i just don't think this will cause them to 'fix' the problem.

jlsherrill:
In the netflix case, their goal (as i see it) was to ensure that if some service goes down, their infrastructure is able to handle and hopefully bring it back up properly.  So the goal is that the user never even notices when something bad is triggered.  This seems quite different than our use case where we have no such goal (that I'm aware of).  I don't know that we'd ever get to the point where a user wouldn't be expected to know (and fix) when a service went down.

If the goal is to improve messaging around down services, that's fine, lets write up some issues and get people to tackle that.  If you're expecting a developer to stop what they are doing in the middle of something else, investigate, think about and improve upon the current service handling i don't think that will ever happen.  There are about 100 things i encounter every day during normal development that i'd like to improve and I have the chance to fix maybe 1% of those.  If i worked on the other 99 i'd never get my main tasks done. 

I agree with what you're trying to do, i just don't think this will accomplish it.  I also think we know about certain things we can do to help with these things, we just haven't done them.

ekohl:
I think @jlsherrill worded a lot of my thoughts quite well. We need looser coupling in the application but I don't know if this will solve it. We need a common mindset on that but I don't know how to achieve that.

parthaa:
@ehelms I 'd be ok if we ran this in some thing like a "Test-Production" environment. May be create a "qe" role.  We could make it a rule/effort to run this environment prior to cutting a release candidate. As others have stated I feel there are several inconsistencies in our daily master check outs to not have to deal with this.

PR #386, opened 2017-02-09
Bats test promote using from environment instead of version

ehelms:
The error in the pipeline was caused by http://projects.theforeman.org/issues/18444 where in the update to the bats test we introduced a second promote that still used `version=1`


jlsherrill:
APJ

Issue #392, opened 2017-02-15
Unify user boxes and temporary boxes


ekohl:
From https://github.com/Katello/forklift/pull/390#issuecomment-280011862

Currently we have a ```boxes.yaml``` and ```.tmp_boxes/*.yaml``` which are loaded through the same logic. We can merge this into a ```boxes.d/*.yaml``` where both user boxes and temporary boxes live.

PR #393, opened 2017-02-15
Unify all boxes in boxes.d and various cleanups

ekohl:
Closes GH-392


docs/development.md
```diff
@@ -16,7 +16,7 @@ This covers how to setup and configure a development environment using the Forkl
 
 A Katello development environment can be deployed on CentOS 6 or 7. Ensure that you have followed the steps to setup Vagrant and the libvirt plugin. There are a variety of useful development environment options that should or can be set when creating a development box. These options are designed to configure your environment ready to use your own fork, and create pull requests. To create a development box:
 
-  1. Copy `boxes.yaml.example` to `boxes.yaml`. If you already have a `boxes.yaml`, you can copy the entries in `boxes.yaml.example` to your `boxes.yaml`.
+  1. Copy `boxes.d/99-local.yaml.example` to `boxes.d/99-local.yaml`. If you already have a `99-local.yaml`, you can copy the entries in `99-local.yaml.example` to your `99-local.yaml`.
```
ehelms: Maybe we add a note here about current users of boxes.yaml and what to do?

Vagrantfile
```diff
@@ -5,43 +5,38 @@ require 'yaml'
 require "#{File.dirname(__FILE__)}/lib/forklift"
 
 VAGRANTFILE_API_VERSION = '2'
-SUPPORT_SSH_INSERT_KEY = Gem.loaded_specs['vagrant'].version >= Gem::Version.create('1.7')
-SUPPORT_NAMED_PROVISIONERS = Gem.loaded_specs['vagrant'].version >= Gem::Version.create('1.7')
-SUPPORT_BOX_CHECK_UPDATE = Gem.loaded_specs['vagrant'].version >= Gem::Version.create('1.5')
```
ehelms: Assuming this means we are enforcing a minimum Vagrant version? If yes, we should probably check for that and error out here. The vagrant on our CI system is quite old IIRC and will break.

ehelms:
This needs a rebase and update.

Vagrantfile
ekohl: I believe we already have a requirement on Version 1.8, at least according to the README. Should we enforce this in the Vagrantfile?

docs/development.md
ekohl: It's automatically migrated. The only thing is that if you're switching between branches after it's been merged then you have some issues.

Vagrantfile
ehelms: We have a requirement for local usage, but continue to maintain compability due to the CI system as far as I know. But we need to double check before we remove any of this.

ekohl: I recall that Vagrant 1.8 was needed for Ansible variables. Since the pipelines now use Ansible variables as well, I reasoned that it was a global hard require.

ehelms:
I have some changes I'd like to make to make re-use of Forklift and its roles easier that would need to built on top of this due to the re-factoring. I went ahead and opened a PR to show the re-factoring -- https://github.com/theforeman/forklift/pull/426

ehelms:
@ekohl I kinda broke this with my refactoring but I like the change if you can revisit it.

ehelms:
@ekohl planning to revisit this change?

ekohl:
Intention would currently be the better word :)

ekohl:
I finally took the time to revisit this. I've broken some APIs by renaming functions but I do think it's cleaner now. It can still use a proper review and maybe every plugin needs a boxes.d directory too.

Longer term I want to split the base boxes into separate files so you can easily disable a group of servers meaning the vagrant status won't show a huge list of servers. This in preparation of adding Ubuntu too.

roles/forklift/tasks/main.yml
```diff
@@ -9,16 +9,34 @@
 
 - name: 'Check Forklift state'
   fail:
-    msg: 'The forklift_state variable is undefined. Please use pass "-e "forklift_state=up"" to spin up the boxes or "-e "forklift_state=destroy"" to destroy the boxes'
-  when: forklift_state is undefined
+    msg: 'The invalid value for forklift_state variable. Please use pass "-e "forklift_state=up"" to spin up the boxes or "-e "forklift_state=destroy"" to destroy the boxes'
```
ehelms: The wording here in the first sentence is a little hard to parse.

roles/forklift/tasks/main.yml
```diff
@@ -9,16 +9,34 @@
...
+  when: forklift_state not in ('up', 'destroy')
 
-- name: 'Bring boxes up'
-  include_tasks: 'up.yml'
```
ehelms: I don't see this file removed.

roles/forklift/tasks/main.yml
```diff
@@ -9,16 +9,34 @@
...
+
+- name: 'Bring up boxes'
+  command: "vagrant up {{ item.key }}"
+  args:
+    chdir: "{{ current_directory }}"
+  with_dict: "{{ forklift_boxes }}"
+  when: forklift_state == "up"
 
 - name: 'Destroy boxes'
-  include_tasks: 'destroy.yml'
```
ehelms: I don't see this file removed.

roles/forklift/tasks/main.yml
```diff
@@ -9,16 +9,34 @@
...
-  when: forklift_state == 'destroy'
+  command: "vagrant destroy {{ item.key }}"
```
ehelms: I wonder if for this and `up`, we should instead do a single `vagrant destroy` with each box space separated? That might be more efficient than calling `vagrant` multiple times.

ekohl: Correct - I implemented that in f25188a9a258448a68a29b56c3662faf1c64625d and then I did a bad rebase on that.

roles/forklift/tasks/main.yml
ekohl: Not sure what I was thinking - probably not all.

ehelms:
Looks good -- rubocop issues need addressing.

ekohl:
It was rubocop clean the last time I submitted it :(

ehelms:
Merged -- @ekohl mind sending an email to -dev to let Forklift users know that on update boxes.yaml they have grown used to will be in a new place?

ekohl:
Done

akofink:
Very nice @ekohl! Much appreciated!

PR #394, opened 2017-02-15
Check when to use --disable-system-checks

ehelms:


waldenraines:
These changes worked for me when doing my very limited testing of `vagrant up centos7-katello-3.1` without any customization.  Someone else who is more versed in this should probably have a look for final approval though.

PR #396, opened 2017-02-16
Update production.md

lzap:
The yaml file is no longer with suffix, I suppose. Also the localhost uses ssh while using local module calls directly, from what I have learned. Both works.

### DIFF
--- a/docs/production.md
+++ b/docs/production.md
@@ -73,7 +73,7 @@ This use case assumes that you are on a box that has SSH access to the target bo
 
 For a release version in production:
 
-    ansible-playbook -l <target-host> playbooks/katello_31.yml
+    ansible-playbook -l <target-host> playbooks/katello.yml
 
 For nightly production:
 
@@ -82,7 +82,7 @@ For nightly production:
 
 After installing a Katello server, you could then spin up a Capsule with the assumption the Katello server can talk to the Capsule and vice versa.
 
-    ansible-playbook -l <target-capsule-host> playbooks/capsule_31.yml
+    ansible-playbook -l <target-capsule-host> playbooks/capsule.yml
 
 ### Local Deployment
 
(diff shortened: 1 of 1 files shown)


docs/production.md
```diff
@@ -93,8 +93,8 @@ After installing a Katello server, you could then spin up a Capsule with the ass
 
 For a release version in production:
 
-    ansible-playbook -l localhost playbooks/katello_31.yml
+    ansible-playbook -i "localhost," -c local playbooks/katello.yml
 
 For nightly production:
 
-    ansible-playbook -l localhost playbooks/katello_nightly.yml
+    ansible-playbook -i "localhost," -c local playbooks/katello_nightly.yml
```
ekohl: This playbook no longer exists either. Nightly is the default for ```playbooks/katello.yml``` if unspecified.

docs/production.md
```diff
@@ -73,7 +73,7 @@ This use case assumes that you are on a box that has SSH access to the target bo
 
 For a release version in production:
 
-    ansible-playbook -l <target-host> playbooks/katello_31.yml
+    ansible-playbook -l <target-host> playbooks/katello.yml
```
ekohl: We now use variables to determine the version. This means you have to set ```foreman_version``` to ```1.14``` and ```katello_version``` to ```3.3```. In the vagrant solution we get this mapping from ```config/versions.yaml``` but ansible-playbook can't read that so you have to specify this as ```-e foreman_version=1.14 -e katello_version=3.3```.

lzap:
What exactly I need to change @ekohl ? I am not sure, new to Ansible.

docs/production.md
ekohl: Here you could simply use ```playbooks/katello.yml``` instead of ```playbooks/katello_nightly.yml```.

docs/production.md
ekohl: This should be ```ansible-playbook -l <target-host> playbooks/katello.yml -e foreman_repositories_version=1.14 -e katello_repositories_version=3.3```

PR #397, opened 2017-02-16
Roles for seeding data

tstrachota:
- users, puppet modules and hosts
- all tasks tagged as `seed`


thomasmckay:
Consider collaborating on a PR like this? https://github.com/theforeman/forklift/pull/400

playbooks/roles/data/puppet_modules/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: 'Install puppet modules'
+  shell: puppet module install {{ item }}
+  with_items: '{{ data_puppet_modules }}'
+  tags:
+    - seed
+
+- name: 'Import puppet classes'
+  shell: >
+    hammer proxy import-classes --name '{{ data_puppet_modules_proxy_name }}'
```
xprazak2: no user and password => 401 Unauthorized

xprazak2: perhaps we could have a role that would set up hammer credentials for us? It would solve [this comment](https://github.com/theforeman/forklift/pull/398#discussion_r102865379) as well.

tstrachota: Good idea, I'll add that

tstrachota:
Credentials from hammer commands removed. Roles now expect it to be preconfigured (for example using https://github.com/theforeman/forklift/pull/399/commits/f68795584990955781258658a34404fdc1ebce71)

tstrachota:
@ekohl `foreman_seeder` is interesting idea. I wanted to create a tool that would seed foreman with hosts with some dummy reports and trends. Mainly for demo purposes to make our demo HW easily re-installable. Plugin with rake tasks could be good solution.

iNecas:
When talking about the seeding, I would recommend thinking about adding this rake scripts (or whatever it would be) directly to the foreman, so that it's easy for the plugin developers to extend it with their set of seed data variants.

xprazak2:
seeding hostgroups fails with:
```
failed: [hostname] (item={u'major': u'16', u'medium': u'Ubuntu mirror', u'name': u'Ubuntu', u'family': u'Debian', u'release_name': u'xenial', u'minor': u'04'}) => {"changed": true, "cmd": "    hammer --output=silent hostgroup info --name \"Forklift Ubuntu 16.04\" || hammer hostgroup create --name 'Forklift Ubuntu 16.04' --operatingsystem 'Ubuntu 16.04' --medium 'Ubuntu mirror' --partition-table 'Preseed default' --parent 'Forklift Base' --organizations 'Default Organization' --locations 'Default Location'", "delta": "0:00:03.839874", "end": "2017-05-31 05:34:09.887563", "failed": true, "item": {"family": "Debian", "major": "16", "medium": "Ubuntu mirror", "minor": "04", "name": "Ubuntu", "release_name": "xenial"}, "rc": 70, "start": "2017-05-31 05:34:06.047689", "stderr": "Could not create the hostgroup:\n  Error: hostgroup not found", "stdout": "", "stdout_lines": [], "warnings": []}
```

because `Forklift Base` is not seeded and it is used as parent.

Using rake tasks for seeding data is an interesting idea and it could be very helpful. But regardless whether they will be in core or `foreman_seeder` plugin, I do not see them coming any time soon. I would not want to reject this in favour of something we haven't even started work on.

ehelms:
What about changing the role name to `foreman_seeder` ? For me, `data` is too generic and does not provide a safe namespace.

tstrachota:
Updated. I renamed the role directory to `foreman_seeder`, split roles for host and hostgroups and added a task for creating a parent hostgroup when it doesn't exist/

xprazak2:
seeding hostgroups works nicely now, but seeding hosts fails with:
```
failed: [10.16.184.33] (item={u'name': u'host1', u'hostgroup': u'Forklift CentOS 7'}) => {"changed": true, "cmd": "hammer host create --hostgroup 'Forklift CentOS 7'  --name 'host1' --mac $(echo -n 52:54:00:; openssl rand -hex 3 | sed 's/\\(..\\)/\\1:/g; s/.$//')", "delta": "0:00:04.474990", "end": "2017-06-07 10:13:37.744758", "failed": true, "item": {"hostgroup": "Forklift CentOS 7", "name": "host1"}, "rc": 65, "start": "2017-06-07 10:13:33.269768", "stderr": "Could not create the host:\n  Domain can't be blank\n  Interfaces.domain can't find domain with this id\n  Architecture can't be blank\n  Root password should be 8 characters or more\n  Root password should not be blank - consider setting a global or host group default", "stdout": "", "stdout_lines": [], "warnings": []}
```

ekohl:
FYI: A very much proof of concept is at https://github.com/theforeman/foreman/pull/4576. I'd be interested in some feedback.

tstrachota:
@xprazak2 yeah, I know about that failure. It's actually why I split the host and hostgroup roles. The failure occurs when you don't have the parent hostgroup fully set up.

The "Forklift Base" is created in foreman_provisioning role: https://github.com/theforeman/forklift/pull/399/files#diff-30428ab296704aa9a29131abf7a67c13R187
I didn't want to duplicate the setup code so when `data/hostgroups` is used without `foreman_provisioning` the role creates empty parent hostgroup. That's why `data/hosts` fails.

I know it's confusing for forklift users when the default role setup fails but I'm not sure if there's some easy solution for that. I'm fine if we decide to remove the hosts role for now.

xprazak2:
I think just adding comment to the role that it requires "Forklift Base" hostgroup and you can use foreman_provisioning to set it up should be enough, no need to remove it. I think we are good here, I have no further comments.

tstrachota:
Thanks @xprazak2 I added a comment about hostgroup setup to `foreman_seeder/hosts/defaults/main.yml`

xprazak2:
Nice, I think this is ready to merge.

PR #398, opened 2017-02-16
Roles for installing plugins

tstrachota:
- plugins:
  - foreman_ansible
  - foreman_discovery
  - foreman_openscap
  - foreman_remote_execution
  - foreman_templates
- tasks tagged as `installation`, `env_setup`, `seed`


playbooks/roles/plugins/templates/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+sync_templates: False
```
ehelms: This should probably be renamed to `templates_sync` to follow convention for variable naming.

playbooks/roles/plugins/remote_execution/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+- name: 'Install remote execution plugin'
+  shell: >
+    foreman-installer
+    --enable-foreman-plugin-remote-execution
+    --enable-foreman-proxy-plugin-remote-execution-ssh
```
ehelms: General comment, one nice thing about the foreman_installer role is you can pass any options to it, for example if I wanted to be able to configure any of the options for a plugin.

playbooks/roles/plugins/openscap/tasks/main.yml
```diff
@@ -0,0 +1,48 @@
...
+
+- name: 'Install foreman openscap puppet module'
+  yum:
+    name: 'puppet-foreman_scap_client'
+    state: 'present'
+  tags:
+    - seed
+
+- name: 'restart foreman-proxy'
+  service: name=foreman-proxy state=restarted
```
ehelms: Please use expanded format here to keep consistent.

playbooks/roles/plugins/openscap/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+openscap_hammer: hammer -u admin -p changeme
```
ehelms: This isn't too robust against the admin password getting configured to something else.

tstrachota:
Updated: 
- I removed credentials from hammer commands. Roles now expect it to be preconfigured (for example using https://github.com/theforeman/forklift/pull/399/commits/f68795584990955781258658a34404fdc1ebce71)
- foreman-installer occurrences in plugin tasks got variable for setting additional parameters

xprazak2:
foreman_ansible role works nicely. Since we install the roles with ansible-galaxy, maybe we could call the [import endpoint](https://github.com/theforeman/foreman_ansible/blob/master/app/controllers/api/v2/ansible_roles_controller.rb#L29) to get them imported into foreman as well?

xprazak2:
for foreman_docker role, I got:
```
TASK [plugins/docker : Create docker compute resource] *************************
Wednesday 31 May 2017  10:31:49 +0200 (0:00:02.801)       0:00:22.489 ********* 
fatal: [hostname]: FAILED! => {"changed": true, "cmd": "hammer --output=silent compute-resource info --name \"docker\" || hammer compute-resource create --name docker --url 'unix://var/run/docker.sock' --provider Docker ", "delta": "0:00:01.824357", "end": "2017-05-31 04:31:53.708066", "failed": true, "rc": 70, "start": "2017-05-31 04:31:51.883709", "stderr": "Could not create the compute resource:\n  ERF42-0410 [Foreman::Exception]: unknown provider", "stdout": "", "stdout_lines": [], "warnings": []}
```

I think you need to call installer with `--enable-foreman-plugin-docker`

playbooks/roles/plugins/ansible/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    name: ansible
+    state: present
+  tags:
+    - installation
+
+- name: 'Install ansible plugin'
+  shell: >
+    foreman-installer
+    --enable-foreman-plugin-ansible
+    --enable-foreman-proxy-plugin-ansible
```
ehelms: On most Katello installations this will likely fail out with no re-course due to the system checks such as requiring 8GB of memory meant to protect users in production but not absolutely required for dev or testing.

playbooks/roles/plugins/docker/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
+- name: 'Prepare hammer taxonomy options'
+  set_fact:
+    docker_hammer_taxonomy_params: "--organizations '{{ docker_organization }}' --locations '{{ docker_location }}'"
+  when: docker_use_taxonomies
```
ehelms: When would docker not use taxonomies?

playbooks/roles/plugins/docker/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+- name: "Install foreman docker plugin"
+  yum:
+    name: 'tfm-rubygem-foreman_docker'
+    state: 'present'
+  tags:
+    - installation
+
+- name: "Install docker"
+  yum:
+    name: 'docker'
```
ehelms: What about if using docker-engine?

ehelms:
Sorry it has taken so long to get around to this in earnest. I think this is looking good with some comments above. I have a general question. The names of the roles do not include the true name of the plugin itself. For example, here it is `docker` instead of `foreman_docker`. I think this:

 a) can add some confusion about what someone is looking at
 b) could cause overlapping namespaces with variables and the role itself if using a role for actual docker, or actual openscap or actual ansible in conjunction

playbooks/roles/plugins/docker/tasks/main.yml
tstrachota: I'm afraid I don't understand. What exactly is docker-engine?

playbooks/roles/plugins/docker/tasks/main.yml
tstrachota: I checked with Daniel, foreman_docker can run without taxonomies.

tstrachota:
@ehelms @xprazak2 updated.

I renamed roles to full names of the plugins and fixed your comments.

The issue with `--disable-system-checks` is solved only partially. Each installer task in the roles has a variable for additional params. When roles are used with katello, one needs to add `--disable-system-checks` param manually. It's rather a workarnound than a great solution. Do you have any ideas for improvements?

I've been thinking about including `foreman_installer_params_internal_use_only` variable if it's defined (this won't cover the situation when running a single plugin role against an already installed server) or checking for presence of `/etc/foreman/plugins/katello.yaml` and disable system checks only for katello installations.

xprazak2:
My previous comments addressed - ansible roles are imported and foreman_docker now installs, but docker service is stopped. Maybe we could add a task with `systemctl start docker` and `systemctl enable docker` into foreman_docker role.

I do not have my own suggestions regarding the `--disable-system-checks`, seems like b) or c) could be a way to go.

tstrachota:
@xprazak2 rebased and updated. Docker service is now started and enabled in the role.

I also added a plugin installer role that detects katello installations and adds `--disable-system-checks` for consideration. We can get rid of that commit if you don't like it.

xprazak2:
I do not mind having a role that disables system checks. I have no additional comments, everything seems to work now 👍

tstrachota:
@ehelms or whoever has merge rights, may I ask you to merge before the branch diverges from the master?

playbooks/roles/plugins/foreman_docker/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+    foreman_docker_hammer_taxonomy_params: "--organizations '{{ foreman_docker_organization }}' --locations '{{ foreman_docker_location }}'"
+  when: foreman_docker_use_taxonomies
+
+- name: 'Prepare blank hammer taxonomy options'
+  set_fact:
+    foreman_docker_hammer_taxonomy_params: ""
+  when: not foreman_docker_use_taxonomies
+
+- name: "Install foreman docker plugin"
+  yum:
```
ekohl: Doesn't the installer have an option for this? There's at least a class for it in puppet-foreman.

playbooks/roles/plugins/foreman_docker/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+
+- name: "Configure docker"
+  lineinfile:
+    dest: /etc/sysconfig/docker
+    regexp: '^OPTIONS='
+    line: "OPTIONS='--selinux-enabled -G docker'"
+  tags:
+    - env_setup
+
+- name: 'Restart Foreman to load the new compute provider type'
```
ekohl: I'd expect the installer to take care of this when that's used.

playbooks/roles/plugins/foreman_openscap/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+  vars:
+    plugin_installer_options: >
+      --enable-foreman-plugin-openscap
+      --enable-foreman-proxy-plugin-openscap
+      {{ foreman_openscap_installer_options }}
+  tags:
+    - installation
+
+- name: 'Install foreman openscap hammer plugin'
+  yum:
```
ekohl: The installer should be able to install this as well.

playbooks/roles/plugins/foreman_openscap/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+    - installation
+
+- name: 'Install foreman openscap puppet module'
+  yum:
+    name: 'puppet-foreman_scap_client'
+    state: 'present'
+  tags:
+    - seed
+
+- name: 'restart foreman-proxy'
```
ekohl: I'd expect that the installer takes care of this.

playbooks/roles/plugins/foreman_remote_execution/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+- name: 'Install emote execution plugin'
```
ekohl: s/emote/remote/

playbooks/roles/plugins/foreman_templates/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
...
+  tags:
+    - installation
+
+- name: 'Fetch community templates'
+  shell: >
+    foreman-rake templates:sync {{ foreman_templates_sync_params }}
+  when: foreman_templates_sync
+  tags:
+    - seed
+
```
ekohl: Nit: can you remove this empty line?

playbooks/roles/plugins/foreman_docker/tasks/main.yml
tstrachota: I don't think we publish options for installing foreman-docker in the installer. `foreman-installer -h | grep docker` is silent.

playbooks/roles/plugins/foreman_docker/tasks/main.yml
tstrachota: I agree, if it was in the installer it would take care of reloading it. But as I mentioned in the previous comment, there's no option for installing foreman-docker.

playbooks/roles/plugins/foreman_openscap/tasks/main.yml
tstrachota: It doesn't install the hammer plugin for me. It might be an installer issue but I wouldn't block this PR until we fix it.

ehelms:
@ekohl Can you re-review when you get an opportunity?

playbooks/roles/plugins/foreman_docker/tasks/main.yml
ekohl: That could be. @ehelms do you know if this is a conscious decision not to expose those options? foreman-installer does expose them.

playbooks/roles/plugins/foreman_openscap/tasks/main.yml
ekohl: Right, it's an additional class that needs an explicit option and only foreman-installer exposes that option.

playbooks/roles/plugins/foreman_docker/tasks/main.yml
ares: I'm not sure about the reasoning, but to introduce it, adding one line to [this file](https://github.com/Katello/katello-installer/blob/master/config/katello-answers.yaml#L74) and a migration for previous installations would enabled it.

tstrachota:
@ekohl @ehelms I updated the PR according to the latest comments. There's difference between forman and katello installer in what options they publish, which prevented me from using installer for complete docker and openscap plugin installation.
I agree it deserves fixing in the installer, but I would say it's an independent issue that shouldn't block this PR.

@ekohl may I ask you to re-review?

ehelms:
@tstrachota If you don;t mind, I just merged a re-org PR that moves the roles directory to the top, adjusting this PR to put your changes there would be appreciated (I also don't mind moving after the fact but I think the roles path will be broken.

https://github.com/theforeman/forklift/pull/476

tstrachota:
@ehelms rebased, roles moved to top level

PR #400, opened 2017-02-16
seed with hammer csv

thomasmckay:
An example of seeding with hammer csv. 

- [ ] install hammer locally on server
- [ ] add a readme describing usage and variables
- [x] customize csv file location
- [x] if csv resource type file does not exist, don't run (skip some resources)
- [x] if csv resource type file already exists, don't overwrite (allow other roles to put in custom files)
- [x] capture stderr and stdout to logs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 *.swn
 katello-installer*
 foreman-installer*
-puppet-*
 boxes.yaml
 .idea
 scripts
--- a/playbooks/roles/csv/tasks/main.yml
+++ b/playbooks/roles/csv/tasks/main.yml
@@ -5,17 +5,8 @@
     state: directory
     owner: vagrant
 
-- name: Create operating-systems.csv
-  template:
-    src: templates/operating-systems.csv.j2
-    dest: /home/vagrant/csv/operating-systems.csv
-    owner: vagrant
-
-- name: Run hammer csv operating-systems
-  shell: >
-    BUNDLE_GEMFILE=~/hammer-cli-katello/Gemfile bundle exec {{ csv_hammer }}
-    csv operating-systems
-    --verbose
-    --file /home/vagrant/csv/operating-systems.csv
-    > /home/vagrant/csv/operating-systems.log
-
+- name: Import resources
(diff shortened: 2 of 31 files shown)


ekohl:
Would it be possible to seed through json instead of csv? I think that would make is easier to reuse the logic for other types.

thomasmckay:
@ekohl The "logic" to import is all written in hammer csv. I suspect you could add some ansible magic to take in a json and convert it into the "with_items" line that would then generate the csv for import.

I didn't include the other resource types but they are listed here https://github.com/Katello/hammer-cli-csv/blob/master/lib/hammer_cli_csv.rb . They are not all fully supported or working. Maybe hooking into forklift would provide an incentive for making it more robust for all the other resource types.

thomasmckay:
I'll add the rest of the resource types to this PR for completeness.

playbooks/roles/csv/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+    owner: vagrant
+
+- name: Run hammer csv operating-systems
+  shell: >
+    BUNDLE_GEMFILE=~/hammer-cli-katello/Gemfile bundle exec {{ csv_hammer }}
+    csv operating-systems
+    --verbose
+    --file /home/vagrant/csv/operating-systems.csv
+    > /home/vagrant/csv/operating-systems.log
+
```
thomasmckay: Is there some way to put the above two sections into a loop that iterates over all the resource types? Would be better than cut&paste of this section.

.gitignore
```diff
@@ -4,7 +4,6 @@
 *.swn
 katello-installer*
 foreman-installer*
-puppet-*
```
thomasmckay: @ehelms This line matched one of my file templates; is there a more specific pattern to ignore? Or I could force not ignoring my file.

ehelms: I think this removal is fine now.

ehelms:
@thomasmckay Hallo! This PR recently celebrated it's birthday. Do you plan to fix it up or should we close it out?

PR #401, opened 2017-02-17
Add topologies to spin up combinations of infrastructure

ehelms:
Floating this idea out there of pre-defined (and as a blueprint for creating "bigger" setups) topologies that spin up multiple concerns for users. While this is similar to the pipelines, they are designed to run scenarios with testing. The idea with topologies would be to spin up sets for the intent of tracking down a bug or creating more encompassing development setups.

The devel topology is still being tested. This would provide the following to start:

nightly
  - server
  - proxy
  - client (registered through proxy)

devel
  - devel server
  - proxy
  - client (registered through proxy)

### DIFF
--- a/docs/advanced_uses.md
+++ b/docs/advanced_uses.md
@@ -1,0 +1,91 @@
+# Advanced Uses
+
+This section covers advanced ways that you can use what Forklift provides.
+
+## Topologies
+
+Similar to the way the pipelines are built for testing, a collection of systems can be defined as well as what goes on to each system in a single playbook. You can then ask Ansible to run and bring up each system with what you need on it and potentially connected to one another. Let's look at a concrete example. The basic multi-system setup is a server running Foreman and Katello, an external Foreman Proxy and a client machine configured through the Foreman Proxy for isolation. Thus, we need three systems with the Proxy talking to the server and the client talking to the proxy. We can achieve this with the following playbook:
+
+```
+- hosts: localhost
+  vars:
+    forklift_name: nightly-server-proxy-client
+    forklift_boxes:
+      server-nightly-centos7:
+        box: centos7
+        memory: 4680
+      proxy-nightly-centos7:
+        box: centos7
+        memory: 3072
+      client-nightly-centos7:
+        box: centos7
+        memory: 3072
+  roles:
+    - forklift
+
+- hosts: server-nightly-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 4
+    katello_repositories_version: nightly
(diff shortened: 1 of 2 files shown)


ekohl:
Looks like there are at least some fixes in here that we can already merge. Other than that I wonder what's the intent of this. Are they supposed to be longer lived boxes?

ehelms:
They could be longer lived but, I am looking at more like deploying somewhat more realistic user deployments. There are things that occur when you use an external proxy that having server + proxy + client up and running could potentially help to uncover.

I could turn this instead into an example to show what's possible with the Forklift role outside of just creating testing pipeline playbooks.

ekohl:
That makes total sense. I do wonder if we could use ansible groups in a flexible way so you can combine your own boxes into a topology. So more a layer on top of what we have now to orchestrate than redefine the entire existing layer as well.

ehelms:
I split out the client related fixes to here -- https://github.com/theforeman/forklift/pull/410

ehelms:
This PR has in turn been updated to add topologies as a documentation effort.

@ekohl I think your suggestion makes sense if we can figure it out. I'm inclined to simply document what's possible for now until there is more demand to support things directly.

ekohl:
There's a merge conflict now, but it makes sense to merge #410 first and then rebase.

ehelms:
Rebased

ehelms:
@ekohl please to take a look

docs/advanced_uses.md
```diff
@@ -0,0 +1,91 @@
...
+  become: yes
+  vars:
+    puppet_repositories_version: 4
+    foreman_proxy_content_server: server-nightly-centos7
+    katello_repositories_version: nightly
+    foreman_repositories_version: nightly
+    katello_repositories_use_koji: true
+    foreman_installer_scenario: foreman-proxy-content
+    foreman_installer_options_internal_use_only:
+      - '--disable-system-checks
```
ekohl: If you're using the internal options, this can be an array

docs/advanced_uses.md
```diff
@@ -0,0 +1,91 @@
...
+    - puppet_repositories
+    - foreman_repositories
+    - katello_repositories
+    - foreman_proxy_content
+    - foreman_installer
+
+- hosts: client-nightly-centos7
+  become: yes
+  vars:
+    katello_client_server: "proxy-nightly-centos7.example.com"
```
ekohl: Can this be ```{{ proxy_fqdn.stdout }}``` or does it have to be hardcoded?

PR #403, opened 2017-02-20
fix line scrolling for important options

thomasmckay:


thomasmckay:
No merge permissions, if someone could do the honors.

PR #405, opened 2017-02-21
Various improvements

ekohl:
* Enable CNAME record testing on bind with 1.15+
* Ensure git is installed when running bats
* Do not force a cache update when installing foreman-release-scl

### DIFF
--- a/playbooks/roles/bats/tasks/main.yml
+++ b/playbooks/roles/bats/tasks/main.yml
@@ -1,4 +1,9 @@
 ---
+- name: "Install git"
+  yum:
+    name: "git"
+    state: "installed"
+
 - name: "Get /usr/bin/bats stat"
   stat:
     path: "/usr/bin/bats"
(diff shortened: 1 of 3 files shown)


playbooks/roles/foreman_installer/tasks/packages.yml
```diff
@@ -8,7 +8,6 @@
 - name: 'Install foreman-release-scl'
   yum:
     name: foreman-release-scl
-    update_cache: yes
```
ehelms: Given `foreman-release-scl` lays down new repositories, are we positive an update to the cache isn't needed to ensure yum picks up the new repositories?

playbooks/roles/bats/tasks/main.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: "Install git"
+  yum:
+    name: "git"
+    state: "installed"
```
ehelms: :+1:

We need to update this role to allow testing updates to bats from a PR or local environment. I'll file an issue.

playbooks/roles/foreman_installer/tasks/packages.yml
ekohl: My experience with yum is that it does this automatically for missing caches.

PR #408, opened 2017-02-21
Fixing issue with disabling ipv6 while pulp is running

daviddavis:
We're seeing workers die because they can't connect to qpid. Seems to be
caused by disabling ipv6 while pulp is running. Making a change to
disable ipv6 before pulp is installed.

More information: https://pulp.plan.io/issues/2586#note-5

### DIFF
--- a/bats/fb-install-katello.bats
+++ b/bats/fb-install-katello.bats
@@ -22,10 +22,6 @@ setup() {
     fi
   fi
 
-  #disable ipv6 for issue #12386
-  sysctl -w net.ipv6.conf.default.disable_ipv6=1
-  sysctl -w net.ipv6.conf.all.disable_ipv6=1
-
   tPackageExists curl || tPackageInstall curl
   if tIsRedHatCompatible; then
     tPackageExists yum-utils || tPackageInstall yum-utils
--- a/pipelines/pipeline_katello_nightly.yml
+++ b/pipelines/pipeline_katello_nightly.yml
@@ -26,6 +26,7 @@
   roles:
     - selinux
     - etc_hosts
+    - disable_ipv6
     - epel_repositories
     - puppet_repositories
     - foreman_repositories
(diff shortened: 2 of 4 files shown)


ehelms:
Seems OK to me -- @jlsherrill ?

ekohl:
I'd like to maintain the comment with the issue number because clearly this is a workaround rather than a solution. IMHO the real problem is that the client doesn't fallback to IPv4 when IPv6 is unavailable.

daviddavis:
What do you mean by client? Pulp or qdrouterd? For pulp, I am not sure I would expect a service that relies on a web connections to work if I disable ipv6 without restarting the service.

ekohl:
I would expect that if pulp is configured to connect to a hostname that has both IPv6 and IPv4 then it should fall back to IPv4 if IPv6 is unavailable.

daviddavis:
It does fall back to ipv4 if ipv6 is unavailable. The issue is ipv6 becoming unavailable while it's connected using ipv6.

ekohl:
That implies there is no proper reconnect functionality in place.

Note that I'm not opposed to this change, but we shouldn't mask the real issue in our pipelines. At the very least we should refer to the issue on redmine why we do it.

daviddavis:
I'm not an expert on the pulp qpid stuff so I don't know what sort of reconnect functionality exists in pulp or the qpid python package. In fact, it may not exist at all. 

I moved the comment though into the disable_ipv6 role yaml but the disabling of ipv6 has nothing to do with pulp AFAICT. So I can also add another comment about why we disable ipv6 when we provision the box (instead of when running bats) if need be.

ekohl:
Please do. I think IPv6 should never be disabled on a system level and any time you do, it needs a good (documented) reason.

ehelms:
@ekohl Can you double check the update? I'd like to run a 3.3 test and need this change unfortunately.

daviddavis:
I expanded the comment. Let me know if it's insufficient.

Issue #409, opened 2017-02-22
Allow bats role to test local changes or a PR


ehelms:

ekohl:
To test local changes you can use:
```yaml
vars:                        
  bats_forklift_dir: /vagrant
  bats_update_forklift: "no" 
```

If you don't use NFS then you do need to use ```vagrant rsync $box``` but it works quite well. I use it in https://github.com/theforeman/forklift/pull/406 as well.

evgeni:
What @ekohl says, or the use of a different repo/branch as implemented in https://github.com/theforeman/forklift/pull/464

evgeni:
@ehelms I think this can be closed now?

PR #420, opened 2017-03-08
Allow use of local bats for testing

ehelms:


playbooks/roles/bats/defaults/main.yaml
```diff
@@ -3,6 +3,7 @@ bats_git_dir: "/root/bats"
 bats_forklift_dir: "/root/forklift"
 bats_output_dir: "/root/bats_results"
 bats_update_forklift: "yes"
+bats_use_local_forklift: False
```
ekohl: Maybe ```bats_git_clone_forklift```?

ehelms:
I updated this based on https://github.com/theforeman/forklift/issues/409 to indicate how to do this in the docs.

PR #423, opened 2017-03-23
Adds mention of vagrant libvirt plugin

johnpmitsch:
Its not obvious to vagrant noobs that you need a libvirt provider. Seeing as a fair amount of forklift users are using libvirt, I added a mention of it in the Requirements


johnpmitsch:
@ekohl updated, thanks!

PR #426, opened 2017-04-02
Use Forklift as a library

ehelms:


ehelms:
See (https://github.com/ehelms/forklift/blob/57b68b37c5ae79646221229f821037363fda0cc2/docs/library.md) for how I envisioned using these changes.

lib/forklift/box_distributor.rb
```diff
@@ -0,0 +1,163 @@
...
+      @ansible_groups[ansible['group'].to_s] << box.fetch('name')
+
+      if ansible.key?('server')
+        @ansible_groups["server-#{box_name}"] = ansible['server']
+      end
+
+      return unless (playbooks = ansible['playbook'])
+
+      [playbooks].flatten.each_with_index do |playbook, index|
+        args = SUPPORT_NAMED_PROVISIONERS ? ["main#{index}", type: 'ansible'] : [:ansible]
```
ekohl: I think ```SUPPORT_NAMED_PROVISIONERS``` was removed in the refactor.

lib/forklift/box_distributor.rb
```diff
@@ -0,0 +1,163 @@
...
+      machine.vm.provision :shell do |shell|
+        shell.inline = box.fetch('shell')
+        shell.privileged = false if box.key?('privileged')
+      end
+    end
+
+    def configure_libvirt(machine, box, networks)
+      machine.vm.provider :libvirt do |p, override|
+        override.vm.box_url = box.fetch('libvirt') if box.fetch('libvirt', false)
+        override.vm.synced_folder '.', '/vagrant', type: 'rsync', rsync__args: ['--max-size=100m']
```
ekohl: I'd like this type to be easily overrideable. Especially with bats testing an nfs mount might be much nicer and those using cacher already have it set up.

lib/forklift/box_distributor.rb
```diff
@@ -0,0 +1,163 @@
...
+      config.hostmanager.manage_guest = true
+      config.hostmanager.include_offline = true
+    end
+
+    def configure_vagrant_cachier(config)
+      return unless Vagrant.has_plugin?('vagrant-cachier')
+      config.cache.scope = :box
+      config.cache.synced_folder_opts = {
+        type: :nfs,
+        mount_options: ['rw', 'vers=3', 'tcp', 'nolock']
```
ekohl: Could we have this configurable? Last time I tried cacher I used nfs4 since that's what I use.

lib/forklift/box_loader.rb
```diff
@@ -1,96 +1,47 @@
...
+      box_files.each { |tmp_boxes| add_boxes(tmp_boxes) }
+      @boxes
     end
 
-    def find_base_box(name)
-      return false if name.nil?
-      @boxes[name]
+    def plugin_base_boxes
+      base_boxes = Dir.glob "#{@root_dir}/plugins/*/base_boxes.yaml"
+      base_boxes.each { |boxes| add_boxes(boxes) }
```
ekohl: Given this loads the boxes as well, should this be ```add_plugin_base_boxes```?

lib/forklift/box_distributor.rb
```diff
@@ -0,0 +1,163 @@
+module Forklift
+  class BoxDistributor
+
+    VAGRANTFILE_API_VERSION = '2'.freeze
+
+    def initialize(boxes)
+      @ansible_groups = {}
+      @boxes = boxes
+      @boxes = @boxes.keys.sort.each_with_object do |hash, key|
```
ehelms: Yup

lib/forklift/box_distributor.rb
ehelms: Per box or per checkout of Forklift?

ekohl: I'd be fine with the latter.

ehelms:
Updated. @ekohl I addressed your requests through the introduction of a `settings.yaml` file. The README tries to lay out how to use it.

ehelms:
@ekohl would you like to take another look at this?

lib/forklift/box_distributor.rb
```diff
@@ -0,0 +1,182 @@
...
+        hash
+      end
+      @settings = settings
+    end
+
+    def settings
+      overrides = {}
+      settings_file = "#{File.dirname(__FILE__)}/../../settings.yaml"
+      default_settings = {
+        'memory' => 4096,
```
ekohl: For the old libvirt and virtualbox images it was 4608. Is this reduction intentional?

lib/forklift/box_distributor.rb
```diff
@@ -0,0 +1,182 @@
...
+        override.vm.box_url = box.fetch('libvirt') if box.fetch('libvirt', false)
+        override.vm.synced_folder '.', '/vagrant', type: @settings['sync_type'], rsync__args: ['--max-size=100m']
+
+        if box.fetch('bridged', false)
+          override.vm.network :public_network, :dev => box.fetch('bridged'), :mode => 'bridge'
+        end
+        networks.each do |network|
+          override.vm.network network['type'], network['options']
+        end
+        p.cpus = box.fetch('cpus') if box.fetch('cpus', false)
```
ekohl: There are defaults for this. Is there a case where this isn't set (and do we care about that case)?

lib/forklift/box_distributor.rb
ehelms: Not entirely, I think my brain thought it was a strange number and rounded down to 4GB instead of 4.5GB. I can update.

ehelms:
@ekohl updated with your comments addressed

lib/forklift/box_distributor.rb
ekohl: Good point

docs/library.md
```diff
@@ -0,0 +1,50 @@
...
+
+    mkdir playbooks
+    mkdir roles
+    mkdir pipelines
+
+Deploy a file with the base set of boxes the plugin requires:
+
+    vim base_boxes.yaml
+
+```
```
ekohl: You could add yaml here for syntax highlighting.

docs/library.md
```diff
@@ -0,0 +1,50 @@
...
+centos7:
+  box_name:   'centos/7'
+  image_name: !ruby/regexp '/CentOS 7.*PV/'
+  default:    true
+  pty:        true
+```
+
+Next, deploy a Vagrantfile that creates a box loader, adds the base boxes defined in the plugin and then call the distributor so Vagrant knows about the boxes:
+
+```
```
ekohl: And ruby

ehelms:
Docs updated for syntax highlighting

PR #429, opened 2017-04-05
fixes #428 - Require ostree for devel install

### LINKED ISSUE #428: Require ostree for a "devel" install

parthaa:


johnpmitsch:
@parthaa should you remove the existing ostree playbooks?

parthaa:
@johnpmitsch in my opinion no because the katello-ostree-devel playbook will help people who havent installed ostree as yet. Its easier for me to say run ```ansible-plabook -l <host> playbooks/katello_ostree_devel.yml``` if you dont have ostree already.  May be over time we can deem this not required as people update their devel envs.

johnpmitsch:
@parthaa I'm ok with this change, others should weigh in though: @ehelms @stbenjam @jlsherrill any objections to adding ostree to a dev install?

jlsherrill:
I think we could go ahead and get rid of the other ostree playbooks, people can just spin up a new dev env.  That said i'm not opposed to leaving it for a bit longer

ekohl:
I'm not sure about this repo, but in theforeman puppet modules we use GH-xxx for GH issue numbers and #xxx for Redmine issues.

johnpmitsch:
@parthaa on second thought I agree with @jlsherrill, I feel like they aren't really used enough to warrant keeping them around when someone can spin up a new env and get ostree. My gut feeling is the "follow up and remove them" task will keep getting pushed back and won't happen, so we should just do it now

johnpmitsch:
@ekohl why not `closes #xxx` in the commit message for GH issues so it [automatically closes the issue](https://help.github.com/articles/closing-issues-via-commit-messages/)?

ekohl:
@johnpmitsch closes GH-xxx also works for that and there's no confusion which issue tracker is used.

johnpmitsch:
@ekohl I didn't realize that works as well, good to know! :smile:

ehelms:
I agree, at least getting rid of https://github.com/theforeman/forklift/blob/master/playbooks/katello_ostree_devel.yml with this PR and then I'll merge it @parthaa

parthaa:
@ehelms updated

johnpmitsch:
@parthaa do we need `playbooks/katello_ostree.yml` and `playbooks/roles/ostree_repositories/tasks/main.yml`?

parthaa:
@johnpmitsch deleted playbooks/katello_ostree.yml since ostree will get installed by default with the prod env also not just dev. 

We need ```playbooks/roles/ostree_repositories/tasks/main.yml``` because that has the upstream repo info for ostree.

PR #430, opened 2017-04-06
Fixes #421 - Add option to set folders to sync into the box

### LINKED ISSUE #421: Add option for rsync-ing local git repos for development boxes
I would like to make some local changes (on hypervisor's filesystem) and test them from centos7-devel box. It would be nice, if foreman and katello folders were synced between the hypervisor and the box. Of course the paths should be configurable, just as the username is right now.

bastilian:
This adds the necessary configuration for vagrant to enable
defining synced folders in `boxes.yaml` in the following form:

```
centos7-devel:
  box: centos7
  private_ip: "172.28.128.15"
  synced_folders:
    - mount_point: '/syncs/foreman'
      path: <%= "#{ENV['HOME']}/RedHat/Foreman/Repos/foreman" %>
      options:
        type: 'nfs'
        nfs_udp: false
```

Since `nfs` requires a private network, this can now be also configured easily via `private_ip`.
`options` are simply passed through to allow using any available adapter. I've only tested `nfs` so far, but `rsync` and `sshfs` should work as well.

### DIFF
--- a/lib/forklift/box_distributor.rb
+++ b/lib/forklift/box_distributor.rb
@@ -72,6 +72,7 @@ module Forklift
         configure_libvirt(machine, box, networks)
         configure_virtualbox(machine, box)
         configure_rackspace(machine, box)
+        configure_synced_folders(machine, box)
 
         yield machine if block_given?
       end
@@ -97,8 +98,7 @@ module Forklift
     def configure_networks(networks)
       return [] if networks.empty?
       networks.map do |network|
-        symbolized_options = network['options'].inject({}) { |memo, (k, v)| memo.update(k.to_sym => v) }
-        network.update('options' => symbolized_options)
+        network.update('options' => symbolized_options(network['options']))
       end
     end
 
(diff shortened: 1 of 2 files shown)


ehelms:
@bastilian will need a rebase

lib/forklift/box_distributor.rb
```diff
@@ -134,6 +135,28 @@ def configure_shell(machine, box)
       end
     end
 
+    # Configures synced folders defined for the box
+    # and the private network required for them
+    def configure_synced_folders(machine, box)
+      configure_private_network(machine, box)
+
+      box.fetch('synced_folders', []).each do |folder|
+        machine.vm.synced_folder folder['path'], folder['mount_point'], folder['options'].symbolize_keys
```
ehelms: /Could you use `with_indifferent_access` here instead of monkey patching Hash?

lib/forklift/box_factory.rb
```diff
@@ -30,6 +32,11 @@ def add_boxes(box_file, version_file)
 
     private
 
+    def load_box_file(file)
+      file = File.read(file)
+      YAML.safe_load(ERB.new(file).result, [Regexp], [], true)
```
ehelms: Why does this need to pass through ERB?

lib/forklift/box_distributor.rb
bastilian: `with_indifferent_access` is a Rails method, like `symbolize_keys`.

The keys only need to be symbols, since Vagrant expects them for options.
I will move the method into `BoxDistributor`, since that is the only place for now where it is needed.

I just saw that `configure_networks` too needs that, I'll use it there as well then.

lib/forklift/box_factory.rb
bastilian: I wanted to be able to use environment variables for defining paths for mounts, as the example shows, passing it through ERB is the simples way to get that.

bastilian:
@ehelms It's rebased and `symbolized_keys` is now `BoxDistributor#symolized_options`. Let me know what you think about the introduction of `ERB`.

lib/forklift/box_factory.rb
ehelms: Ahh, nice!

ehelms:
Looks great, thanks @bastilian !

lib/forklift/box_factory.rb
ekohl: This is something I've wanted but wasn't sure how to add. Cool!

bastilian: Use it wisely though. :) It can be hard, if overused, to debug, when the ERB output disrupts the YAML syntax.

PR #431, opened 2017-04-06
Enable bridging when using VirtualBox as provider

bastilian:
When a bridge device was set, it was not configured when VirtualBox is used as a provider.
This adds the configuration to enable it there as well.

### DIFF
--- a/lib/forklift/box_distributor.rb
+++ b/lib/forklift/box_distributor.rb
@@ -157,9 +157,13 @@ module Forklift
         p.cpus = box.fetch('cpus') if box.fetch('cpus', false)
         p.memory = box.fetch('memory') if box.fetch('memory', false)
 
-        if box.fetch('name').to_s.include?('devel')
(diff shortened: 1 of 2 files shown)


lib/forklift/box_distributor.rb
```diff
@@ -157,9 +157,13 @@ def configure_virtualbox(machine, box)
...
         p.memory = box.fetch('memory') if box.fetch('memory', false)
 
-        if box.fetch('name').to_s.include?('devel')
-          override.vm.network :forwarded_port, guest: 3000, host: 3330
-          override.vm.network :forwarded_port, guest: 443, host: 4430
+        bridged = box.fetch('bridged', false)
+
+        if bridged
+          override.vm.network :public_network, bridge: bridged
+        elsif box.fetch('name').to_s.include?('devel') && !bridged
```
ekohl: The ```!bridged``` is redundant since it's in the elsif branch.

PR #435, opened 2017-04-07
Use override for devel bridge port forwarding

ehelms:


bastilian:
Ah! good catch! Though i’m not sure if this is still needed with bridging available and being able to define networks, with which you can also configure port forwarding. I'd actually remove the port forwarding parts here, and move the networking definition from `configure_libvirt` to it's own method, that way the configuration for libvirt and virtualbox is consistent and could actually be refactored into one.

Another thing is that using the boxes in development also requires webpack to be available, for some, which would need to be configured to be forwarded here as well, when keeping the forwarding here. The changing of the port for webpack also brings a couple of other issues. That was the reason why I added bridging. :D

cfouant:
works for me, thanks @ehelms ACK

PR #437, opened 2017-04-07
Configure private network only if synced folders are set

bastilian:


ehelms:
I was going to try to test this against CI directly -- do you mind rebasing the branch the PR is coming from with master?

bastilian:
@ehelms like so: https://github.com/theforeman/forklift/pull/438

PR #442, opened 2017-04-20
allow setting provider specific options in the boxes yaml

evgeni:
this allows to pass arbitrary options to the providers, which can be used to set options that are not exposed via a specific named parameter.

examples are cpu-mode, management_network_* etc.


dLobatog:
I was doing this manually but never committing the file 😄 
Thanks @evgeni !

PR #446, opened 2017-05-01
Add a dynflow devel box

akofink:


playbooks/roles/dynflow_devel/tasks/git.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: 'Install git'
+  become: true
+  yum: name=git-core state=present
```
akofink: How do I reuse tasks? i.e. this is directly copied from [hammer_devel](https://github.com/theforeman/forklift/blob/master/playbooks/roles/hammer_devel/tasks/git.yml)

ekohl: I believe by extracting them to separate roles or includes

playbooks/roles/dynflow_devel/tasks/dynflow_install.yml
```diff
@@ -0,0 +1,36 @@
...
+  yum:
+    name:
+      - gcc-c++
+      - mysql-devel
+      - postgresql-devel
+      - sqlite-devel
+    state: present
+
+- name: 'Install bundler'
+  gem: name=bundler state=present
```
ekohl: I tend to use bundler from the system ruby. Any reason to prefer it as a gem?

akofink: There is a pretty major version difference, it seems. From rubygems I see bundler 1.14.6. The rubygem-bundler package is 1.7.8. It seems like a newer version would be desirable in development. Is there a reason you use the packaged version?

ekohl: I generally prefer system packages, maybe because I trust them more to not do stupid things. It mostly stems from my time as a sysadmin where you'd never go around the package manager. This may not apply to dev boxes though :).

akofink: I'm willing to change it if you'd like, @ekohl. I mainly just copied [hammer_install.yml](https://github.com/theforeman/forklift/blob/master/playbooks/roles/hammer_devel/tasks/hammer_install.yml#L28)

ekohl: @akofink it's fine, but a ruby_devel role could be nice as a base rol for both.

akofink:
I moved git and ruby-2-3 out into separate roles. Because hammer-devel used the same tasks, I modified it to use the external roles.

PR #452, opened 2017-05-17
Various bats improvements

ekohl:


bats/fb-content-katello.bats
```diff
@@ -109,7 +109,8 @@ setup() {
 }
 
 @test "install subscription manager" {
-  cat > /etc/yum.repos.d/candlepin.repo << EOF
+  if tIsRHEL 6; then
+    cat > /etc/yum.repos.d/subscription-manager.repo << EOF
```
jlsherrill: we can just drop the rhel6 support now, katello and foreman no longer support it.

ekohl: We still have pipeline tests for older foreman/katello versions so I figured I'd leave it in for now, mostly because I don't know how much that is still used.

jlsherrill: ahhh, k
works for me!

jlsherrill:
ACK thanks @ekohl !

PR #453, opened 2017-05-17
bats: Configure hammer and use that configuration

ekohl:


ekohl:
Still missing is the pipeline changes to set the proper host as environment variable, but that's already broken.

ehelms:
I thought the installer configured all of this for us as of [1]?

[1] https://github.com/Katello/katello-installer/commit/730de817f3681978ba5efc33b7394337719aa577

ekohl:
That's the katello scenario, not the proxy.

ehelms:
Ahhh, but the hammer setup is being for all tests. Maybe the proxy tests should be just run on the server? They don't appear to be doing anything specific that has to be done on the proxy.

ekohl:
That's possible. That still means we can remove the username/password things and clean up a bit.

ehelms:
Agreed. I think it would simplify things a bit to just run hammer on the server itself since I also doubt its common to install and run hammer on a proxy.

ekohl:
Part of https://github.com/theforeman/forklift/pull/451

PR #456, opened 2017-05-30
Fixes #19663 - format foreman installer options as a list in the docs

evgeni:


docs/development.md
```diff
@@ -34,7 +34,11 @@ centos7-devel:
     group: 'devel'
     variables:
       katello_devel_github_username: <REPLACE ME>
-      foreman_installer_options_internal_use_only: "--katello-devel-github-username myname --katello-devel-upstream-remote-name origin --katello-devel-extra-plugins theforeman/foreman_remote_execution --katello-devel-extra-plugins theforeman/foreman_discovery"
+      foreman_installer_options_internal_use_only:
+        - "--katello-devel-github-username <REPLACE ME>"
```
ekohl: I wonder if you can use ```{{ katello_devel_github_username }}``` at this point.

evgeni: Probably we can, will test in a second. Thanks for the suggestion!

ekohl: Given we override the ```foreman_installer_options_internal_use_only``` variable it looks like ```katello_devel_github_username``` isn't used at all actually. This example can be simplified to get rid of it.

evgeni: I thought about that too. I prefer to use `{{ katello_devel_github_username }}` as that allows us to yank the `foreman_installer_options_internal_use_only` at a later point and still have the username set properly.

ekohl:
Thanks!

PR #461, opened 2017-06-02
Dont break ansible provisions if a box doesnt have group defined

sean797:
to test, add this into boxes.yaml and it will break all ansible provisioning by creating an unnamed group in `.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory`

```yaml
test:
  box: centos7
  ansible:
    server: 'centos7-katello-3-2'
```

### DIFF
--- a/lib/forklift/box_distributor.rb
+++ b/lib/forklift/box_distributor.rb
@@ -111,13 +111,16 @@ module Forklift
 
     def configure_ansible(machine, ansible, box_name)
       return unless ansible
-      unless @ansible_groups[ansible['group'].to_s]
-        @ansible_groups[ansible['group'].to_s] = []
+
+      if ansible.key?('group') || !ansible['group'].nil?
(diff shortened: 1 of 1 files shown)


lib/forklift/box_distributor.rb
```diff
@@ -111,6 +111,8 @@ def configure_networks(networks)
 
     def configure_ansible(machine, ansible, box_name)
       return unless ansible
+      return if !ansible.key?('group') || ansible['group'].nil?
```
ehelms: Could we just skip the group setup portion down below so the box is still setup and configured? Basically, just not require a group to be specified in all cases as I think thats a valid use case.

sean797:
Thanks for reviewing @ehelms 👍 Nice idea.. Updated.

ehelms:
Looks good, thanks @sean797 !

PR #465, opened 2017-06-02
split bats into the real tests and roles to set stuff up

evgeni:
before we had bats doing a lot of setup and then running tests.
this moves the setup tasks into a series of own roles (haveged, disable_firewall, clean_puppet) and lets bats do what it is supposed to do: run tests.


boxes.yaml.example
```diff
@@ -44,6 +44,7 @@ centos7-bats-ci:
       bats_tests:
         - "swapfile.bats"
         - "fb-install-katello.bats"
```
ekohl: IMHO this can be left out now. We should rely on ansible for installation.

playbooks/roles/bats/defaults/main.yaml
```diff
@@ -5,5 +5,6 @@ bats_output_dir: "/root/bats_results"
 bats_update_forklift: "yes"
 bats_tests:
   - "fb-install-katello.bats"
```
ekohl: Should also be left out IMHO.

ekohl:
Thanks you for this! I wanted to do this but hadn't gotten around to it.

boxes.yaml.example
evgeni: You mean basically completely dropping `fb-install-katello.bats`? Yeah we can do that.

ekohl: The file can stay, but we shouldn't recommend using it I think.

evgeni: If the file is not used and not recommended, why should it stay? Archaeology lessons can still be done using the Git history, an unused file is just confusing. Same applies for `katello-bats` btw, I think it's unused by now?

ekohl: That's the problem I have with this repo: I don't know how people use it. Maybe @ehelms or @stbenjam know it better.

ehelms: One of the biggest things we need is https://github.com/evgeni/forklift/blob/99fdda8d44631373e75ece1bb3aefdc2a970f52a/bats/fb-install-katello.bats#L47-L52 which was put in to help out on the CI system as far as I remember.

So maybe we need like a setup role to do a few of those items (instead of "testing" them). Otherwise I think it's safe to remove the install test file.

ekohl: For now I'd suggest a haveged role rather than a more global setup role which may get parameters. Start simple.

bats/fb-test-katello.bats
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bats
```
ehelms: @evgeni I think one additional check could be valuable here -- doing a curl of `/katello/api/ping` since this provides information of about services that the status check does not, e.g. whether or not Katello can talk to Pulp or Candlepin and whether or not Pulp is having any connectivity issues with the workers. These are things that can produce big smells for installation health.

evgeni: That's what `hammer ping` does. No need to `curl` yourself ;)

bats/fb-test-katello.bats
```diff
@@ -0,0 +1,42 @@
...
+  curl -sk "https://localhost$URL_PREFIX/users/login" | grep -q login-form
+}
+
+@test "check smart proxy is registered" {
+  count=$(hammer --csv proxy list | wc -l)
+  [ $count -gt 1 ]
+}
+
+@test "check host is registered" {
+  [ x$FOREMAN_VERSION = "x1.3" ] && skip "Only supported on 1.4+"
```
ehelms: Ha - I think we could drop this line

evgeni: So I did in my local branch, let me polish and push that before you continue review.

evgeni:
@ekohl @ehelms @stbenjam please re-review, I made all the setup steps into own roles, and killed `fb-install-katello.bats` :)

ehelms:
Looks good to me, will let @ekohl have a final pass at the updated version

bats/fb-test-katello.bats
```diff
@@ -0,0 +1,41 @@
...
+  puppet agent -t -v
+}
+
+@test "check web app is still up" {
+  curl -sk "https://localhost$URL_PREFIX/users/login" | grep -q login-form
+}
+
+@test "check smart proxy is registered" {
+  count=$(hammer --csv proxy list | wc -l)
+  [ $count -gt 1 ]
```
ekohl: I wonder if grepping for the hostname would be a better check, but I think this is directly ported so it's fine to leave this in.

playbooks/roles/clean_puppet/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
```
ekohl: I'd drop this role. It was originally added in the foreman-bats because you want to start with a clean install, but in the ansible context it doesn't really make sense (unless you do it before the foreman-installer run).

playbooks/roles/haveged/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+- name: 'Install Haveged (faster population of PRNG)'
+  yum:
+    name: haveged
+    state: latest
```
ekohl: I think just ensuring it's installed is good enough.

bats/fb-test-katello.bats
evgeni: yeah. `grep -q $hostname` sounds good to me.

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

playbooks/roles/haveged/tasks/main.yml
evgeni: Thats how it was in the freeipa role.

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

bats/fb-test-katello.bats
evgeni: ack. Was not sure about the history and thought porting was cheaper then thinking.

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

playbooks/roles/clean_puppet/tasks/main.yml
evgeni: ack. Was not sure about the history and thought porting was cheaper then thinking.

evgeni:
@ekohl there we go, all 3 comments done

ekohl:
Thanks!

PR #466, opened 2017-06-05
Increase default RAM to 6gbs

johnpmitsch:
With puppet 4, katello is more resource intensive and we should
default to a higher amount of RAM to avoid OOM errors. This can
always be overridden with `memory:` option in boxes.yaml


ehelms:
@johnpmitsch Have you been hitting OOM errors? In production testing? devel?

johnpmitsch:
@ehelms yes, consistently on the katello 3.4 production box.

jlsherrill:
ACK from me, i feel like if you just leave a box running you'll start seeing OOMs

PR #467, opened 2017-06-12
bats cleanup

evgeni:
just a bit more cleanup for bats. now we can run bats also on a RHEL box that was previously subscribed to $something ;)

### DIFF
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -124,7 +124,7 @@ EOF
   run yum erase -y 'katello-ca-consumer-*'
   rpm -Uvh http://localhost/pub/katello-ca-consumer-latest.noarch.rpm || true
   subscription-manager register --force --org="Test_Organization" --activationkey="Test AK" || true
-  subscription-manager status | grep -q "Current"
+  subscription-manager list --consumed | grep "Test Product"
 }
 
 @test "check content host is registered" {
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -121,6 +121,7 @@ EOF
     rpm -e `rpm -qf /etc/rhsm/ca/candlepin-local.pem`
   fi
 
+  run yum erase -y 'katello-ca-consumer-*'
   rpm -Uvh http://localhost/pub/katello-ca-consumer-latest.noarch.rpm || true
   subscription-manager register --force --org="Test_Organization" --activationkey="Test AK" || true
   subscription-manager status | grep -q "Current"
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -122,8 +122,8 @@ EOF
   fi
 
   run yum erase -y 'katello-ca-consumer-*'
-  rpm -Uvh http://localhost/pub/katello-ca-consumer-latest.noarch.rpm || true
-  subscription-manager register --force --org="Test_Organization" --activationkey="Test AK" || true
+  run rpm -Uvh http://localhost/pub/katello-ca-consumer-latest.noarch.rpm
+  run subscription-manager register --force --org="Test_Organization" --activationkey="Test AK"
   subscription-manager list --consumed | grep "Test Product"
 }
 
--- a/bats/fb-finish-katello.bats
+++ b/bats/fb-finish-katello.bats
@@ -10,7 +10,6 @@ load foreman_helper
   tail -n100 /var/log/{apache2,httpd}/*_log /var/log/foreman{-proxy,}/*log /var/log/messages > /root/last_logs || true
   foreman-debug -q -d /root/foreman-debug || true
   if tIsRedHatCompatible; then
-    tPackageExists sos || tPackageInstall sos
     sosreport --batch --tmp-dir=/root || true
   fi
 }
--- a/playbooks/roles/bats/tasks/main.yml
+++ b/playbooks/roles/bats/tasks/main.yml
@@ -9,6 +9,7 @@
     - ruby
     - yum-utils
     - git
+    - sos
 
 - name: "Get /usr/bin/bats stat"
   stat:


evgeni:
@ekohl @ehelms would love some input from you :)

bats/fb-content-katello.bats
```diff
@@ -121,8 +121,13 @@ EOF
     rpm -e `rpm -qf /etc/rhsm/ca/candlepin-local.pem`
   fi
 
+  yum erase -y 'katello-ca-consumer-*' || true
   rpm -Uvh http://localhost/pub/katello-ca-consumer-latest.noarch.rpm || true
   subscription-manager register --force --org="Test_Organization" --activationkey="Test AK" || true
+  if tIsRHEL && ! tIsCentOSCompatible; then
+    # skipping the rest of the test as we don't provide RHEL subs
+    skip
+  fi
```
ehelms: Will the line below this fail on RHEL? Is that cause Current won't be present since there will be 0 subs? Even custom product subs?

bats/fb-finish-katello.bats
```diff
@@ -10,7 +10,6 @@ load foreman_helper
   tail -n100 /var/log/{apache2,httpd}/*_log /var/log/foreman{-proxy,}/*log /var/log/messages > /root/last_logs || true
   foreman-debug -q -d /root/foreman-debug || true
   if tIsRedHatCompatible; then
-    tPackageExists sos || tPackageInstall sos
```
ehelms: Why does this need to be removed?

bats/fb-content-katello.bats
evgeni: It fails on RHEL as we don't provide a subscription for RHEL itself, so the output will be "Unknown".
As an alternative we could look at `sub-man list --consumed` and verify that lists the subs we actually added to the AK earlier.

bats/fb-finish-katello.bats
evgeni: I moved this to the general package installation in `playbooks/roles/bats/tasks/main.yml`, not removed it.
The background is: we're installing it *after* resubscribing the box, but nothing provides `sos` in any of our content views. And in the case of RHEL, we don't have any regular online repositories anymore.

bats/fb-content-katello.bats
evgeni: Yupp, just checked, we can do `subscription-manager list --consumed | grep "Test Product"`, what do you think?

bats/fb-content-katello.bats
```diff
@@ -121,8 +121,13 @@ EOF
     rpm -e `rpm -qf /etc/rhsm/ca/candlepin-local.pem`
   fi
 
+  yum erase -y 'katello-ca-consumer-*' || true
```
ekohl: There is a bats way to run a command and ignore its status code by using ```run```. I think that's cleaner than ```|| true```. Maybe the lines below should also be changed. It also makes me wonder when ignoring those status codes makes sense.

evgeni: These `|| true` have been there since @stbenjam added the code in 0ce8c83a, no idea why :(

bats/fb-content-katello.bats
evgeni: done

bats/fb-content-katello.bats
evgeni: moved all these `|| true` to `run`, but did not dare to remove it completelly

bats/fb-content-katello.bats
```diff
@@ -121,9 +121,10 @@ EOF
     rpm -e `rpm -qf /etc/rhsm/ca/candlepin-local.pem`
   fi
 
-  rpm -Uvh http://localhost/pub/katello-ca-consumer-latest.noarch.rpm || true
-  subscription-manager register --force --org="Test_Organization" --activationkey="Test AK" || true
-  subscription-manager status | grep -q "Current"
+  run yum erase -y 'katello-ca-consumer-*'
+  run rpm -Uvh http://localhost/pub/katello-ca-consumer-latest.noarch.rpm
+  run subscription-manager register --force --org="Test_Organization" --activationkey="Test AK"
```
ehelms: What happens with bats if this command or the one above fail? Will it fail the test? Cause I think these are two good assertions to ensure work.

evgeni: `run` ignores the exit code (basically it's the same as `|| true` here, let's ignore the fact that it sets vars for further analyzing the exit code and output) so it won't fail = no change to old behavior.

I agree that adding those as further tests is a good idea, but
* they were not used as such before this PR
* I am not sure what `sub-man` exits if it cannot add *all* installed product subs

Given that we are checking that we end up with a sub for "Test Product", I think it is safe to assume that the steps before were OK and not try to check their exit codes (at least at the moment).

ehelms: Will this mask the error and hide it? I'd like to know that both installation of the bootstrap RPM succeeded as well as that registration did not have an error associated with it. While I agree the final condition should be enough to verify the workflow, register is such an important operation that it, in my opinion, should ensure if it errors we see that and it fails the test.

evgeni: Yes, this hides the output (but so did `|| true`).
Let me see how `subcription-manager` behaves when it registers fine, but does not get all the subs.

evgeni:
``` ✗ register subscription manager
   (in test file bats/fb-content-katello.bats, line 126)
     `subscription-manager register --force --org="Test_Organization" --activationkey="Test AK"' failed
   Retrieving http://localhost/pub/katello-ca-consumer-latest.noarch.rpm
   Preparing...                          ########################################
   Updating / installing...
   katello-ca-consumer-pipeline-satellite########################################
   The system has been registered with ID: 67f8e2fb-c507-4473-a3c9-31a1b25b8f82 
   
   Installed Product Current Status:
   Product Name: Red Hat Software Collections (for RHEL Server)
   Status:       Not Subscribed
   
   Product Name: Red Hat Enterprise Linux Server
   Status:       Not Subscribed
   
   Unable to find available subscriptions for all your installed products.
```

:crying_cat_face:

ehelms: Lame, but thanks for trying it!

PR #468, opened 2017-06-13
Don't skip installer on bats playbook

ehelms:
Due to install test being removed


ehelms:
This will fix current pipeline breakage.

jlsherrill:
ACK

PR #469, opened 2017-06-13
Set katello scenario for bats installation

ehelms:


ehelms:
The little things that go unnoticed and break things in our pipeline.

jlsherrill:
ACK

PR #472, opened 2017-06-15
move swapfile to a role and call it in the pipeline before install

evgeni:


playbooks/roles/swapfile/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    path: "{{ swapfile_path }}"
+  register: swapfile_check
+- name: create swap file {{ swapfile_path }}
+  command: fallocate -l {{ swapfile_size }} {{ swapfile_path }}
+  when: not swapfile_check.stat.exists
+- name: set permissions on swap file
+  file:
+    path: "{{ swapfile_path }}"
+    mode: 0600
+- name: mkswap {{ swapfile_path }}
```
ehelms: Does this work? I've seen errors, no entry or just the literal words in the past when using this in a name.

playbooks/roles/swapfile/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
+---
```
ehelms: Do you mind putting blank lines between each task? I find it helps with readability, and yes its fair to ask for a lint rule around this.

playbooks/roles/swapfile/tasks/main.yml
evgeni: Sure. It just does not work with `with_*` loops.

proof:
```
TASK [swapfile : create swap file /swapfile] ***********************************
Thursday 15 June 2017  18:31:12 +0200 (0:00:00.446)       0:00:07.499 ********* 
changed: [localhost]
```

playbooks/roles/swapfile/tasks/main.yml
evgeni: done (yes, a lint rule would be good)

ehelms:
Thanks @evgeni !

PR #474, opened 2017-06-16
swapfile: don't calculate checksum or md5sum of the file

evgeni:
this gives quite an performance boost on bigger files like our swapfile

poor mans benchmark on my laptop:

    stat -------------------------------------------- 6.71s
    get_md5=False ----------------------------------- 3.49s
    get_checksum=False ------------------------------ 3.25s
    get_checksum=False get_md5=False ---------------- 0.22s


ekohl:
Thanks!

PR #478, opened 2017-06-28
Add katello-host-tools and errata testing

ehelms:
This adds two things:

 1) Testing that errata applicability is working and installation of errata works
 2) Support for new katello-host-tools packaging [1]

Once [1] is merged, I'd like to merge this so that nightly will verify that our packaging updates were correct.

@jlsherrill Please take a look

[1] https://github.com/Katello/katello-packaging/pull/476

### DIFF
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -32,7 +32,7 @@ setup() {
 @test "create package repository" {
   hammer repository create --organization="${ORGANIZATION}" \
     --product="${PRODUCT}" --content-type="yum" --name "${YUM_REPOSITORY}" \
-    --url https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/zoo/ | grep -q "Repository created"
+    --url https://jlsherrill.fedorapeople.org/fake-repos/needed-errata/ | grep -q "Repository created"
 }
 
 @test "upload package" {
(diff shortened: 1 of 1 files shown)


jlsherrill:
This seems sane, ACK from me

bats/fb-content-katello.bats
```diff
@@ -140,12 +140,21 @@ EOF
...
 }
 
-@test "install katello-agent" {
-  tPackageInstall katello-agent && tPackageExists katello-agent
+@test "install katello-host-tools" {
+  tPackageInstall katello-host-tools && tPackageExists katello-host-tools
+}
+
+@test "install package locally" {
+  run yum -y remove walrus
```
ekohl: Won't yum exit with 0 if it's already removed? In that case we can remove the ```run```.

ehelms: I don't know actually, but I don't consider this an assertion that we are testing, more of test cleanup so my preference is to use run in those situations. SInce the katello-host-tools has been merged to packaging, I'm going to merge this in order for it to be tested.

ekohl: While it may not be an assertion I do think you want to be sure it's not installed before. Otherwise you're not testing what you think you're testing. I don't mind you merging it but in my experience this is the sort of thing that can mask real problems.

ehelms: Hrmm, another reason I thought to choose run was my assumptions around how the line below this would work:

```
tPackageInstall walrus-0.71 && tPackageExists walrus-0.71
```

In that, this should assert that 0.71 is installed and if not error out. I was not intending to test that the package could be installed but rather that at this point, walrus-0.71 is on the system, to allow for there to be applicable errata. So I think I assumed that `tPackageExists` would error if some other version of walrus was installed.

ekohl: Ah, so in puppet terms:
```puppet
package { 'walrus':
  ensure => '0.71'
}
```

Or ```puppet resource package walrus ensure=0.71```

Issue #439, opened 2017-04-10
forklift wrapper command


bastilian:
The idea is to have a `forklift` command that is a wrapper around vagrant, which can be run from any directory and looks for a `boxes.yml` (or `forklift_boxes.yml`) in that directory. Possibly only limited to these boxes.

This would allow to define boxes required for plugin development in plugin directories and manage them from there. The `boxes.yml` can then also be shared, and plugin authors can include forklift in their repositories (plus the playbooks required to run).

With having now `ERB` support for YAML files, user specific settings can be put into other places.

Setting up a development environment for a specific plugin is then just a matter of running `forklift up` and mounting the local clone in the VM.

PR #460, opened 2017-05-31
Pipelines for 3.4

ehelms:

PR #425, opened 2017-04-01
Add Debian support

ekohl:
This allows you to provision a Foreman 1.14 install on Debian. There are
still many corners that problably don't support it and Katello boxes are
also generated but it's a start.

PR #480, opened 2017-07-02
[WIP] Add a forklift script

ekohl:
This is based on the idea in https://github.com/theforeman/forklift/issues/439 to provide a wrapper. The initial focus is on removing the duplication in pipelines and make them easy to run. Since writing Python is much easier for me I decided to write a simple wrapper using click.

My goal is to move a lot of the hardcoding pipeline options into config/versions.yaml and implement the upgrade command to verify upgrades. By doing this through versions.yml we should automatically get matrix pipelines to upgrade from any base version to any other version. We could even implement a from-to range upgrade (3.0 -> 3.1 -> 3.2 -> 3.3 -> 3.4 -> nightly).

Let me know what you think.

### DIFF
--- a/bin/forklift
+++ b/bin/forklift
@@ -1,0 +1,276 @@
+#!/usr/bin/env python3
+
+# Needs at least python 3.4 or the pathlib library
+# Needs click
+# Needs yaml
+
+import subprocess
+from collections import namedtuple
+from copy import deepcopy
+from distutils.version import StrictVersion
+from pathlib import Path
+
+import click
+import yaml
+
+ROOT = (Path(__file__).parent / '..').absolute()
+VERSIONS = ROOT / 'config' / 'versions.yaml'
+PIPELINES = ROOT / 'pipelines'
+
+
+# From https://stackoverflow.com/a/20666342
+def merge(source, destination):
+    """
+    >>> a = {'first': {'all_rows': {'pass': 'dog', 'number': '1'} } }
+    >>> b = {'first': {'all_rows': {'fail': 'cat', 'number': '5'} } }
+    >>> merge(b, a) == {'first': {'all_rows': {'pass': 'dog', 'fail': 'cat', 'number': '5'}}}
+    True
+    """
+    for key, value in source.items():
+        if isinstance(value, dict):
+            # get node or create one
+            node = destination.setdefault(key, {})
+            merge(value, node)
+        else:
+            destination[key] = value
+
+    return destination
+
+
+# From https://stackoverflow.com/a/1061350
+class ComparableMixin(object):
+    def __eq__(self, other):
+        return not self < other and not other < self
+
+    def __ne__(self, other):
+        return self < other or other < self
+
+    def __gt__(self, other):
+        return other < self
+
+    def __ge__(self, other):
+        return not self < other
+
+    def __le__(self, other):
+        return not other < self
+
+
+class Host(namedtuple('Host', 'name box memory'), ComparableMixin):
+    def __str__(self):
+        return self.name
+
+    def __lt__(self, other):
+        return self.name < other.name
+
+
+class Installer(namedtuple('Installer', 'foreman katello puppet boxes'), ComparableMixin):
+    """
+    The configuration for a specific installer version
+
+    A wrapper around entries in config/versions.yaml
+    """
+    # TODO: cache this?
+    @property
+    def _version(self):
+        version = '999.999.999' if self.foreman == 'nightly' else self.foreman
+        return StrictVersion(version)
+
+    def __str__(self):
+        return 'foreman-installer {}/{}'.format(self.foreman, self.katello)
+
+    def __lt__(self, other):
+        return self._version < other._version
+
+
+class Segment(namedtuple('Segment', 'product version installer common_roles bats_tests hosts'),
+              ComparableMixin):
+    """
+    A segment within a pipeline. This is a specific product version and the configuration.
+    """
+    @property
+    def _version(self):
+        version = '999.999.999' if self.foreman == 'nightly' else self.foreman
+        return StrictVersion(version)
+
+    def __str__(self):
+        return '{} {} segment'.format(self.product, self.version)
+
+    def __lt__(self, other):
+        return self.product < other.product and self._version < other._version
+
+
+class Pipeline(namedtuple('Pipeline', 'name filename product segments hosts')):
+    @property
+    def _plays(self):
+        boxes = {h.name: {'box': h.box, 'memory': h.memory} for h in self.hosts.values()}
+        variables = {
+            'forklift_name': self.name,
+            'forklift_boxes': boxes,
+        }
+        yield play('localhost', variables, ['forklift'], False)
+
+        groups = sorted_groups(self.hosts.keys())
+
+        for segment in self.segments:
+            if segment.common_roles:
+                variables = {
+                    'katello_repositories_version': segment.installer.katello,
+                    'foreman_repositories_version': segment.installer.foreman,
+                    'puppet_repositories_version': segment.installer.puppet,
+                    'katello_repositories_use_koji': True,
+                    # TODO: 3.3 has foreman_repositories_use_koji=true
+                    'foreman_installer_skip_installer': True,
+                }
+                yield play([h.name for h in self.hosts.values()], variables, segment.common_roles)
+
+            for group in groups:
+                variables = segment.hosts[group]['variables']
+                if group == 'proxy':
+                    variables['foreman_proxy_content_server'] = self.hosts['server'].name
+
+                yield play(self.hosts[group].name, variables, segment.hosts[group]['roles'])
+
+        if segment.bats_tests:
+            yield play(self.hosts['server'].name, {'bats_tests': segment.bats_tests}, ['bats'])
+
+    @property
+    def playbook(self):
+        return list(self._plays)
+
+    def write(self, path):
+        with path.open('w') as fp:
+            yaml.safe_dump(self.playbook, fp, allow_unicode=True, default_flow_style=False)
+
+
+def sorted_groups(groups):
+    order = ('server', 'proxy')
+
+    def get_group_index(group):
+        try:
+            return order.index(group)
+        except ValueError:
+            return len(order)
+
+    return sorted(groups, key=get_group_index)
+
+
+def get_segments(config, product, versions):
+    default_pipelines = config['pipelines']
+
+    for version in config['installers']:
+        if version in versions:
+            installer = Installer(version['foreman'], version['katello'], version['puppet'],
+                                  version['boxes'])
+
+            pipelines = deepcopy(default_pipelines)
+            if 'pipelines' in version:
+                merge(version['pipelines'], pipelines)
+            pipeline = pipelines[product]
+
+            yield Segment(
+                product=product,
+                version=getattr(installer, product),
+                installer=installer,
+                common_roles=pipeline.get('common_roles', []),
+                bats_tests=pipeline.get('bats_tests', []),
+                hosts=pipeline.get('hosts', {}),
+            )
+
+
+def get_hosts(product, hosts, versions, box):
+    return {group: Host('pipeline-{}-{}-{}-{}'.format(product, group, versions, box), box,
+                        options['memory']) for group, options in hosts.items()}
+
+
+def get_pipeline(config, product, versions, box):
+    segments = list(get_segments(config, product, versions))
+
+    if len(segments) != len(versions):
+        segment_versions = [segment.version for segment in segments]
+        missing = ', '.join(version for version in versions if version not in segment_versions)
+        raise ValueError('Version {} not found'.format(missing))
+
+    for segment in segments:
+        if box not in segment.installer.boxes:
+            raise ValueError("Box '{}' is not supported in {}".format(box, segment))
+
+    return Pipeline(
+        name='pipeline-{}-{}'.format(product, '-'.join(versions)),
+        filename='pipeline_{}_{}.yml'.format(product, '_'.join(versions).replace('.', '')),
+        product=product,
+        segments=sorted(segments),
+        hosts=get_hosts(product, segments[-1].hosts, '-'.join(versions), box),
+    )
+
+
+def play(hosts, variables, roles, become=True):
+    return {
+        'hosts': hosts,
+        'become': become,
+        'vars': variables,
+        'roles': roles,
+    }
+
+
+def run_playbook(path, *options, **variables):
+    var_options = (('-e', '{}={}'.format(key, value)) for key, value in variables.items())
+    command = ['ansible-playbook', path] + list(options) + list(*var_options)
+    return subprocess.check_call(command)
+
+
+def run_pipeline(path, destroy):
+    try:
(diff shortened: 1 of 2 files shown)


ekohl:
The second commit generalizes the configuration and allows overriding some variables. I think that I'll refactor it to a ```config.yml``` instead of hardcoded values. Maybe ```versions.yml``` will just be a section ```versions``` in ```config.yml```. Hopefully this will make it trivial to use the same pipeline infrastructure on Satellite with just some additional configs.

ehelms:
Would you mind including a few examples of how one would use it achieve the typical use cases? This does add to odd ball mixture of languages being used throughout the repository.

Do you see this wrapper script helping at all with our ideas of re-organizing the structure of the repository? I was starting to swing back to this idea to push what I have for containers into the mainline branch.

ekohl:
While it does add to the variety of languages, we are using python already. I think it could easily be made compatible with python 2 which is already required for Ansible. Then you just need 1 library (click).

The biggest benefit I see is that https://github.com/theforeman/forklift/pull/460 becomes just some additional lines to define a 3.4 version in ```versions.yml```.

Some examples for the installation pipeline that should be 
```bash
# Default to katello nightly
forklift pipeline install

# Explicit versions
forklift pipeline install katello nightly
forklift pipeline install katello 3.4

# Not yet supported, but should be
forklift pipeline install foreman nightly
forklift pipeline install foreman 1.15

# Or other boxes.
# I'm considering explicit parameters so you can test an EL7 server with an EL6 client
forklift pipeline install katello 3.0 --box centos6
```

The not yet implemented upgrade pipeline:
```bash
# Not yet sure if there are sane defaults or if they should become required
forklift pipeline upgrade

# Test an upgrade from 3.3 to 3.4
forklift pipeline upgrade katello 3.3 3.4

# Test all the way from 3.0 to nightly
forklift pipeline upgrade katello 3.0 3.1 3.2 3.3 3.4 nightly

# And we should support Foreman too
forklift pipeline upgrade foreman 1.14 .1.15

# Also different boxes
forklift pipeline upgrade foreman 1.14 .1.15 --box debian8
```

ekohl:
I should note that it may be possible to do this in pure ansible and if it is, then I'd be happy to use that but for me this was the easiest way to build a pipeline generator.

Another benefit is that it's self-documenting: you can use --help and get a listing of options. It'll use docstrings for groups and commands. The options can have a help parameter as well. Click even has out of the box [bash completion support](http://click.pocoo.org/5/bashcomplete/). That could give us a basic supported feature set.

ekohl:
@ehelms I've updated the PR. It includes some commits that are not related and I can submit them in a separate PR. Other than that I'm happy with the installation pipeline. I still have to look at the generated playbooks for the upgrade part but there's an example in there to add a Foreman pipeline as well. There's still some more work to be done to easily add more pipelines. For example a separate DB server. At this point I'm looking for feedback.

bats/fb-test-foreman.bats
```diff
@@ -0,0 +1,29 @@
...
+  hammer --csv proxy list | grep -q "$(hostname -f)"
+}
+
+@test "check host is registered" {
+  hammer host info --name "$(hostname -f)"
+}
+
+@test "check hammer ping" {
+  hammer ping
+}
```
ehelms: This file does not appear to differ that much from `fb-test-katello.bats`. Could we centralize those into a `fb-health-check.bats` or something to that effect?

ehelms: Oh, I see you paired down the `fb-test-katello`. Still, I'd consider renaming this to be more descriptive of its intent. Also, thoughts on dropping the leading `fb`. I am not even sure I know what that stands for?

roles/puppet_repositories/tasks/3-redhat.yml
```diff
@@ -0,0 +1,5 @@
+--
```
ehelms: Why the `-redhat` file names?

ehelms:
Our `forklift` role is starting to remind me of Linchpin -- https://opensource.com/article/17/6/linchpin

ehelms:
Oops, wrong click

roles/puppet_repositories/tasks/3-redhat.yml
ekohl: That's a preparation for #425.

ekohl:
> Our forklift role is starting to remind me of Linchpin -- https://opensource.com/article/17/6/linchpin

I hadn't looked into it, but we could have a look and see if we can converge on that.

bats/fb-test-foreman.bats
ekohl: I'm **guessing** Foreman Bats, but no real preference from my side. I'm not even sure if we should continue with writing tests in bats. Personally I'm still interested in getting a Robotello test set that is light enough to quickly test, but I haven't used it yet so I don't know if that's feasible.

ehelms:
Couple of questions:

 1) Do you see this forklift script as only a way to control the vagrant + ansible pieces?
 2) Do you envision using this as the sole script for all deployment types in the repository?
 3) Do you think this script could help decouple the "provisioning" from the "install" ? e.g. the vagrant up/destroy from the playbooks run for example
 4) Would stand-alone use of the playbooks be done through ansible or the script?
 5) I am trying to think about whether this could be used to facilitate container based deployments.

ekohl:
> Do you see this forklift script as only a way to control the vagrant + ansible pieces?

It should be a wrapper. That means you can still dive under the hood and call it manually but for those who are new it should provide easier onboarding.

> Do you envision using this as the sole script for all deployment types in the repository?

I'd prefer if common tasks can be done through it and you don't need to know what happens under the hood but it should be preferred because it's easier and better, not because we say it should be.

> Do you think this script could help decouple the "provisioning" from the "install" ? e.g. the vagrant up/destroy from the playbooks run for example

Not sure what you mean here. Do you mean moving all the vagrant controls into the wrapper so the flow is purely forklift -> vagrant -> ansible? I hadn't thought about that but it could be a nice flow. Maybe define more boxes only as needed to keep vagrant listings short and fast.

> Would stand-alone use of the playbooks be done through ansible or the script?

At some point we could provide a wrapper but it should add value. I'd expect direct use of playbooks with ansible to remain usable.

> I am trying to think about whether this could be used to facilitate container based deployments.

In my experience it's easy to add additional commands using [click](http://click.pocoo.org/5/). There's a [guide on buidling git style applications](http://click.pocoo.org/5/complex/#building-a-git-clone). I could see a flow:

```bash
# Set up a local openshift or point it to a remote openshift
forklift openshift init
# Build a deployment on the openshift
forklift openshift up
# Get a short overview (URL, health etc)
forklift openshift status
# Stop the deployment
forklift openshift halt
```

Under water it would read and write config files or call ansible.

In the same way we could create a vagrant wrapper where the create function that defines the box.

ehelms:
To clarify my question around decoupling provision from install, currently our pipeline playbooks are responsible for creating the environment (vagrant up) and install and testing and also bringing the environment down. This limits the re-usability of the pipeline playbooks.

ekohl:
You're right and I mostly tried to work on that in this script. Another thing I considered is using a separate hosts file per pipeline run. That would allow using group instead of hosts in playbooks and make them more generic.

bin/forklift
```diff
@@ -0,0 +1,273 @@
+#!/usr/bin/env python3
+
+# Needs at least python 3.4 or the pathlib library
```
ehelms: Does EL7 have python 3.4?

ehelms:
@ekohl this may take me a bit to review as you are using some Python beyond my knowledge :)

ehelms:
I've always thought the hosts file was an interesting idea given that's how you declare a set of hosts in Ansible.

bin/forklift
```diff
@@ -0,0 +1,273 @@
...
+        return other < self
+
+    def __ge__(self, other):
+        return not self < other
+
+    def __le__(self, other):
+        return not other < self
+
+
+class Host(namedtuple('Host', 'name box memory'), ComparableMixin):
```
ehelms: What's a `namedtuple` and how does it benefit here first more traditional argument passing?

bin/forklift
```diff
@@ -0,0 +1,273 @@
...
+
+
+class Installer(namedtuple('Installer', 'foreman katello puppet boxes'), ComparableMixin):
+    """
+    The configuration for a specific installer version
+
+    A wrapper around entries in config/versions.yaml
+    """
+    # TODO: cache this?
+    @property
```
ehelms: What does this decorator do?

bin/forklift
```diff
@@ -0,0 +1,273 @@
...
+                installer=installer,
+                common_roles=pipeline.get('common_roles', []),
+                bats_tests=pipeline.get('bats_tests', []),
+                hosts=pipeline.get('hosts', {}),
+            )
+
+
+def get_hosts(product, hosts, versions, box):
+    return {group: Host('pipeline-{}-{}-{}-{}'.format(product, group, versions, box), box, options['memory'])
+            for group, options in hosts.items()}
```
ehelms: This a hard line to parse, I'd consider splitting it up for readability.

bin/forklift
```diff
@@ -0,0 +1,273 @@
...
+
+
+def get_pipeline(config, product, versions, box):
+    segments = [s for s in get_segments(config) if s.product == product and s.version in versions]
```
ehelms: Never been a fan of single character variables. Is this an instance where pep8 is OK with them?

bin/forklift
```diff
@@ -0,0 +1,273 @@
...
+    for segment in segments:
+        if box not in segment.installer.boxes:
+            raise ValueError("Box '{}' is not supported in {}".format(box, segment))
+
+    return Pipeline(
+            name='pipeline-{}-{}'.format(product, '-'.join(versions)),
+            filename='pipeline_{}_{}.yml'.format(product, '_'.join(versions).replace('.', '')),
+            product=product,
+            segments=sorted(segments),
+            hosts=get_hosts(product, segments[-1].hosts, '-'.join(versions), box),
```
ehelms: Are these improperly indented? They appear to be 2 indents deep instead of 1.

bin/forklift
```diff
@@ -0,0 +1,273 @@
...
+            raise click.ClickException('Shiva failed to destroy')
+
+
+@click.group()
+def forklift():
+    pass
+
+
+@forklift.command()
+@click.argument('product', default='katello', type=click.Choice(['katello', 'foreman']))
```
ehelms: At first glance, product seems like an odd choice of words since we refer to these as projects in all other places.

bin/forklift
```diff
@@ -0,0 +1,273 @@
...
+@click.argument('versions', nargs=-1)
+@click.option('--box', default='centos7', help='Box to deploy on')
+@click.option('--versions-config', default=VERSIONS.as_posix(), type=click.Path(),
+              help='Path to the YAML file containing the version configuration')
+@click.option('--pipelines', default=PIPELINES.as_posix(), type=click.Path(),
+              help='Directory where to write the pipeline YAML files')
+@click.option('--destroy', default='success', type=click.Choice(['always', 'success', 'never']),
```
ehelms: I think never is a better default personally. I will try to spin these up for testing locally and get so mad at wasting an hour repeatedly.

bin/forklift
```diff
@@ -0,0 +1,273 @@
...
+              help='When to destroy the created boxes')
+def pipeline(product, versions, box, versions_config, pipelines, destroy):
+    """
+    Run an installation pipeline. This installs a server with the product (either katello or
+    foreman) plus a proxy that's registered to the server. Some bats tests are run to verify it all
+    works as intended. If this all succeeds the boxes are destroyed unless --keep was given.
+    """
+
+    with open(versions_config) as fp:
```
ehelms: What does `fp` stand for?

ehelms:
How does this handle the case where, in an upgrade, I want to run the bats test before hand (to seed some data in) then upgrade with data, then re-run some bats?

bin/forklift
ekohl: namedtuple is pretty much what it says: a tuple but with named for the arguments so you can use ```host.name``` rather than ```host[0]```. I could have made it explicit by assigning the variables in ```__init__``` but ```namedtuple also gives you a usable ```__repr__``` implementation.

bin/forklift
ekohl: It changes the function so you can use it like a property. In this case call it like ```installer._version``` rather than ```installer._version()```. You can also use it to define a setter function so you can control that as well, but that doesn't really make sense in this case.

bin/forklift
ekohl: I agree, I had that for a bit but then I refactored it later. Will fix that up.

bin/forklift
ekohl: Generally not a fan either. I had ```segment``` before but ended up choosing this to limit line length. pep8 doesn't explicitly mention this. I can change ```get_segments``` to do the filtering instead.

bin/forklift
ekohl: You're right, vim on my laptop isn't properly configured.

bin/forklift
ekohl: Naming is one of the hard problems :) I struggled with it and chose product when I was only intending it for foreman or katello. Now that the code supports things like foreman-with-dns or foreman-split-db it no longer makes sense and limiting the choices here doesn't either. Suggestions are very welcome.

bin/forklift
ekohl: That's fair. I do like the delete-on-success model so I can tests multiple pipelines in a loop because my desktop doesn't have the resources to run multiple in parallel. Maybe we also need a halt-on-failure so you can investigate later.

bin/forklift
ekohl: File Pointer. I often use it instead of ```f``` to avoid single letter variables.

ekohl:
Like I said: the upgrade is still work in progress. I'd like a model where you can include roles or playbooks from files and just set variables.

bin/forklift
ekohl: No, I think it's in EPEL7 and there's an SCL. I'm trying to focus on Python 3 but this script can easily be rewritten to support both python 2 and python 3.

ehelms: I just want t make sure this is usable on most developers systems and if not out of the box, document what a user would need to do.

bin/forklift
ehelms: Ahh, so it makes a function more ruby-esque. Cool.

bin/forklift
ehelms: What about just `file` ?

config/versions.yaml
```diff
@@ -36,3 +128,10 @@ installers:
     puppet: 4
     boxes:
       - 'centos7'
+    pipeline:
```
ehelms: Does this define overrides for the base pipeline config?

bin/forklift
ekohl: ```file()``` is a built in function which you'd shadow.

config/versions.yaml
ekohl: Exactly

bin/forklift
ekohl: That makes sense and I'll get on that.

ekohl:
I still think this could be a good idea, but I have no time to work on this.

PR #484, opened 2017-07-07
 Remove org_id from katello client playbook

johnpmitsch:
org_id isn't necessary and is throwing an error since it needs to
have activation_key_id as well.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -210,7 +210,6 @@ awesome-client:
     playbook: 'playbooks/katello_client.yml'
     variables:
       katello_client_server: 'your-katello-server-name'
-      katello_client_organization: 'Default_Organization'
       katello_client_environment: 'Library'
       katello_client_username: 'admin'
       katello_client_password: 'changeme'
--- a/roles/katello_client/defaults/main.yml
+++ b/roles/katello_client/defaults/main.yml
@@ -1,5 +1,4 @@
 ---
-katello_client_organization: "Default_Organization"
 katello_client_environment: "Library"
 katello_client_username: "admin"
 katello_client_password: "changeme"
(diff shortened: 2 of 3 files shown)


ehelms:
@johnpmitsch I don't get how this will work if I have multiple organizations how does it know which environment and organization to use?

johnpmitsch:
@ehelms I was wondering that myself, but somehow it worked for default organization. I'll try again with multiple orgs and comment back

ehelms:
@johnpmitsch ping

johnpmitsch:
@ehelms I think something changed in ansible that is making it impossible to register to a default org in satellite without an activation key. This is the error one currently gets running the client playbook
```
TASK [katello_client : Register client with subscription-manager] **************
Tuesday 25 July 2017  08:56:26 -0400 (0:00:45.427)       0:00:51.319 ********** 
fatal: [almond-client]: FAILED! => {"changed": false, "failed": true, "msg": "parameters are required together: ['activationkey', 'org_id']"}
```
Their [docs](http://docs.ansible.com/ansible/latest/redhat_subscription_module.html) say that org_id is to be used with activation key id	

> Organization ID to use in conjunction with activationkey

I know this makes sense for the portal, but if Katello supports just registering to an organization without an activation key, shouldn't ansible's subscription module support that? Their docs for the sub man module mention Satellite/Katello in them. 

I can raise an issue with them if my train of thought makes sense and I'm not missing something obvious :)

ehelms:
@johnpmitsch agreed, sounds like a bug in their module

johnpmitsch:
@ehelms submitted here https://github.com/ansible/ansible/issues/27283

johnpmitsch:
Closing this until we hear more from ansible

PR #486, opened 2017-07-10
allow scaling of memory and cpu for boxes that define own values

evgeni:
useful for setups where you want to scale up the pre-defined boxes
without touching base_boxes.yaml

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,8 @@ Some settings can be customized for the entirety of the deployment, they are:
 
  * memory: Memory to give boxes by default unless specified by a box
  * cpus: Number of CPUs to give boxes by default unless specified by a box
+ * scale_memory: Factor to multiply memory of boxes that specify an own value
+ * scale_cpus: Factor to multiply CPUs of boxes that specify an own value
  * sync_type: type of sync to use for transfer to the Vagrant box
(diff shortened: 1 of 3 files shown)


ekohl:
I'm not sure how much sense this makes compared the complexity it adds. Is this still needed after https://github.com/theforeman/forklift/pull/485 has been merged?

evgeni:
So the problem I want to solve is that all the pipeline boxes in config/base_boxes.yaml define a very tight memory value. I have enough RAM and want to use it, but without modifying base_boxes.yaml. and the value from settings is ignored as it gets overridden by the box.

README.md
```diff
@@ -131,6 +131,8 @@ Some settings can be customized for the entirety of the deployment, they are:
 
  * memory: Memory to give boxes by default unless specified by a box
  * cpus: Number of CPUs to give boxes by default unless specified by a box
+ * scale_memory: Factor to scale memory of boxes that specify an own value
+ * scale_cpus: Factor to scale CPUs of boxes that specify an own value
```
sean797: Factor to multiply... is clearer IMO.

evgeni: done

ekohl:
Thanks!

ekohl:
Rubocop failures were unrelated.

PR #490, opened 2017-07-11
bats: Move hammer ping to katello tests

ekohl:


ekohl:
@ehelms I'm not sure where the sleep is needed so I'd prefer a separate issue.

PR #491, opened 2017-07-12
Sleep for a minute before testing ping

ehelms:
Sometimes foreman-tasks has not fully started by the time ping runs so lets sleep for a bit before checking the ping.

### DIFF
--- a/bats/fb-test-katello.bats
+++ b/bats/fb-test-katello.bats
@@ -4,7 +4,10 @@
 set -o pipefail
 
 @test "check hammer ping" {
-  hammer ping
+  local next_wait_time=0
+  until hammer ping || [ $next_wait_time -eq 12 ]; do
+     sleep $(( next_wait_time++ ))
+  done
 }
 
 @test "check katello-service status" {


bats/fb-test-katello.bats
```diff
@@ -4,6 +4,7 @@
 set -o pipefail
 
 @test "check hammer ping" {
+  sleep 60
```
ekohl: A slightly more advanced version (based on https://stackoverflow.com/questions/12321469/retry-a-bash-command-with-timeout#12321815) could be:

```bash
@test "check hammer ping" {
  local next_wait_time=0
  until hammer ping || [ $next_wait_time -eq 12 ]; do
     sleep $(( next_wait_time++ ))
  done
}
```

ehelms: So, wake every second, run hammer ping allow it to fail up to 12 times?

ehelms:
Took your idea and updated this with it.

bats/fb-test-katello.bats
ekohl: It's slightly different. It's an increasing backoff since the sleep time is increased every time. So the sleep times are 1, 2, 3, 4, 5. The 12 comes down to roughly 1 minute.

PR #492, opened 2017-07-13
drop bootstrap_vagrant.sh

evgeni:


evgeni:
Nuking them altogether would also work just fine for me ;)

ekohl:
@ehelms I don't think this file is used anymore. Nuke it?

ehelms:
Umm, I still use this sometimes to do some by hand install and setup of bats. Perhaps I should get out of that habit it just works so good :)

evgeni:
On most modern systems you can just do apt/yum/dnf install bats these days. ;)

ehelms:
Fair enough

ehelms:
If you are nuking the file, please update the PR so we can get it merged.

evgeni:
changed the commit to nuke bootstrap_vagrant.sh

ekohl:
Since bats is in Fedora, EPEL6, EPEL7 and Debian Stretch + Jessie backports I think we can change the ansible role to just install the package though we might need to enable Jessie backports first. That's useful anyway because Jessie needs it for Puppet 5.

evgeni:
Yeah. I did not do that change yet because that would mean work for me downstream (RHEL does not ship bats, and we don't enable EPEL on Satellite boxes), but I think that's doable.

ekohl:
We could have an ```install_package.yaml``` and ```install_git.yaml``` where Jessie could use the latter as well. Maybe you can even automatically fall back to a git install if package installation fails.

ekohl:
I include that in https://github.com/theforeman/forklift/pull/510

PR #510, opened 2017-07-28
Improvements to bats/Debian support

ekohl:
With the goal of running a pipeline on Debian this allows using bats from a package (but falling back to git) and allows splitting the packages based on the OS. It also fixes the haveged role.

### DIFF
--- a/roles/bats/tasks/main.yml
+++ b/roles/bats/tasks/main.yml
@@ -11,6 +11,12 @@
     - git
     - sos
 
+- name: "Install bats from package"
+  package:
+    name: "bats"
+    state: installed
+  ignore_errors: yes
+
 - name: "Get /usr/bin/bats stat"
   stat:
     path: "/usr/bin/bats"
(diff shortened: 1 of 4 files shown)


ekohl:
Currently it still fails tests on Debian with Puppetlabs packages because it doesn't appear to load ```/etc/profile.d/puppet-agent.sh``` with bats. @evgeni do you have any idea what the best fix is?

evgeni:
I hate vendors droping stuff to `/etc/profile.d/` and expect shit to work.

We could add something like
```yaml
- name: reset ssh connection to allow user changes to affect 'current login user'
  meta: reset_connection
```
after installing puppetlabs stuff (see http://docs.ansible.com/ansible/latest/meta_module.html, requires ansible 2.3), untested?

ekohl:
I can try that when I'm not on train wifi but that shouldn't hold back this PR IMHO.

evgeni:
`reset_connection` seems not to help here, but as @ekohl said this should not block this PR.

→ ACK

evgeni:
Debian does not load `/etc/profile.d/*` on non-login shells (like the one used by Ansible). So the `puppet` binary is not found after a reconnect either.

ekohl:
@ehelms thoughts?

PR #512, opened 2017-08-02
Unique hostnames via host shortname

ehelms:
This won't make sense on a single, isolated libvirt setup. But for some larger hypervisor environments, or if using a Foreman/Katello to register to for content, having the same hostname used by multiple users can result in missing data.


ehelms:
@ekohl or @evgeni thoughts?

evgeni:
Yes! I wanted to add something like this last week already, but then forgot!

ekohl:
Interesting thought. I was playing with the idea of making the whole domain configurable so maybe the variable should be ```domain```. Then at some point we should add a config option somewhere.

Another thing I'm wondering: how much of this breaks assumptions that the domain is ```example.com``` in our playbooks?

ehelms:
I can't think of anything that assumes `example.com` @ekohl . I can add a domain parameter to make it configurable, but I feel like in the base case this is still a viable option to achieve automatic uniqueness.

lib/forklift/box_distributor.rb
```diff
@@ -67,10 +67,11 @@ def define_vm(config, box = {})
 
         machine.vm.box_url = box.fetch('box_url') if box.key?('box_url')
 
+        host_short_name = `hostname -s`.strip
```
ekohl: I'd prefer ```domain = "#{`hostname -s`.strip}.example.com"```. The configurability can be added later.

evgeni:
We could use `etc_hosts_domain` from `roles/etc_hosts/vars/main.yml`? As that defines the current `example.com` base for a lot of places.

ehelms:
@evgeni use it how?

evgeni:
Argh, ewoud confused me :D

@ekohl suggested to use `{{ ansible_domain }}`, and my braint went ahead and thought "we could also do `{{ etc_hosts_domain | default(ansible_domain) }}` thus alowing the user to override it easily".

But we're in forklift code here, none of the ansible vars are available yet.

ehelms:
Oi, I take that back. Quite a few things are married to `example.com` such as provisioning setup and etc_hosts.

ehelms:
Updated the playbooks and roles.

roles/freeipa_server/defaults/main.yml
```diff
@@ -6,7 +6,7 @@ katello_server_ip: "{{hostvars[katello_server_origin]['ansible_ssh_host']}}"
 
 freeipa_server_ip: "{{ ansible_eth0.ipv4.address }}"
 freeipa_server_realm: "EXAMPLE.COM"
```
ekohl: I don't know FreeIPA but should this be ```{{ ansible_domain|upper }}```?

pipelines/pipeline_smartproxy_powerdns.yml
```diff
@@ -31,7 +31,7 @@
     powerdns_db_pass: "ajooy7HeeCh8loo"
     powerdns_conf_mode: "0644"
     powerdns_zones:
-      - "example.com"
+      - "{{ ansible_domain }}"
```
ekohl: This breaks the bats tests since ```fb-proxy-dns.bats``` hardcodes the domain to ```example.com```.

roles/pulp_repositories/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
```
ekohl: Unrelated change that snuck in with ```git add .````?

roles/freeipa_server/defaults/main.yml
ehelms: Good find, my grepping failed.

pipelines/pipeline_smartproxy_powerdns.yml
ehelms: Hrmm, so should I revert this or update the bats tests to grab the domain?

ekohl: For now I'd revert it since the actual domain doesn't really matter

ehelms:
Fully updated

evgeni:
LGTM!

PR #514, opened 2017-08-14
Use deep_merge on box hashes

ehelms:
This prevents wiping out of ansible configuration when needing to override a parameter. For example:

```
puppet-test:
  box: centos7-katello-nightly
  ansible:
    variable:
      foreman_installer_module_prs: theforeman/puppet/560
```

Without this change, the above would do nothing as the playbook definition and other variables from the base centos7-katello-nightly get wiped out since these are nested key/values.

### DIFF
--- a/lib/forklift/box_factory.rb
+++ b/lib/forklift/box_factory.rb
@@ -2,8 +2,16 @@ require 'json'
 require 'erb'
 require 'yaml'
 
+if Gem.loaded_specs['vagrant']
+  require 'vagrant/util/deep_merge'
+else
+  require 'deep_merge'
+end
+
 module Forklift
   class BoxFactory
+
+    include Vagrant::Util::DeepMerge if Gem.loaded_specs['vagrant']
 
     attr_accessor :boxes, :shells
 
@@ -49,7 +57,7 @@ module Forklift
         box['shell'] += " --installer-options='#{box['installer']}' " if box['shell'] && box['installer']
 
         if @boxes[name]
-          @boxes[name].merge!(box)
+          @boxes[name].deep_merge!(box)
         else
           @boxes[name] = box
         end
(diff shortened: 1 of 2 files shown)


ehelms:
@ekohl mind checking again? I had to get a little creative to allow this in test environment and Vagrant.

lib/forklift/box_factory.rb
```diff
@@ -76,7 +84,7 @@ def process_versions(config, versions)
 
     def layer_base_box(box)
       return box unless (base_box = find_base_box(box['box']))
-      base_box.merge(box)
+      base_box.deep_merge(box)
```
ekohl: I think this should be ```deep_merge!```

ehelms: Thats strange because Vagrant defines only `deep_merge` -- https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/util/deep_merge.rb

ehelms: What verison of Vagrant are you on? Cause this patch has been working fine for me for a few days now.

ehelms:
@ekohl Got some updates in here, and tests added if you want to revisit

ehelms:
@ekohl If you could find time to swing back to this, I'd appreciate. Otherwise I just stay on my branch all the time.

ekohl:
I'm currently using Vagrant 1.9.7 with vagrant-libvirt 0.0.40 under ruby 2.3.3-r1 and I still get the error.

lib/forklift/box_factory.rb
ekohl: It does make it work for me. On line 60 you already use ```deep_merge!```.

ekohl:
I need this now because I need to inherit the python interpreter variable in Ubuntu. After some investigation it looks like the vagrant deep_merge version for me only creates ```deep_merge!``` and not ```deep_merge``` without a bang. You're right that the current code shouldn't need it, but it doesn't. The fallback code doesn't appear to work for me so we should remove that or fix it.

Also needs a rebase now.

ekohl:
Closing since #587 was merged.

PR #518, opened 2017-08-16
Add documentation on installing Vagrant

ekohl:


ares:
:+1:

ekohl:
Merging it because we have a small workshop on using Forklift so this is useful documentation to have. We can always change it later.

PR #521, opened 2017-08-30
[WIP] Add a split katello nightly pipeline

ekohl:
Needs https://github.com/Katello/puppet-katello/pull/209 and https://github.com/Katello/puppet-service_wait/pull/28 and even then it's still a WIP.

### DIFF
--- a/pipelines/pipeline_katello_nightly_split.yml
+++ b/pipelines/pipeline_katello_nightly_split.yml
@@ -73,6 +73,7 @@
     foreman_installer_scenario: katello
     foreman_installer_options_internal_use_only:
       - "--disable-system-checks"
+      - "--certs-generate false"
       - "--katello-manage-application true"
       - "--katello-manage-candlepin false"
       - "--katello-candlepin-hostname pipeline-services-nightly-centos7-split.{{ ansible_domain }}"
--- a/pipelines/pipeline_katello_nightly_split.yml
+++ b/pipelines/pipeline_katello_nightly_split.yml
@@ -1,0 +1,118 @@
+- hosts: localhost
+  vars:
+    forklift_name: pipeline-katello-nightly-split
+    forklift_boxes:
+      pipeline-app-nightly-centos7-split:
+        box: centos7
+        memory: 4096
+        ansible:
+          group: "{{ forklift_name }}"
+      pipeline-services-nightly-centos7-split:
+        box: centos7
+        memory: 4096
+        ansible:
+          group: "{{ forklift_name }}"
+      pipeline-proxy-nightly-centos7-split:
+        box: centos7
+        memory: 3072
+        ansible:
+          group: "{{ forklift_name }}"
+  roles:
+    - forklift
+
+- hosts:
+  - pipeline-app-nightly-centos7-split
+  - pipeline-proxy-nightly-centos7-split
+  - pipeline-services-nightly-centos7-split
+  become: yes
+  vars:
+    foreman_repositories_version: nightly
+    foreman_repositories_environment: staging
+    katello_repositories_version: nightly
+    katello_repositories_environment: staging
+    foreman_installer_skip_installer: true
+    foreman_installer_additional_packages:
+      - foreman-installer-katello
+    foreman_installer_module_prs: 'katello/katello/209,katello/foreman_proxy_content/136'
+  roles:
+    - selinux
+    - etc_hosts
+    - epel_repositories
+    - puppet_repositories
+    - foreman_repositories
+    - katello_repositories
+    - disable_firewall
+    - haveged
+    - update_os_packages
+    - foreman_installer
+
+- hosts: pipeline-services-nightly-centos7-split
+  become: yes
+  vars:
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--no-enable-foreman"
+      - "--no-enable-foreman-cli"
+      - "--no-enable-foreman-plugin-tasks"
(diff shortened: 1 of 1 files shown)


ekohl:
@timogoebel I'm now running into the problem that the app VM doesn't trust the certs from the services VM. @Klaas- mentioned you also looked into this.

timogoebel:
Yes. I'll share our work with you tomorrow. Basically we created a central ca server that generated rpms that we copied to the other servers.
One limitation is, that the certs module just has classes and no defined types. That lead to some duplicate declarations.

pipelines/pipeline_katello_nightly_split.yml
```diff
@@ -0,0 +1,116 @@
...
+  vars:
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--no-enable-foreman"
+      - "--no-enable-foreman-cli"
+      - "--no-enable-foreman-plugin-tasks"
+      - "--no-enable-foreman-proxy"
+      - "--no-enable-foreman-proxy-content"
+      - "--no-enable-foreman-proxy-plugin-pulp"
```
ehelms: We need the three above this or else the services box won't have a proxy running on it (which is required).

ekohl: Good point, hadn't reached a point yet where I could actually use the service.

timogoebel:
@ekohl: You can find our approach in this gist: https://gist.github.com/timogoebel/e4d26249d2b993fd4f084d65cacbed52
Might be an interesting read. I think it's obvious that this is very, very WIP code.

pipelines/pipeline_katello_nightly_split.yml
ekohl: It does mean we have a dependency cycle: smart proxy wants to register to foreman, foreman isn't provisioned yet but needs Candlepin in turn. Deploying a server with Candlepin means we have to fix the CRL in Pulp. A simple crl-updater script might be sufficient.

ekohl:
@timogoebel interesting approach. Generally my goal is very similar but I want a more automated way for the certificates. Some minor improvements:

You can move https://gist.github.com/timogoebel/e4d26249d2b993fd4f084d65cacbed52#file-gui-pp-L125-L129 into an else block so other nodes simply can't execute cron.

https://gist.github.com/timogoebel/e4d26249d2b993fd4f084d65cacbed52#file-qpid-pp-L24-L25 is fixed if you upgrade to katello-qpid 3.1.0.

sean797:
Leaving the certificate issues to one side for a moment, (which can be resolved with https://github.com/Katello/puppet-certs/pull/160 but we're not sure if they should be) there are some oauth key & secrets that need to be the same across nodes. (https://github.com/Katello/puppet-katello/pull/212) I would like to propose we look into merging https://github.com/theforeman/forklift/pull/516 and extend https://github.com/sean797/ansible-role-foreman_installer to ensure they are the same. It already does a very similar thing for oauth options when deploying a Smart Proxy. https://github.com/sean797/ansible-role-foreman_installer/blob/master/tasks/katello-proxy-certs.yml#L125-L127

pipelines/pipeline_katello_nightly_split.yml
```diff
@@ -0,0 +1,116 @@
...
+  vars:
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--katello-manage-application true"
+      - "--katello-manage-candlepin false"
+      - "--katello-candlepin-hostname pipeline-services-nightly-centos7-split.{{ ansible_domain }}"
+      - "--katello-manage-qpid false"
+      - "--katello-qpid-hostname pipeline-services-nightly-centos7-split.{{ ansible_domain }}"
+      - "--katello-manage-pulp false"
```
sean797: This should also include `--no-enable-foreman-proxy --no-enable-foreman-proxy-content --no-enable-foreman-proxy-plugin-pulp`. Though I don't think its currently possible to deploy Katello without a integrated foreman-proxy, I'll test it in the next couple of days.

If foreman-proxy is running on one machine (app) and Pulp on another (service) actions like a manifest refresh will fail as `pulp_url` is created using the foreman-proxies hostname.

https://github.com/Katello/katello/blob/master/app/models/katello/ping.rb#L101-L102
https://github.com/Katello/katello/blob/master/app/lib/katello/capsule_content.rb#L134-L136

pipelines/pipeline_katello_nightly_split.yml
```diff
@@ -0,0 +1,119 @@
...
+      - "--no-enable-foreman-plugin-tasks"
+      - "--no-enable-foreman-proxy"
+      - "--no-enable-foreman-proxy-content"
+      - "--no-enable-foreman-proxy-plugin-pulp"
+      - "--no-enable-puppet"
+      - "--katello-manage-application false"
+      - "--katello-manage-pulp true"
+      - "--katello-manage-candlepin true"
+      - "--katello-manage-qpid true"
+      - "--certs-group root"
```
sean797: I think this needs to be `tomcat`, or candlepin won't start and will error with 

```
 Error injecting constructor, java.io.FileNotFoundException: /etc/pki/katello/private/katello-default-ca.key (Permission denied)

ehelms:
This PR have been open for over a year. @ekohl Do you want to keep this open for future updates ?

ekohl:
While I'd love to finish this, I'm going to close it for now. There have been many changes in the roles and playbooks that it'll be faster to rewrite this when it's ready.

PR #522, opened 2017-09-02
Add Foreman bats CI box

ehelms:
Start of box definition to perform Foreman bats CI the same as Katello. This goes along with some in progress work (https://github.com/theforeman/foreman-infra/pull/323/commits/54958a6d73c59e4b8b265e97b6208af48344a96e).


boxes.yaml.example
```diff
@@ -47,6 +47,17 @@ centos7-bats-ci:
         - "fb-content-katello.bats"
         - "fb-destroy-organization.bats"
         - "fb-finish-katello.bats"
+
+centos7-foreman-bats-ci:
+  box: centos7-bats
+  ansible:
+    playbook: 'playbooks/bats_pipeline_foreman_nightly.yml'
```
ekohl: Should we rename ```playbooks/bats_pipeline_nightly.yml``` to ```playbooks/bats_pipeline_katello_nightly.yml```?

boxes.yaml.example
```diff
@@ -47,6 +47,17 @@ centos7-bats-ci:
...
+    group: 'bats'
+    variables:
+      bats_tests:
+        - "fb-test-foreman.bats"
+        - "fb-finish-katello.bats"
```
ekohl: Should we rename this one to ```fb-finish-foreman.bats``` or just ```fb-finish.bats```?

boxes.yaml.example
```diff
@@ -38,15 +38,26 @@ katello-client:
 centos7-bats-ci:
```
ekohl: It might be nicer to rename this to centos7-katello-bats-ci

ekohl:
I just tried this and I got an error with the swapfile.

```
[root@centos7-foreman-bats-ci ~]# swapon /swapfile
swapon: /swapfile: swapon failed: Invalid argument
```

It refuses because there are holes in it and I think that's because I use btrfs on /var/lib/images

boxes.yaml.example
```diff
@@ -38,15 +38,26 @@ katello-client:
...
   box: centos7-bats
```
ekohl: I don't think the centos7-bats box has much use anymore.

boxes.yaml.example
```diff
@@ -38,15 +38,26 @@ katello-client:
...
   ansible:
-    playbook: 'playbooks/bats_pipeline_nightly.yml'
+    playbook: 'playbooks/bats_pipeline_katello_nightly.yml'
     group: 'bats'
     variables:
```
ekohl: Can we move these variables to the playbook?

boxes.yaml.example
```diff
@@ -38,15 +38,26 @@ katello-client:
...
       bats_tests:
         - "fb-test-foreman.bats"
         - "fb-test-katello.bats"
         - "fb-content-katello.bats"
         - "fb-destroy-organization.bats"
-        - "fb-finish-katello.bats"
+        - "fb-finish.bats"
+
+centos7-foreman-bats-ci:
+  box: centos7-bats
```
ekohl: This can also just use the ```centos7``` box.

boxes.yaml.example
```diff
@@ -38,15 +38,26 @@ katello-client:
...
+  ansible:
+    playbook: 'playbooks/bats_pipeline_foreman_nightly.yml'
+    group: 'bats'
+    variables:
```
ekohl: Can these be moved to the playbook as well?

boxes.yaml.example
ehelms: Note this will require a CI change -- https://github.com/theforeman/foreman-infra/pull/333

ehelms:
This has been updated, but note that -- https://github.com/theforeman/foreman-infra/pull/333 will need to go in at the same time.

PR #523, opened 2017-09-02
Updates for containers to build

ehelms:


containers/roles/foreman/tasks/hacks.yml
```diff
@@ -2,11 +2,11 @@
 - name: 'Add Katello verify SSL'
   get_url:
     url: https://raw.githubusercontent.com/ehelms/katello/verify-ssl-off/app/services/katello/pulp/server.rb
```
ekohl: What's your plan to remove these hacks?

ehelms: They are github issues, I just need to take time to do proper fixes into the respective projects for them (I had sorta forgot about them to be honest).

PR #524, opened 2017-09-07
Split provisioning role

xprazak2:
Split provisioning role into smaller parts
that could be reused downstream.


roles/foreman_provisioning_installer/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: 'install foreman-libvirt'
+  yum: name=foreman-libvirt state=present
```
ekohl: It should be sufficient to add ```--enable-foreman-compute-libvirt``` as an installer option

roles/ruby_libvirt_gem/tasks/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: 'Install libvirt gem'
+  yum: name=tfm-rubygem-ruby-libvirt state=present
```
ekohl: I realise this was present before, but is this really needed? ```foreman-libvirt``` depends on this so it should be redundant.

roles/foreman_provisioning_installer/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+
+- name: 'enable provisioning with foreman-installer'
```
ekohl: Any chance we can re-use the ```foreman_installer``` role here?

roles/foreman_provisioning_infrastructure/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
+---
+- name: 'restart foreman-proxy'
```
ekohl: Starting with a restart proxy and foreman doesn't make sense to me. Since foreman-proxy doesn't need the libvirt gems I think this can safely be removed.

roles/foreman_provisioning_infrastructure/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+  service: name=foreman-proxy state=restarted
+
+- name: 'restart foreman'
```
ekohl: If the installer option is used this can be removed, otherwise it can be moved to where the package is installed.

xprazak2:
Changes:
* using `--enable-foreman-compute-libvirt` now
* reusing `foreman-installer` role
* I kept the ruby-libvirt gem since it is required by libvirt DNS and DHCP providers, but removed the restart tasks since installer is executed and it should do it for us, right?

.gitignore
```diff
@@ -14,7 +14,7 @@ Gemfile.lock
 .ruby-gemset
 *.retry
 *.pyc
-user_playbooks
+user_playbooks/*
```
ekohl: Why is this change needed?

roles/foreman_installer/defaults/main.yml
```diff
@@ -4,6 +4,7 @@ foreman_installer_admin_password: changeme
 foreman_installer_skip_installer: False
 foreman_installer_verbose: True
 foreman_installer_scenario: foreman
+foreman_installer_scenario_flag: '{{ "--scenario" if foreman_installer_scenario != "" else "" }}'
```
ekohl: This feels like it should either be in ```install.yml``` or also used in ```upgrade.yml```.

roles/foreman_installer/tasks/upgrade.yml
```diff
@@ -18,6 +18,5 @@
     --upgrade
     --scenario "{{ foreman_installer_scenario }}"
     --certs-update-all
-    {{ foreman_installer_options }}
-    {{ foreman_installer_options_internal_use_only | join(" ") }}
+    {{ foreman_installer_options }} {{ foreman_installer_options_internal_use_only | join(" ") }}
```
ekohl: nit: please revert this so it's consistent with ```install.yaml```

xprazak2:
I think I figured it out. It is supposed to be `rubygem-ruby-libvirt` not `tfm-rubygem-ruby-libvirt` because proxy does not use gems from sw collection.

So I install it, restart proxy and then continue with running installer.

ekohl:
@xprazak2 that makes sense

xprazak2:
@ekohl, I moved the installation of `rubygem-ruby-libvirt` into a separate role, `foreman_installer_scenario_flag` is used in both `install.yml` and `update.yml`. Anything else I should change?

roles/foreman_installer/defaults/main.yml
ekohl: (done)

roles/restart_proxy/tasks/main.yml
```diff
@@ -0,0 +1,3 @@
+---
```
ekohl: I wonder if this should be its own role. or if this should be in the ```ruby_libvirt_gem``` role as a trigger so it only restarts when it's installed.

xprazak2:
I moved the proxy restart into a handler.

xprazak2:
I found out katello_provisioning does not work properly, changes will be needed.

xprazak2:
It should work as expected now

ares:
what's the status here? @ekohl approved the PR (there was change afterwards though)

ares:
is there something missing in this PR blocking the merge?

xprazak2:
Rebased and fixed merge conflicts.

ehelms:
Odd, sorry @xprazak2 this is still complaining of needing a rebase

xprazak2:
Rebased.

PR #525, opened 2017-09-11
Add SSH agent forwarding, and SSH push only

akofink:
To use these features, add `ssh_forward_agent: True` to settings.yaml, and add the `foreman_installer_trust_github` option to boxes.yaml:

E.g.
```yaml
centos7-devel:
  box: centos7
  ansible:
    group: 'devel'
    variables:
      katello_devel_github_username: '<GitHub username>'
      foreman_installer_trust_github: True
      ssh_forward_agent: True
      foreman_installer_options: "--katello-devel-use-ssh-fork true"
```

### DIFF
--- a/lib/forklift/box_distributor.rb
+++ b/lib/forklift/box_distributor.rb
@@ -65,6 +65,7 @@ module Forklift
       config.vm.define box.fetch('name'), primary: box.fetch('default', false) do |machine|
         machine.vm.box = box.fetch('box_name')
         config.ssh.insert_key = false if SUPPORT_SSH_INSERT_KEY
+        config.ssh.forward_agent = box.fetch('ssh_forward_agent', nil) || @settings.fetch('ssh_forward_agent', false)
         machine.vm.box_check_update = true if SUPPORT_BOX_CHECK_UPDATE
 
         machine.vm.box_url = box.fetch('box_url') if box.key?('box_url')
--- a/roles/foreman_devel/defaults/main.yml
+++ b/roles/foreman_devel/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+foreman_devel_github_push_ssh: False
--- a/boxes.yaml.example
+++ b/boxes.yaml.example
@@ -5,6 +5,8 @@ centos7-devel:
     playbook: 'playbooks/devel.yml'
     group: 'devel'
     variables:
+      ssh_forward_agent: true
+      foreman_devel_github_push_ssh: True
       katello_devel_github_username: <REPLACE ME>
 
 centos7-hammer-devel:
--- a/docs/development.md
+++ b/docs/development.md
@@ -19,7 +19,11 @@ A Katello development environment can be deployed on CentOS 6 or 7. Ensure that 
 
   1. Copy `boxes.yaml.example` to `boxes.yaml`. If you already have a `boxes.yaml`, you can copy the entries in `boxes.yaml.example` to your `boxes.yaml`.
   2. Now, replace `<my_github_username>` with your github username
-  3. Fill in any other options, examples:
+  3. Fill in any ansible options, examples:
+    * `foreman_devel_github_push_ssh`: Force git to push over SSH when HTTPS remote is configured
+    * `ssh_forward_agent`: Forward local SSH keys to the box via ssh-agent
+    * `katello_devel_github_username`: Your GitHub username to set up repository forks
+  4. Fill in any foreman-installer options, examples:
     * `--katello-devel-use-ssh-fork`: will add your fork by SSH instead of HTTPS
     * `--katello-devel-fork-remote-name`: will change the naming convention for your fork's remote
     * `--katello-devel-upstream-remote-name`: will change the naming convention for the upstream (non-fork) repositories remote
@@ -35,11 +39,11 @@ centos7-devel:
     group: 'devel'
     variables:
       katello_devel_github_username: <REPLACE ME>
-      foreman_installer_options_internal_use_only:
-        - "--katello-devel-github-username {{ katello_devel_github_username }}"
-        - "--katello-devel-upstream-remote-name origin"
-        - "--katello-devel-extra-plugins theforeman/foreman_remote_execution"
-        - "--katello-devel-extra-plugins theforeman/foreman_discovery"
+      foreman_installer_options: "
+        --katello-devel-upstream-remote-name origin
+        --katello-devel-extra-plugins theforeman/foreman_remote_execution
+        --katello-devel-extra-plugins theforeman/foreman_discovery
+        "
 ```
 
 Lastly, spin up the box:
--- a/playbooks/devel.yml
+++ b/playbooks/devel.yml
@@ -17,4 +17,5 @@
     - puppet_repositories
     - foreman_repositories
     - katello_repositories
+    - foreman_devel
     - foreman_installer
--- a/roles/foreman_devel/tasks/github_push_ssh.yml
+++ b/roles/foreman_devel/tasks/github_push_ssh.yml
@@ -1,0 +1,11 @@
+---
+- name: "install git"
+  package:
+    name: 'git'
+    state: 'present'
+
+- name: 'Always push to github over ssh'
+  git_config:
+    name: 'url."git@github.com:".pushInsteadOf'
+    value: 'https://github.com/'
+    scope: system
--- a/roles/foreman_devel/tasks/main.yml
+++ b/roles/foreman_devel/tasks/main.yml
@@ -1,0 +1,3 @@
+---
+- include: github_push_ssh.yml
+  when: foreman_devel_github_push_ssh
--- a/settings.yaml.example
+++ b/settings.yaml.example
@@ -7,3 +7,5 @@ sync_type: 'rsync'
 
 cachier:
   mount_options: ['rw', 'vers=3', 'tcp', 'nolock']
+
+ssh_forward_agent: True


roles/ssh_keys/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+ssh_key_name_src: "id_rsa_vagrant"
```
zjhuntin: The assumption being that you'd have a separate SSH key for vagrant?

akofink: Yeah, I didn't know if people would want their actual key copied by default, but you can set this parameter in the boxes.yml.

ekohl: Another approach is to suggest SSH agent forwarding. I recall you can use ```vagrant ssh box -- -A``` or add ```config.ssh.forward_agent``` to the ```Vagrantfile```.

akofink: Wow, agent forwarding is awesome! *But*, I don't think it solves the known_hosts issue, unless you have more information?

ekohl: No, but known_hosts copying is much less of an issue security wise so we could do that by default or just ensure github.com is in the system known hosts.

akofink:
Okay, so here's a way to use SSH agent forwarding and trust github in the known_hosts file. There are many ways to do it, and this might not be the best way. :sheep: Let me know what you think.

roles/foreman_installer/tasks/module_pr.yml
```diff
@@ -32,3 +32,13 @@
   with_items: "{{ foreman_installer_module_prs.split(',') }}"
   tags:
     - packages
+
+- name: 'Trust github'
```
akofink: I couldn't really decide where this should go. Maybe a new task (`pre_install`) rather than here?

ekohl: A new task would be good since even if you don't checkout a module PR then you still want this for the devel git checkouts.

lib/forklift/box_distributor.rb
```diff
@@ -63,6 +63,7 @@ def define_vm(config, box = {})
       config.vm.define box.fetch('name'), primary: box.fetch('default', false) do |machine|
         machine.vm.box = box.fetch('box_name')
         config.ssh.insert_key = false if SUPPORT_SSH_INSERT_KEY
+        config.ssh.forward_agent = true
```
ekohl: I'd prefer if this was configurable and off by default. In most cases agent forwarding is not wanted because of the security risk. Please do add the option to ```boxes.yaml.example``` though.

roles/foreman_installer/tasks/module_pr.yml
```diff
@@ -32,3 +32,13 @@
...
+  blockinfile:
```
ekohl: Would https://docs.ansible.com/ansible/latest/known_hosts_module.html make sense to use?

akofink:
@ekohl I'm trying to get ssh agent forwarding working, and it it works during `vagrant ssh <box>`, but it doesn't seem to work during `vagrant [up|provision] <box>`. Mind testing to see if you're seeing the same behavior?

roles/foreman_installer/tasks/pre_install.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: 'Ensure known_hosts file exists'
```
ekohl: I think https://docs.ansible.com/ansible/latest/file_module.html would be better to use.

roles/foreman_installer/tasks/pre_install.yml
```diff
@@ -0,0 +1,33 @@
...
+  copy:
+    dest: '/home/vagrant/.ssh/known_hosts'
+    content: ''
+    owner: 'vagrant'
+    group: 'vagrant'
+    mode: 0600
+  when: foreman_installer_trust_github
```
ekohl: You could move this to ```main.yml```

lib/forklift/box_distributor.rb
```diff
@@ -63,6 +63,7 @@ def define_vm(config, box = {})
       config.vm.define box.fetch('name'), primary: box.fetch('default', false) do |machine|
         machine.vm.box = box.fetch('box_name')
         config.ssh.insert_key = false if SUPPORT_SSH_INSERT_KEY
+        config.ssh.forward_agent = @settings.fetch('ssh_forward_agent', false)
```
ekohl: Please add ```|| box.fetch('ssh_forward_agent', false)``` so you can also override it at the box level.

roles/foreman_installer/tasks/pre_install.yml
akofink: From those docs:

>>If file, the file will NOT be created if it does not exist, see the copy or template module if you want that behavior.

ekohl:
> @ekohl I'm trying to get ssh agent forwarding working, and it it works during vagrant ssh <box>, but it doesn't seem to work during vagrant [up|provision] <box>. Mind testing to see if you're seeing the same behavior?

I suspect that ansible does its own SSH and it doesn't set agent forwarding: https://github.com/theforeman/forklift/blob/2de65395c5c07b8215f9a7270252d4bc9ca9aa8f/playbooks/inventory/vagrant.py#L53-L56

On an unrelated note: you can use the string option (so not an array) ```foreman_installer_options``` in your ```boxes.yaml``` so you don't have to override the internal options.

roles/foreman_installer/tasks/pre_install.yml
ekohl: The known_hosts module doesn't make any statement about creating the file but if it does, then you can use the file resource after setting the known hosts. That prevents wiping and resetting the keys on every run.

akofink: In the copy module, I need to add `force: no`, and it won't touch the file if it's already there.

roles/foreman_installer/tasks/trust_github.yml
```diff
@@ -0,0 +1,37 @@
+---
+- name: 'Add SSH_AUTH_SOCK to env_keep in sudoers'
+  lineinfile:
+    dest: /etc/sudoers
+    state: present
+    regexp: SSH_AUTH_SOCK
+    line: 'Defaults env_keep += "SSH_AUTH_SOCK"'
```
akofink: So... I think there might be a better way to do this, but since we're using `become: true` in Ansible, the ssh agent forwarding doesn't work without this environment variable. See [this issue](https://github.com/ansible/ansible/issues/7235) in Ansible.

roles/foreman_installer/tasks/trust_github.yml
```diff
@@ -0,0 +1,37 @@
...
+
+- name: 'Ensure known_hosts file exists'
+  copy:
```
ehelms: Would the file module not be more appropriate?

roles/foreman_installer/tasks/trust_github.yml
```diff
@@ -0,0 +1,37 @@
...
+- name: 'Trust 192.30.253.112'
+  known_hosts:
+    name: '192.30.253.112'
+    path: '/home/vagrant/.ssh/known_hosts'
+    key: "192.30.253.112 ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ=="
+    state: present
+
+- name: 'Trust 192.30.253.113'
+  known_hosts:
+    name: '192.30.253.113'
```
ehelms: What does this IP represent?

roles/foreman_installer/tasks/trust_github.yml
```diff
@@ -0,0 +1,37 @@
...
+- name: 'Trust github'
+  known_hosts:
+    name: 'github.com'
+    path: '/home/vagrant/.ssh/known_hosts'
+    key: "github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ=="
+    state: present
+
+- name: 'Trust 192.30.253.112'
+  known_hosts:
+    name: '192.30.253.112'
```
ehelms: What does this IP represent?

roles/foreman_installer/tasks/trust_github.yml
akofink: @ehelms https://github.com/theforeman/forklift/pull/525#discussion_r138335184

roles/foreman_installer/tasks/trust_github.yml
```diff
@@ -0,0 +1,37 @@
...
+    owner: 'vagrant'
+    group: 'vagrant'
+    force: no
+    mode: 0600
+
+- name: 'Trust github'
+  known_hosts:
+    name: 'github.com'
+    path: '/home/vagrant/.ssh/known_hosts'
+    key: "github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ=="
```
ehelms: Variablize the key part if its used multiple places?

roles/foreman_installer/tasks/trust_github.yml
akofink:
```sh
$ nslookup github.com                                                                                                                                                                                                                                                 [7:43:29]
Server:         10.5.30.160
Address:        10.5.30.160#53

Non-authoritative answer:
Name:   github.com
Address: 192.30.253.113
Name:   github.com
Address: 192.30.253.112
```

roles/foreman_installer/tasks/trust_github.yml
ekohl: Another approach is more of a git trick in ~vagrant/.gitconfig
```
[url "git@github.com:"]
        pushInsteadOf = "https://github.com/"
```

This can be set by ```git config --global url."git@github.com:".pushInsteadOf https://github.com/```

Pulling over HTTPS is fine since they're public repos and the installer doesn't need to push.

akofink: So, this would mean I would disable the ssh fork and set this setting to push only over SSH, right?

ekohl: Correct.

roles/foreman_installer/tasks/trust_github.yml
```diff
@@ -0,0 +1,35 @@
+---
```
ekohl: I think this should be a separate role, like ```foreman_devel```

roles/foreman_installer/tasks/trust_github.yml
```diff
@@ -0,0 +1,35 @@
...
+- name: "install git"
+  package:
+    name: 'git'
+    state: 'present'
+
+- name: 'Pull from github over https'
+  git_config:
+    name: 'url."git@github.com:".pushInsteadOf'
+    value: 'https://github.com/'
+    scope: system
```
akofink: I don't like the current implementation, mostly because this 'Pull from github over https' is for an https configuration, not an ssh configuration; however, I would use the https remote now that I know about this config in Git. Because of this paired with SSH agent forwarding, we no longer need to have public RSA keys on the box itself. I'm going to change this PR accordingly and also create a new foreman_devel role to contain this type of logic, as suggested above.

akofink:
This should be good to go.

roles/foreman_devel/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+foreman_devel_github_push_ssh: False
```
ekohl: On last thing: could you add this one to ```boxes.yaml.example``` too?

akofink:
Updated. I also went ahead and changed the Readme to use the proper `foreman_installer_options:` rather than the internal_use_only one.

ekohl:
@akofink are you waiting for anything to merge this?

akofink:
I don't have merge access here.

ekohl:
It used to be that Github showed you that but now it just has a 'Member' badge that tells me nothing :(

PR #528, opened 2017-09-19
Fix rubocop failures

akofink:
Not sure why these failures started, but hopefully this will fix them.


Rakefile
```diff
@@ -9,7 +9,7 @@ end
 begin
   require 'rubocop/rake_task'
   RuboCop::RakeTask.new
-rescue
+rescue Exception
```
ekohl: Should this be a ```LoadError```?

akofink:
Updated

PR #529, opened 2017-09-21
Add optional storage configuration for libvirt provider.

Ntr0:
This commit allows to define additional volumes in the boxes
configuration (e.g. boxes.yml). If using libvirt as provider,
additional volumes will then be created.

### DIFF
--- a/lib/forklift/box_distributor.rb
+++ b/lib/forklift/box_distributor.rb
@@ -204,6 +204,16 @@ module Forklift
         p.memory = box.fetch('memory').to_i * @settings['scale_memory'].to_i if box.fetch('memory', false)
         p.machine_virtual_size = box.fetch('disk_size') if box.fetch('disk_size', false)
 
+        box.fetch('add_disks', []).each do |disk|
+          type = disk.fetch('type', 'raw')
+          device = disk.fetch('device')
+          size = disk.fetch('size')
+          if type.nil? || device.nil? || size.nil?
+            raise "Error in add_disks configuration: type, device or size are missing #{disk}"
+          end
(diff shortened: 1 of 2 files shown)


lib/forklift/box_distributor.rb
```diff
@@ -187,6 +187,20 @@ def configure_libvirt(machine, box, networks = [])
         p.memory = box.fetch('memory').to_i * @settings['scale_memory'].to_i if box.fetch('memory', false)
         p.machine_virtual_size = box.fetch('disk_size') if box.fetch('disk_size', false)
 
+        if box.fetch('add_volumes', false)
```
ekohl: I think this can be slightly simplified to ```box.fetch('add_volumes', []).each do |volume|```

Ntr0: Thanks for the hint.

ehelms:
Can you add a note or warning in the README that these additional volumes will not be logically extended which most people will by default thinks this implies?

ekohl:
Looks like you added boxes.yaml and user_playbooks. There is also a merge conflict in README.md. Would you mind fixing that?

ehelms:
We'd like to keep user_playbooks/.gitkeep in the repository. Also looks like this needs to be rebased.

Ntr0:
So, i messed up a bit with the rebase, but its fine now. What do you think?

ehelms:
@Ntr0 the code looks good, this still is unmergable due to conflicts. I'd recommend trying to squash all these commits into one.

Ntr0:
@ehelms could please take another look?

ehelms:
Thanks @Ntr0 !

PR #531, opened 2017-09-28
Allow a box to be specified primary.

mdellweg:
The present version takes the 'default' parameter of the boxes, which seems to be set on almost all boxes. This clearly contradicts the statement, that only a single box may be specified as primary.

### DIFF
--- a/boxes.yaml.example
+++ b/boxes.yaml.example
@@ -1,5 +1,6 @@
 ---
 centos7-devel:
+  primary: true
   box: centos7
   ansible:
     playbook: 'playbooks/devel.yml'
--- a/config/base_boxes.yaml
+++ b/config/base_boxes.yaml
@@ -6,25 +6,21 @@ boxes:
   debian8:
     box_name:   'debian/jessie64'
     image_name: !ruby/regexp '/Debian.*Jessie/'
-    default:    true
     pty:        true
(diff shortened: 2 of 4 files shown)


mdellweg:
Maybe 'default' has been mistaken for 'autostart'.

ekohl:
Does this mean the ```vagrant (up|ssh|...)``` (without parameters) works on that box?

mdellweg:
Yes, that is exactly the puropse. However `up` seems to be sensitive on `autostart`, whereas `ssh` is sensitive to `primary`. `vagrant halt` stops all boxes.

evgeni:
This still leaves multiple boxes with `default: true` in `config/base_boxes.yaml`.
Is this setting used at all after this patch? quick code-grep says no. I guess it should be removed then, too?

mdellweg:
I tried to `git blame` the history of this default parameter, but the original intention did not become clear to me. If you like, i can remove it.

evgeni:
I *think* it should be dropped, to make the config less confusing.

@ehelms what do you say?

ehelms:
Remove all the defaults and primary IMO. There is no main purpose for this repository and its boxes. If anything, -devel is spun up the most (besides bats-ci box).

PR #534, opened 2017-10-05
[WIP] use ansible debug callback instead of human_log

evgeni:
Ansible ships a debug callback since 2.2, which formats the output like
our human_log, but does not require us carrying a file that needs
updates now and then.

See https://docs.ansible.com/ansible/devel/plugins/callback/debug.html


ekohl:
Neat! Less code to maintain.

PR #535, opened 2017-10-06
Add Foreman 1.16 and Katello 3.5

iNecas:


config/versions.yaml
```diff
@@ -33,6 +33,13 @@ installers:
       - 'centos7'
       - 'debian8'
 
+  - foreman: '1.16'
+    katello: '3.5'
+    puppet: 4
+    boxes:
+      - 'centos7'
+      - 'debian8'
```
ekohl: I think Debian 9 is also supported. @mmoll?

mmoll: yes. 🌀

iNecas: Good catch! Updated

PR #536, opened 2017-10-06
sleep longer for errata

evgeni:


evgeni:
trust me, I don't like it either.

bats/fb-content-katello.bats
```diff
@@ -161,8 +161,11 @@ EOF
   tPackageInstall walrus-0.71 && tPackageExists walrus-0.71
 }
 
+@test "wait for errata to become available" {
+  sleep 60
```
ehelms: This should probably just be a "run" since its not really testing anything.

ekohl: On the other hand, when does sleep fail other than being interrupted?

PR #537, opened 2017-10-09
Add option to override domain, configure sshfs

xprazak2:
2 changes:
* add an option to override domain for a box
* use sshfs to share folders - requires vagrant-sshfs plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ ansible -- updates the Ansible provisioner configuration including the
 libvirt_options -- sets Libvirt specific options
 virtualbox_options -- sets VirtualBox specific options
 rackspace_options -- sets Rackspace specific options
+domain -- forklift uses short name of your host + 'example.com' as domain name for your boxes. You can use this option to override it.
+sshfs -- if you have vagrant-sshfs plugin, you can use sshfs to share folders between your host and guest. See an example below for details.
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -127,6 +129,23 @@ static:
...
+```
+with-sshfs:
+  box: centos7
+  sshfs:
+    host_path: '/some/host/path'
+    guest_path: '/some/guest/path'
+    reverse: True
+```
+
+If you want to mount in the opposite direction, just change `revese` to `False` or remove it entirely.
```
ekohl: There's a typo in reverse.

lib/forklift/box_distributor.rb
```diff
@@ -84,11 +85,24 @@ def define_vm(config, box = {})
...
         configure_rackspace(machine, box)
         configure_synced_folders(machine, box)
+        configure_sshfs(config, box)
 
         yield machine if block_given?
       end
     end
 
+    def create_domain(box)
+      box['domain'] || "#{`hostname -s`.strip}.example.com"
```
ekohl: Would it make sense to also allow overriding this on a settings level?

xprazak2:
Now we can set the domain for all boxes from settings. We can override the settings per-box by adding a domain directly to a box config. I also fixed the typo.

PR #540, opened 2017-10-12
[RFC] Add postgres playbook

akofink:
In the interest of disconnected DB, I present: a DB box.

### DIFF
--- a/roles/postgres/tasks/main.yml
+++ b/roles/postgres/tasks/main.yml
@@ -1,0 +1,54 @@
+---
+- name: 'Install postgresql-server'
+  become: true
+  package:
+    name:
+      - 'postgresql'
+      - 'postgresql-server'
+      - 'python-psycopg2' # required by postgresql_* modules
+    state: 'latest'
+
+- name: 'Initialize postgresql'
+  become: true
+  command: 'postgresql-setup initdb'
+  register: command_result
+  failed_when: command_result.rc != 0 and 'Data directory is not empty!' not in command_result.stdout
+
+- name: 'Enable postgresql on startup'
+  become: true
+  systemd:
+    name: postgresql
+    state: started
+    enabled: True
+
+- name: 'postgresql tasks'
+  become: true
+  become_user: postgres
+  block:
+  - name: 'Create the postgres user'
+    postgresql_user:
+      name: "{{ postgres_db_user }}"
+      password: "{{ postgres_db_password }}"
+  - name: 'Allow password auth to postgres'
+    when: postgres_db_password_auth
+    block:
+    - name: 'Allow passwords on subnet'
+      lineinfile:
+        path: /var/lib/pgsql/data/pg_hba.conf
+        regexp: 'host\s+all\s+all\s+{{ postgres_db_listen_subnet }}\s+(ident|md5)'
+        line: 'host all all {{ postgres_db_listen_subnet }} md5'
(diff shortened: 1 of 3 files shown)


ehelms:
Looks similar to my year old PR -- https://github.com/theforeman/forklift/pull/269

akofink:
Very nice, indeed. Though, this is just a postgres box with no katello/candlepin configuration because I was mainly aiming at katello-devel (which doesn't yet support candlepin DB options). I think it would be really nice to have persistent data between installs and even across boxes.

It's unclear to me why your PR wasn't merged.

ekohl:
> It's unclear to me why your PR wasn't merged.

Lack of time? We have too many open issues and PRs in this repo.

akofink:
Well, this is an RFC, so if one of those C's is 'close it. we don't need it', then that's fine too. :sob:

roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+      - 'postgresql'
+      - 'postgresql-server'
+      - 'python-psycopg2' # required by postgresql_* modules
+    state: 'latest'
+
+- name: 'Initialize postgresql'
+  become: true
+  command: 'postgresql-setup initdb'
+  register: command_result
+  failed_when: command_result.rc != 0 and 'Data directory is not empty!' not in command_result.stdout
```
komidore64: I think the recommended way to check for a non-zero exitstatus would be:
```
failed_when: (command_result | failed) and [...]
```

ekohl: I wonder if I like the version of @ehelms with ```creates``` better or this one.

roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+
+- name: 'Enable postgresql on startup'
+  become: true
+  systemd:
```
evgeni: Any reason not to use the more generic `service` module here?

roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+      regexp: 'listen_addresses = .'
+      line: "listen_addresses = '*'"
+  - name: 'Create databases'
+    postgresql_db:
+      name: "{{ item }}"
+    with_items: "{{ postgres_databases }}"
+
+- name: 'Restart postgresql'
+  become: true
+  systemd:
```
evgeni: Any reason not to use the more generic `service` module here?

roles/postgres/tasks/main.yml
evgeni: I think I like `creates` too.

roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+  block:
+  - name: 'Create the postgres user'
+    postgresql_user:
+      name: "{{ postgres_db_user }}"
+      password: "{{ postgres_db_password }}"
+  - name: 'Allow password auth to postgres'
+    when: postgres_db_password_auth
+    block:
+    - name: 'Allow passwords on subnet'
+      lineinfile:
```
evgeni: I think I'd prefer a template instead of `lineinfile` here.

akofink:
Yeah, let's use a [pre-built container](https://hub.docker.com/r/ansible/postgresql/) instead.

PR #541, opened 2017-10-13
Fixes bats playbook

johnpmitsch:


johnpmitsch:
this is actually updated in master

PR #542, opened 2017-10-16
fix katello hammer ping test

evgeni:


bats/fb-test-katello.bats
```diff
@@ -5,8 +5,12 @@ set -o pipefail
 
 @test "check hammer ping" {
   local next_wait_time=0
-  until hammer ping || [ $next_wait_time -eq 12 ]; do
-     sleep $(( next_wait_time++ ))
+  until hammer ping; do
+    if [ $next_wait_time -eq 12 ]; then
+      # make one last try, also makes the error nice
+      hammer ping
```
ekohl: Since we run in ```set -e``` mode I think this will fail on the first time rather than retry in the loop but I'm not sure.

evgeni: this line is only run once when the counter reached 12, especially to fail here and produce a nice error saying "hammer ping" failed, and output the stdout/stderr.

evgeni: and line 8 does not fail, because of `until`

ekohl: You're right - my bad.

PR #545, opened 2017-10-19
Use curl instead of wget in bats

akofink:
Mostly because base [centos7] systems don't have wget installed.

Found this when trying to run bats on centos7-hammer-devel box foretello PRs to test for breakage.

### DIFF
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -36,7 +36,7 @@ setup() {
 }
 
 @test "upload package" {
-  wget https://repos.fedorapeople.org/repos/pulp/pulp/demo_repos/test_errata_install/animaniacs-0.1-1.noarch.rpm -P /tmp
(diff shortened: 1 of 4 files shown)


ekohl:
Normally we use it via the bats role which has https://github.com/theforeman/forklift/blob/c10243609452845993cdf0a95e39f887e2f5474b/roles/bats/vars/RedHat.yml#L3. Is that an option?

akofink:
I was trying to run bats on devel upstream hammer box (centos7 base) by [just] cloning forklift. This PR is mostly a nitpick about using default packages rather than those you have to install.

ekohl:
In the past we had ```tPackageInstalled wget || tPackageInstall wget``` inside bats files so they could used standalone. I wouldn't mind reintroducing that but using curl is fine too.

akofink:
Up to you. I'm willing to change it but prefer curl.

ekohl:
Let's go your route, but then a step further. You can remove wget from the list of packages to install for bats.

akofink:
Updated

PR #546, opened 2017-10-19
Add a centos7-katello-35 example box

akofink:


ehelms:
This box already exists per -- https://github.com/theforeman/forklift/pull/535

PR #549, opened 2017-10-25
Use Koji repositories for Foreman nightly

ehelms:
This is why nightlies keep failing and will perpetually fail. @ekohl


ekohl:
I had spotted this at some point but forgot about it again.

PR #550, opened 2017-10-26
Add a robotello role

ekohl:
Verified it installed and ran the tests. Since I ran it on a vanilla foreman the actual test failed but it was started.


roles/robottelo/vars/main.yml
```diff
@@ -0,0 +1,7 @@
+---
```
ehelms: I think this should be defaults not vars if the desire is for them to be configurable? Or at least, the ones that we can envision a user wanting to such as test_arguments

ekohl: Good find. I started with copies from downstream, didn't notice it was vars instead of defaults.

PR #552, opened 2017-11-01
Update pipeline playbooks for 3.5 RC2

akofink:


pipelines/pipeline_katello_35.yml
```diff
@@ -4,7 +4,7 @@
     forklift_boxes:
       pipeline-katello-3.5-centos7:
         box: centos7
-        memory: 4680
+        memory: 6000
```
ekohl: Should we use 6144 so it's a nice 6G?

akofink: I don't see why it matters, but I'll go ahead and change it.

akofink:
Updated @ekohl

akofink:
Wait to merge.

PR #560, opened 2017-11-07
Use dd to allocate bats swapfile

jlsherrill:
as some versions of swapon require it to be fully allocated

### DIFF
--- a/roles/swapfile/tasks/main.yml
+++ b/roles/swapfile/tasks/main.yml
@@ -7,7 +7,7 @@
   register: swapfile_check
 
 - name: create swap file {{ swapfile_path }}
-  command: fallocate -l {{ swapfile_size }} {{ swapfile_path }}
+  command: dd if=/dev/zero of={{ swapfile_path }} count={{ swapfile_size }} bs=1MiB
   when: not swapfile_check.stat.exists
 
 - name: set permissions on swap file
--- a/roles/swapfile/defaults/main.yml
+++ b/roles/swapfile/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
 swapfile_path: /swapfile
-swapfile_size: 5120M
+swapfile_size: 5120
--- a/rules/TaskManyArgs.py
+++ b/rules/TaskManyArgs.py
@@ -7,5 +7,8 @@ class TaskManyArgs(AnsibleLintRule):
     tags = ['task']
 
     def match(self, file, text):
+        if text.lstrip().startswith('command:'):
+            return False
+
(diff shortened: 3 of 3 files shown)


roles/swapfile/tasks/main.yml
```diff
@@ -7,7 +7,7 @@
   register: swapfile_check
 
 - name: create swap file {{ swapfile_path }}
-  command: fallocate -l {{ swapfile_size }} {{ swapfile_path }}
+  command: dd if=/dev/zero of={{ swapfile_path }} count={{ swapfile_size }} bs=1MiB
```
jlsherrill: This is failing with `[E303] Use ":" YAML format when arguments are over 3`, i've tried:
```
  command: 
     dd if=/dev/zero of={{ swapfile_path }} count={{ swapfile_size }} bs=1MiB

```

```
  command: 
     "dd if=/dev/zero of={{ swapfile_path }} count={{ swapfile_size }} bs=1MiB"
```

```
  command: "dd if=/dev/zero of={{ swapfile_path }} count={{ swapfile_size }} bs=1MiB"
```

all seem to fail with the same error

akofink: This is a result of the dd syntax. It's the same as yaml syntax and is confusing the linter. I suggest disabling that lint rule for this task.

akofink: You need to use `warn: False` to skip ansible-lint for this task.

jlsherrill: I'm guessing this is a bug in ansible-lint, but i've even tried skipping it via: https://github.com/willthames/ansible-lint#false-positives   and i can't get it to pass 😞

ehelms:
Not clear why its failing to ignore, but I'd propose this modification to the rule since commands can be nearly anything.

```
from ansiblelint import AnsibleLintRule
  
class TaskManyArgs(AnsibleLintRule):
    id = 'E303'
    shortdesc = 'Use ":" YAML format when arguments are over 3'
    description = ''
    tags = ['task']

    def match(self, file, text):
        if "command:" in text:
            return False

        count = len([part for part in text.split(" ") if "=" in part])
        return count > 3
```

jlsherrill:
@ehelms thanks!  That was my thinking as well.

ekohl:
@ehelms Aren't you worried about false positives? Would ```text.lstrip().startswith('command:')``` be better?

ehelms:
Sure, that seems safer

ehelms:
@jlsherrill ping to update

jlsherrill:
@ehelms thanks for the reminder! updated

Issue #498, opened 2017-07-14
Switch Candlepin to use cpdb instead of liquibase


ehelms:

sean797:
@ehelms I think this can be closed https://github.com/theforeman/puppet-candlepin/commit/51ca40909f7ae82bf06f0571df673397daa56201 🎉

ekohl:
I agree. Yay

ekohl:
I now see it's forklift and about containers so this still needs to happen in https://github.com/theforeman/forklift/blob/3538c2e2a60cbb89f67e0db01e28a26bc85326af/containers/roles/candlepin/files/entrypoint.sh#L9-L17

PR #561, opened 2017-11-08
Use cpdb over liquibase

ekohl:
Untested, but copied from puppet-candlepin.

Closes GH-498


ehelms:
I'll give this a test today.

PR #562, opened 2017-11-09
Remove old Foreman and Katello versions

sean797:


roles/foreman_repositories/tasks/redhat_release_repos.yml
```diff
@@ -6,11 +6,3 @@
   tags:
     - packages
   when: foreman_repositories_version != 1.13 or ansible_distribution_major_version != "6"
```
ekohl: This is redundant now.

config/base_boxes.yaml
```diff
@@ -13,11 +13,6 @@ boxes:
     image_name: !ruby/regexp '/Debian.*Stretch/'
     pty:        true
 
-  centos6:
-    box_name:   'centos/6'
-    image_name: !ruby/regexp '/CentOS 6.*PV/'
-    pty:        true
```
ehelms: Technically this could be useful for client testing EL6.

config/base_boxes.yaml
```diff
@@ -13,11 +13,6 @@ boxes:
     image_name: !ruby/regexp '/Debian.*Stretch/'
     pty:        true
 
-  centos6:
```
ekohl: It may be useful to keep this for a bit.

roles/katello_repositories/tasks/main.yml
```diff
@@ -4,6 +4,3 @@
 
 - include_tasks: koji_repos.yml
   when: katello_repositories_use_koji
-
-- include_tasks: qpid.yml
-  when: ansible_distribution_major_version == "6" and katello_repositories_version == 3.2
```
ehelms: Useful for client testing on EL6

ehelms:
Looks generally good, I think I'd just like to keep the ability to client test EL6.

bats/fb-proxy-dns.bats
```diff
@@ -144,13 +144,8 @@ setup() {
...
 
 @test "verify record type CNAME" {
-  if [[ $NAMESERVER == "bind" ]] ; then
-    (
-      [[ $FOREMAN_VERSION == "1.11" ]] ||
-      [[ $FOREMAN_VERSION == "1.12" ]] ||
-      [[ $FOREMAN_VERSION == "1.13" ]] ||
-      [[ $FOREMAN_VERSION == "1.14" ]]
-    ) && skip "CNAME records are only supported on 1.15+"
+  if [[ $NAMESERVER == "bind" ]] && [[ $(echo "$FOREMAN_VERSION < "1.15"" | bc) == 1 ]] ; then
```
ekohl: The quoting here looks weird

roles/katello_repositories/tasks/main.yml
sean797: Really? That `when` statement implies only Katello 3.2, which this PR removes.

ehelms: Ahh, it was needed for 3.2 on RHEL6, but it is also required for a client on EL6.

sean797:
Thanks for reviewing guys! Updated 👍

ekohl:
Looks like 3.3 also fails so mind removing that too?

sean797:
sure, updated.

ekohl:
@ehelms any objections left?

PR #563, opened 2017-11-13
Add conditional path testing for Travis

ehelms:
This is a WIP to see only run some testing depending on the changes made.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,25 @@ language: python
 python:
   - 2.7
 install:
-  - pip install ansible-lint
-  - rvm install 2.2.4
+  - |
+    if git diff origin/master --name-only | grep -q "lib\|test"; then
+      rvm install 2.2.4
+    fi
+
+  - |
+    if git diff origin/master --name-only | grep -q "playbooks\|roles\|containers/roles"; then
+      pip install ansible-lint
+    fi
 script:
-  - bundle install
-  - bundle exec rake
-  - ansible-lint -r rules/ playbooks/*
-  - ansible-lint -r rules/ roles/*
-  - ansible-lint -r rules/ containers/roles/*
+  - |
+    if git diff origin/master --name-only | grep -q "lib\|test"; then
+      bundle install
+      bundle exec rake
+    fi
+
+  - |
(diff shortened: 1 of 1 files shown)


ehelms:
If there is a cleaner way to determine and perform these checks, I am open to the idea. For now this aims to make this smarter, and faster.

.travis.yml
```diff
@@ -3,11 +3,29 @@ language: python
 python:
   - 2.7
 install:
-  - pip install ansible-lint
-  - rvm install 2.2.4
+  - changes=$(git diff origin/master --name-only | grep "lib\|test" || true)
```
ekohl: Doesn't the ```|| true``` defeat the purpose?

.travis.yml
```diff
@@ -3,11 +3,29 @@ language: python
...
+  - |
+    if [[ ${changes[@]} ]];then
```
ekohl: Does this test the length? Wouldn't ```grep -q``` and the exit code work just as well?

.travis.yml
```diff
@@ -3,11 +3,29 @@ language: python
...
+      rvm install 2.2.4
+      fi
```
ekohl: Odd alignment

.travis.yml
ehelms: Nope -- without the `|| true` if there is no changes, the grep will throw an `exit 1` which marks the build as failed. Whereas with it, this will capture the output of the grep and prevent it failing so the `if` check can execute.

.travis.yml
ehelms: It does. I was not able to get the latter concept to work within the confines of Travis and it's signal checking.

.travis.yml
ehelms: Will fix

ehelms:
updated

.travis.yml
ekohl: I wonder if ```if git diff origin/master --name-only | grep-q "lib\|test" ; then``` would work but if this works then so be it.

ehelms:
That did indeed work @ekohl -- updated!

PR #566, opened 2017-11-14
[WIP] Add container testing

ehelms:


containers/install-ansible-container.yml
```diff
@@ -12,13 +12,8 @@
...
 
-    - name: 'Install python3-devel'
-      become: true
-      yum:
-        name: python3-devel
-
     - name: 'Install ansible-container'
       become: true
-      command: 'pip3 install -e .[docker,openshift]'
+      command: 'pip install -e .[docker,openshift] --upgrade'
```
ekohl: I tried to use the native pip module but couldn't get it to work. Have you tried it? Another thing I tried was to just ```pip install -e https://github.com/ansible/ansible-container.git[docker,openshift]``` but don't recall why it didn't work. I think it was because it always upgraded but with ```--upgrade``` it might be redundant.

containers/ansible.cfg
```diff
@@ -1,5 +1,5 @@
 [defaults]
-vault_password_file = .vault_pass
+#vault_password_file = .vault_pass
```
ekohl: Is this no longer needed?

ehelms: I need to re-examine where an dhow it's needed. For the moment I commented it out since it was making life more difficult.

containers/install-ansible-container.yml
ehelms: The native pip module? As in pip whatever version of python pip ends up pointed at?

ehelms:
This is step 1 of container testing to at least check that changes build. The next step, given some blog posts I have seen, is to see if we can run minishift inside Travis and do a live test.

containers/install-ansible-container.yml
ekohl: https://docs.ansible.com/ansible/latest/pip_module.html

ehelms:
I've done a major update to this which not only builds the containers, but spins up Openshift, deploys the containers to it and then attempts to test that the UI comes up and that all backend services are running via ping. This is not an extensive test but I think sets a good benchmark for this early work and changes to it that it can build, deploy and services come up. This is currently using a crude Ruby script to do the testing and should likely be moved to pytest or Robottelo. I can either do that as part of this change, or make an issue to do that later on. I wanted to do the simplest test setup I could such that any contribution such as (https://github.com/theforeman/forklift/pull/561) can get tested and in sooner.

containers/roles/certificates/tasks/ca.yml
```diff
@@ -1,8 +1,3 @@
-- name: 'Install openssl'
-  yum:
```
ekohl: Is this no longer needed or are you now assuming it's present? On the latter it might be better to use ```package``` instead.

ehelms: This annoyingly causes breakages on Travis due to it having to use the apt module. I can add it back and you can see the error. My googling indicated no one had solved it.

ehelms:
Cleaned up the 0001.patch file.

ehelms:
See -- https://github.com/theforeman/forklift/pull/605

PR #568, opened 2017-11-21
Kill shells

ekohl:
I should note that for me it's not rsyncing /vagrant (though it says it does) meaning it doesn't actually work for me. That makes me wonder if we should just convert this last shell to ansible.

### DIFF
--- a/config/base_boxes.yaml
+++ b/config/base_boxes.yaml
@@ -1,7 +1,3 @@
-shells:
-  runcible_shell: &runcible_shell
-    shell: '/vagrant/shells/runcible.sh'
-
 boxes:
   debian8:
     box_name:   'debian/jessie64'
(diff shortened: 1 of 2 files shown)


config/base_boxes.yaml
```diff
@@ -50,13 +46,7 @@ boxes:
         group: 'freeipa_server'
         server: 'centos7-katello-nightly'
 
-  centos7-bats:
-    box: centos7
-    ansible:
-      playbook: 'playbooks/bats_pipeline_nightly.yml'
-      group: 'bats'
-
```
beav: is this removal intentional? I'm not sure how it is related to the shells removal

ekohl: It can be removed but you're right that it doesn't belong in this PR.

beav:
ACK from me, thanks for getting rid of this extra piece of abstraction.

ekohl:
Rebased to re-add the bats box. I'll have a look at cleaning up the base boxes after https://github.com/theforeman/forklift/pull/393 is merged.

beav:
ACK again!

PR #569, opened 2017-11-21
Make the swapfile role idempotent

ekohl:


roles/swapfile/tasks/main.yml
```diff
@@ -26,8 +26,13 @@
...
     state: present
 
-- name: swapon {{ swapfile_path }}
+- name: get active swap
+  command: swapon --summary | grep '^{{ swapfile_path }}'
+  register: swapfile_active
+
+- name: activate {{ swapfile_path }}
   command: swapon {{ swapfile_path }}
+  when: swapfile_active.rc != 0
```
ekohl: Looks like this doesn't work. Is there a native way to grep in ansible?

ehelms: You can do `swapfile_active | failed`

ekohl: Doesn't appear to help. Somehow the activation is still skipped so the command somehow succeeds. Perhaps it needs to be quoted because of the pipe?

ehelms:
LGTM -- is this still WIP?

ekohl:
Yes, somehow the grep is never applied. Do I need to quote it?

ehelms:
You need to switch to using `shell` as command expects a single command and nothing piped together.

ekohl:
That worked

PR #570, opened 2017-11-28
Improve Debian support

ekohl:
See individual commits


config/base_boxes.yaml
```diff
@@ -3,27 +3,47 @@ boxes:
     box_name:   'debian/jessie64'
     image_name: !ruby/regexp '/Debian.*Jessie/'
     pty:        true
+    scenarios:
+      - foreman
```
ehelms: Do we want the base boxes to have scenarios?

config/base_boxes.yaml
```diff
@@ -3,27 +3,47 @@ boxes:
...
 
   debian9:
     box_name:   'debian/stretch64'
     image_name: !ruby/regexp '/Debian.*Stretch/'
     pty:        true
+    scenarios:
+      - foreman
```
ehelms: Do we want the base boxes to have scenarios?

config/base_boxes.yaml
```diff
@@ -3,27 +3,47 @@ boxes:
...
 
   centos6:
     box_name:   'centos/6'
     image_name: !ruby/regexp '/CentOS 6.*PV/'
     pty:        true
+    scenarios:
+      - foreman
+      - katello
```
ehelms: Do we want the base boxes to have scenarios? For example, what if I just want a centos6 box?

config/base_boxes.yaml
```diff
@@ -3,27 +3,47 @@ boxes:
...
 
   centos7:
     box_name:   'centos/7'
     image_name: !ruby/regexp '/CentOS 7.*PV/'
     pty:        true
+    scenarios:
+      - foreman
+      - katello
```
ehelms: Do we want the base boxes to have scenarios?

config/base_boxes.yaml
```diff
@@ -3,27 +3,47 @@ boxes:
...
     box_name: 'fedora/26-cloud-base'
     image_name: !ruby/regexp '/Fedora 26.*/'
     pty: true
 
+  ubuntu1604:
+    box_name: 'generic/ubuntu1604'
+    image_name: !ruby/regexp '/Ubuntu.*16\.04/'
+    pty: true
+    scenarios:
+      - foreman
```
ehelms: Do we want the base boxes to have scenarios?

roles/foreman_repositories/tasks/main.yml
```diff
@@ -1,12 +1,5 @@
 ---
-- include_tasks: debian_release_repos.yml
-  when: ansible_os_family == 'Debian' and foreman_repositories_use_release
-
-- include_tasks: redhat_release_repos.yml
-  when: ansible_os_family == 'RedHat' and foreman_repositories_use_release and not foreman_repositories_use_koji
-
-- include_tasks: koji_repos.yml
-  when: foreman_repositories_use_koji
+- include_tasks: "{{ ansible_os_family|lower }}_{{ foreman_repositories_flavor }}_repos.yml"
```
ehelms: I don't guess there is a 'el' or 'enterprise linux' "family" ? 'redhat' sounds specific when you are considering there is an RH product based on these projects.

roles/foreman_repositories/defaults/main.yml
```diff
@@ -1,4 +1,5 @@
 ---
 foreman_repositories_version: nightly
-foreman_repositories_use_release: True
-foreman_repositories_use_koji: False
+foreman_repositories_flavor: release
```
ehelms: Maybe comment the other option?

roles/debian_backports/vars/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+backports_distribution: "{{ansible_distribution_release}}"
```
ehelms: I think to follow normal convention you'd want these to be 'debian_backports` to align with the role name.

config/base_boxes.yaml
ekohl: I wasn't sure where to put it otherwise. Also note that this is just a variable that's later read by the BoxFactory to generate the scenario boxes based on it.

roles/foreman_repositories/tasks/main.yml
ekohl: Yes, it's the same with Debian (osfamily) vs Debian (distro). There the situation is easier because they use dist release names which are unique. This doesn't handle Fedora but it didn't do so before either. This is IMHO where something like Hiera could be nice and I'd gladly hear of better patterns to solve multi-distro tasks. It's something I think is the weakest point of Ansible.

roles/foreman_repositories/defaults/main.yml
ekohl: You mean use_koji or describe what it is in a comment?

roles/debian_backports/vars/main.yml
ekohl: You're right, I just copied and made minor style changes to this module but sounds like a good modification for consistency.

ekohl:
Could use another look. I also wonder if we should rename ```katello_repositories_use_koji``` too.

config/base_boxes.yaml
ehelms: Oh, hrmm, what is a "scenario" again in this context?

ekohl: Pretty much what the installer considers a scenario. I do derive that the content proxy scenario is available if the katello one is since we set the server property there.

ekohl: https://github.com/ekohl/forklift/blob/92b52547ece66a65786cfd0cf91549743fcc1cbc/lib/forklift/box_factory.rb#L55-L67 is the section that processes it.

ehelms:
I think making `katello_repositories` follow a similar structure of naming would be good.

ekohl:
Updated to add the katello and katello_client renames.

PR #574, opened 2017-12-01
refs(572) - Support virtualenv for install ansible-container

ehelms:
I had to revert to using command given the special nature in which ansible-container installs, seems the pip module doesn't always like the engine specification piece (i.e. [docker,openshift])

### DIFF
--- a/containers/install-ansible-container.yml
+++ b/containers/install-ansible-container.yml
@@ -2,8 +2,21 @@
 - hosts: localhost
   gather_facts: no
   tasks:
+    - set_fact:
+        inside_virtual_env: "{{ lookup('pipe', 'echo $VIRTUAL_ENV') }}"
+
+    - name: 'Create .tmp directory'
+      file:
+        path: .tmp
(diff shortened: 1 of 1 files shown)


ekohl:
Would passing ```virtualenv: ...``` be a cleaner solution?

ehelms:
Generally yes, but in this case we can detect it so why not take that extra step out of users hands?

ekohl:
I'd expect that then you don't need the temporary directory to clone from.

ehelms:
I don't follow? Without cloning it, I can't install it from git.

ekohl:
The previous pip install didn't need the whole temporary directory and I wonder if we can avoid the whole ```.tmp && git clone && pip install```.

ehelms:
I should have clarified, in my initial testing, the pip install worked. On subsequent retries and environments it failed hard which is part of the reason for the revert.

ehelms:
@ekohl mind re-reviewing today?

PR #575, opened 2017-12-04
Use puppet 5 in 1.16

ekohl:
For Debian Stretch there are no puppet 4 packages and we support 5.


mmoll:
this also fixes problems on Ubuntu/xenial

PR #578, opened 2017-12-15
Add puppet bats tests

ekohl:
Ported from foreman-bats with some cleanups.


ehelms:
Should we drop some of the same tests from https://github.com/theforeman/forklift/blob/64821e72eddd7a74c18342e3691411f30beffc55/bats/fb-test-foreman.bats ?

PR #581, opened 2018-01-09
Turn Runcible test into playbook

ehelms:


ehelms:
Note this is using Pulp repositories per a change we are planning to make to use them directly from here out.

playbooks/runcible.yml
```diff
@@ -0,0 +1,105 @@
...
+        baseurl: http://cbs.centos.org/repos/atomic7-testing/x86_64/os/
+        gpgcheck: no
+        enabled: yes
+      become: true
+
+    - name: 'Install packages'
+      yum:
+        name: "{{ item }}"
+        state: "present"
+        update_cache: true
```
ekohl: I wonder if this means it will refresh the cache for every listed package or just the first one.

playbooks/runcible.yml
```diff
@@ -0,0 +1,105 @@
...
+      command: "gem install librarian-puppet"
+      args:
+        chdir: "/home/vagrant"
+
+    - name: "Clone puppet-pulp"
+      git:
+        repo: https://github.com/katello/puppet-pulp.git
+        dest: "/home/vagrant/puppet-pulp"
+
+    - name: "Deploy Puppetfile"
```
ekohl: Can't librarian-puppet install it directly from a git repo?

playbooks/runcible.yml
```diff
@@ -0,0 +1,105 @@
...
+        path: /etc/pulp/server.conf
+        mode: "0644"
+
+    - name: "Clone runcible"
+      git:
+        repo: https://github.com/katello/runcible.git
+        dest: "/home/vagrant/runcible"
+
+    - name: "Install bundler"
+      command: "gem install bundler"
```
ekohl: https://docs.ansible.com/ansible/latest/gem_module.html would be cleaner I think

playbooks/runcible.yml
```diff
@@ -0,0 +1,105 @@
...
+        - ostree
+
+    - name: "Install RVM gpg key"
+      shell: "gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3"
+
+    - name: "Install RVM"
+      shell: "curl -sSL https://get.rvm.io | bash -s stable --ruby"
+
+    - name: "Install librarian-puppet"
+      command: "gem install librarian-puppet"
```
ekohl: https://docs.ansible.com/ansible/latest/gem_module.html would be cleaner I think

playbooks/runcible.yml
```diff
@@ -0,0 +1,105 @@
...
+        - git
+        - nss
+        - puppet
+        - glib2
+        - ostree
+
+    - name: "Install RVM gpg key"
+      shell: "gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3"
+
+    - name: "Install RVM"
```
ekohl: Is RVM actually used? I don't see anything activating it.

ehelms: Aye it is, bet deceptive perhaps but without this it does not work.

playbooks/runcible.yml
ehelms: Does not appear so

playbooks/runcible.yml
ehelms: Didn't work when using RVM.

ehelms:
Updated code

playbooks/runcible.yml
ekohl: I'm guessing it installs something into the bash login profile and ansible uses that implicitly.

playbooks/runcible.yml
ekohl: That's odd because we have git repos in our installers' Puppetfiles.

playbooks/runcible.yml
```diff
@@ -0,0 +1,104 @@
...
+        content: "forge \"https://forgeapi.puppetlabs.com\"\nmetadata"
+        dest: "/home/vagrant/puppet-pulp/Puppetfile"
+
+    - name: "Install Puppet modules"
+      shell: "/home/vagrant/bin/librarian-puppet install"
+      args:
+        chdir: "/home/vagrant/puppet-pulp"
+      remote_user: "vagrant"
+
+    - name: "Link pulp puppet module"
```
ekohl: Is this still needed with the new Puppetfile w/ metadata?

ekohl:
I think you can also remove ```runcible.sh```

playbooks/runcible.yml
ehelms: Right - but I don't think you can point librarian-puppet directly at git. The CLI does not support this.

playbooks/runcible.yml
ehelms: Without this link you run into --  Error while evaluating a Resource Statement, Could not find declared class pulp at /home/vagrant/test.pp:1:1

ehelms:
@ekohl Any further requests here?

PR #583, opened 2018-01-15
Switch staging repositories to use pulp repos

ehelms:
No longer use the Koji repositories for staging. This should switch the pipeline over to these as well which is needed for it to pass.

### DIFF
--- a/roles/katello_repositories/tasks/staging_repos.yml
+++ b/roles/katello_repositories/tasks/staging_repos.yml
@@ -32,7 +32,7 @@
   yum_repository:
     name: pulp-koji
     description: "Pulp {{ katello_repositories_version }} Koji Repository"
-    baseurl: "http://koji.katello.org/releases/yum/katello-{{ katello_repositories_version }}/pulp/el{{ ansible_distribution_major_version }}/x86_64/"
+    baseurl: "https://repos.fedorapeople.org/repos/pulp/pulp/{{ kaatello_repositories_pulp_release }}/{{ katello_repositories_pulp_version }}/{{ ansible_distribution_major_version }}Server/x86_64/"
     priority: 1
     gpgcheck: 0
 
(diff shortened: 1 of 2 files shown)


roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -33,6 +33,7 @@
     name: pulp-koji
     description: "Pulp {{ katello_repositories_version }} Koji Repository"
     baseurl: "http://koji.katello.org/releases/yum/katello-{{ katello_repositories_version }}/pulp/el{{ ansible_distribution_major_version }}/x86_64/"
```
ekohl: Shouldn't this line be removed?

ekohl:
Related: do we need to modify repoclosure too?

roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -33,6 +33,7 @@
...
+    baseurl: "https://repos.fedorapeople.org/repos/pulp/pulp/{{ kaatello_repositories_pulp_release }}/{{ katello_repositories_pulp_version }}/{{ ansible_distribution_major_version }}Server/x86_64/"
```
ekohl: Shouldn't ```x86_64``` be ```$basearch```?

ehelms: It can be, nothing else in this file uses it though, so I am being consistent.

ehelms:
Good catch on repoclosure -- https://github.com/theforeman/foreman-packaging/pull/2098

ehelms:
@ekohl Updated

ekohl:
At some point we should split off pulp_repositories which you've already done in https://github.com/theforeman/forklift/pull/581. Would it make sense to do it right straight away?

ehelms:
I thought about that, and I am torn still given you need Pulp to install. This follows the current model of the Katello release repo RPM that configures the Pulp RPM so I'd prefer to leave it this way until I think more.

PR #587, opened 2018-01-16
Deep merge box settings

ekohl:
This allows setting variables in base boxes and overriding a different set in variables on higher levels.

It's based on https://github.com/theforeman/forklift/pull/514

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@
 source 'https://rubygems.org'
 
 group :test do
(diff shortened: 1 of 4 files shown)


ehelms:
Looks like you are hitting the same problem. In the test environment, vagrant is not present so there is no function deep_merge to call on Vagrant.

ekohl:
Updated

PR #588, opened 2018-01-16
Expand functionality needed to run systests

ekohl:
This is part of what foreman-bats currently offers.

### DIFF
--- a/bats/fb-test-puppet.bats
+++ b/bats/fb-test-puppet.bats
@@ -6,6 +6,10 @@ set -o pipefail
 load os_helper
 load foreman_helper
 
+if [[ -e /etc/profile.d/puppet-agent.sh ]] ; then
+  . /etc/profile.d/puppet-agent.sh
+fi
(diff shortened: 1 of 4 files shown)


bats/fb-test-puppet.bats
```diff
@@ -6,6 +6,10 @@ set -o pipefail
 load os_helper
 load foreman_helper
 
+if [[ -e /etc/profile.d/puppet-agent.sh ]] ; then
+  . /etc/profile.d/puppet-agent.sh
```
ehelms: Where does this file come from?

ekohl: The puppet package installs this. Since all the binaries are in /opt/puppetlabs/bin this ensures ```$PATH``` is set up correctly. It's a workaround since non-interactive shells don't source ```/etc/profile.d/*.sh``` and I wasn't sure how to do this otherwise when running via ansible.

PR #590, opened 2018-01-19
Add hammer tests

ekohl:


ekohl:
Note I left part of it out since it wasn't used and somewhat ugly. https://community.theforeman.org/t/conversion-of-hammer-tests-in-systest-to-forklift/8093

roles/foreman_testing/tasks/main.yaml
```diff
@@ -0,0 +1,10 @@
+---
+- name: "Run bats tests"
+  include_role:
```
evgeni: I think you can also define `meta/main.yml` with
```
dependencies:
  - { role: 'bats', when: foreman_testing_bats }
  - { role: 'hammer_tests', when: foreman_testing_hammer_tests }
```

Not sure what is best, tho. (https://docs.ansible.com/ansible/latest/playbooks_reuse_roles.html says "You can conditionally execute a role. This is not generally recommended with the classic syntax, but is common when using `import_role` or `include_role`")

evgeni:
Looks correct on first glance, will give it a few test rounds now.

ekohl:
> The hammer tests are disabled on purpose? If so, ACK ;)

They are indeed since that matches the current behavior. They also need orgs and locs enabled which is not true on the default vanilla foreman.

PR #591, opened 2018-01-20
Replace foreman-tasks with dynflow service and other fixes

ehelms:


containers/roles/dynflow/files/start-dynflowd.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+/usr/sbin/dynflowd --foreman-root=/usr/share/foreman run
```
ekohl: Feels like a bit of overhead to use a bash script just to run it.

ehelms: What would you consider using instead?

ekohl: Isn't it an option to do this directly in the yaml file?

ehelms: It is, however, best practices I have seen is to keep the logic in a bash script even if its simplistic. I think to make the role more complete and to allow easier re-factoring of the start script (or if it becomes multiline)

ehelms:
I'll look at #561

PR #593, opened 2018-01-22
Add information and playbooks for katello 3.6

zjhuntin:


config/versions.yaml
```diff
@@ -16,6 +16,15 @@ installers:
       - 'debian9'
       - 'ubuntu1604'
 
+  - foreman: '1.17'
+    katello: '3.6'
+    puppet: 5
+    boxes:
+      - 'centos7'
+      - 'debian8'
```
evgeni: I *think* foreman 1.17 does not support Debian 8 anymore, but @mmoll can confirm that

evgeni:
LGTM besides the Debian 8 comment.

config/versions.yaml
ekohl: It was indeed dropped.

pipelines/pipeline_katello_35_to_36_upgrade.yml
```diff
@@ -0,0 +1,106 @@
...
+      - "--disable-system-checks"
+      - "--foreman-proxy-trusted-hosts {{ server_fqdn.stdout }}"
+      - "--foreman-proxy-trusted-hosts {{ ansible_nodename }}"
+      - "--foreman-proxy-foreman-base-url https://{{ server_fqdn.stdout }}"
+      - "--foreman-proxy-register-in-foreman true"
+      - "--foreman-proxy-oauth-consumer-key {{ oauth_consumer_key.stdout }}"
+      - "--foreman-proxy-oauth-consumer-secret {{ oauth_consumer_secret.stdout }}"
+      - "--foreman-proxy-content-certs-tar {{ foreman_proxy_content_certs_tar }}"
+      - "--foreman-proxy-content-parent-fqdn {{ server_fqdn.stdout }}"
+      - "--foreman-proxy-content-pulp-oauth-secret {{ pulp_oauth_secret.stdout }}"
```
ekohl: This line should be dropped similar to https://github.com/theforeman/forklift/pull/594

pipelines/pipeline_katello_36.yml
```diff
@@ -0,0 +1,83 @@
...
+      - foreman-installer-katello
+  roles:
+    - foreman_proxy_content
+    - foreman_installer
+
+- hosts: pipeline-katello-3.6-centos7
+  become: true
+  vars:
+    bats_tests:
+      - fb-test-katello.bats
```
ehelms: I think we may want to add the `fb-test-puppet.bats` here

zjhuntin: Added

ekohl:
Thanks!

PR #603, opened 2018-02-08
Add F27 box, small cleanup of dup boxes

jturel:
Thought it might be useful to have an F27 box. Would be for me at least!

Also cleaned up what I believe to be duplicate box definitions (fedora26, ubuntu1604) - let me know if I am not understanding their existence properly.


jturel:
Looks like we have a virtually identical PR open!

PR #609, opened 2018-02-09
Update webpack documentation

ekohl:
Needs https://github.com/theforeman/puppet-katello_devel/pull/96 to land in nightlies first.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -40,7 +40,6 @@ centos7-devel:
     variables:
       katello_devel_github_username: <REPLACE ME>
       foreman_installer_options: "
-        --katello-devel-upstream-remote-name origin
         --katello-devel-extra-plugins theforeman/foreman_remote_execution
         --katello-devel-extra-plugins theforeman/foreman_discovery
         "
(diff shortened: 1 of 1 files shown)


ekohl:
Related PR was merged. I also think we should add a post message to katello-installer telling the users which URLs are now available.

ekohl:
https://github.com/Katello/katello-installer/pull/589 should improve that

docs/development.md
```diff
@@ -229,38 +233,6 @@ then add
...
-Replace centos7-devel.example.com with the public hostname of your development server, if different.
-
-Remember to run `source ~/foreman/.env` before running your Webpack server, so that the right options to enable HTTPS are loaded.
-
-Additionally, Foreman uses [secureheaders](https://github.com/twitter/secureheaders) to protect users from different attacks. secureheaders will block requests to the webpack development server, as it's not running in localhost:3000. To allow it, change the following lines in `~/foreman/config/initializers/secure_headers.rb`, substituting centos7-devel.example.com for your fqdn, you can get this value by running `hostname -f`.
-
-```
-    :style_src   => ["'unsafe-inline'", "'self'", 'centos7-devel.example.com:3808'],
-    :script_src  => ["'unsafe-eval'", "'unsafe-inline'", "'self'", 'centos7-devel.example.com:3808'],
-```
```
ehelms: This is handled by puppet-katello_devel or just old information?

docs/development.md
```diff
@@ -229,38 +233,6 @@ then add
...
-
-```yaml
-:webpack_dev_server: true
-:webpack_dev_server_https: true
-```
-
-This will instruct Foreman to use Webpack from an https source. Now we need to make sure Webpack is using the right certificates to serve the development content through HTTPS. We're going to save these configuration values on `~/foreman/.env`:
-
-```
-WEBPACK_OPTS='--https --key /etc/pki/katello/private/katello-default-ca.key --cert /etc/pki/katello/certs/katello-default-ca.crt --cacert /etc/pki/katello/certs/katello-default-ca.crt --host 0.0.0.0 --public centos7-devel.example.com'
```
ehelms: This is handled by puppet-katello_devel or just old information?

docs/development.md
ekohl: I didn't need it and in my experience the gem generated it with the right hostnames.

docs/development.md
ekohl: This is now handled by puppet-katello_devel.

PR #612, opened 2018-02-15
Fix TOC links

kallies:
Link to "Using Forklift as a Library" is wrong, TOC needs an update.


ekohl:
Manual TOCs are the worst. Thanks for fixing this!

PR #617, opened 2018-02-16
Enhance discovery setup

ares:


roles/plugins/foreman_discovery/tasks/main.yml
```diff
@@ -4,4 +4,16 @@
...
     plugin_installer_options: >
       --enable-foreman-plugin-discovery
+      --enable-foreman-proxy-plugin-discovery
       {{ foreman_discovery_installer_options }}
+
+# TODO until http://projects.theforeman.org/issues/22479 is fixed
+- name: 'check if FDI tar file exists'
+  stat: path=/var/lib/tftpboot/boot/fdi-image-latest.tar
+  register: fdi_image_tar_file
+- name: 'Install latest FDI image'
```
ehelms: Mind an empty line above this to bring it in line with other task files?

ares: I wanted to keep the together as the comment applies to both tasks, should I copy the comment then?

ehelms: Maybe move to a separate file then and include it here? `hacks.yaml` or `workarounds.yaml` ? Will make it clearer to me and allow keeping conventions.

ekohl: I'd prefer to get the bug fixed instead.

ares: I see you already sent a PR, if that gets merged, I'm happy to remove this workaround, though I think it shouldn't block this PR if it's not merge shortly

roles/plugins/foreman_discovery/tasks/main.yml
```diff
@@ -4,4 +4,9 @@
   vars:
     plugin_installer_options: >
       --enable-foreman-plugin-discovery
+      --enable-foreman-proxy-plugin-discovery
       {{ foreman_discovery_installer_options }}
+
+# TODO until http://projects.theforeman.org/issues/22479 is fixed
```
ekohl: This is now fixed in nightlies

ares:
@ekohl removed the workaround, should be ready for merge

ekohl:
Isn't there an option to download images (that defaults to false)?

ares:
yeah sorry, I only deleted the workaround without re-adding the parameter, should be good now

ares:
it's a bit frustrating to see something is approved but not merged :-)

ekohl:
If there are more reviewers on a PR I tend to give them a chance to have another look. @ehelms?

ares:
Sure, though their comments were addressed already :-) Also we have 3 acks at https://github.com/theforeman/forklift/pull/614 and 4th reviewer said LGTM **https://github.com/theforeman/forklift/pull/614#issuecomment-367993619

ekohl:
Thanks!

PR #618, opened 2018-02-16
Fix provisioning setup

ares:
This makes provisioning not only working again but also adds more OSes configuration. With this one is able to provision centos, fedora, debian, ubuntu

### DIFF
--- a/roles/foreman_provisioning/tasks/main.yml
+++ b/roles/foreman_provisioning/tasks/main.yml
@@ -1,4 +1,9 @@
 ---
+# Make sure admin does not default to any taxonomy
+- name: 'disable default context for admin'
+  shell: >
+    {{ foreman_provisioning_hammer }} user update --login admin --default-organization-id 0 --default-location-id 0
+
 # Get the smart proxy ID of the local katello:
 - name: 'get smart proxy id'
   shell: >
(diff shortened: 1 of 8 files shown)


roles/foreman_provisioning/tasks/main.yml
```diff
@@ -145,10 +144,28 @@
   shell: >
     {{ foreman_provisioning_hammer }} settings set --name query_local_nameservers --value true
 
-# testing both --name and --title for backward compatibility
+# make installation media available
+- name: 'assign organization and location to all seeded media'
```
xprazak2: I think this is a usability issue and should be fixed in out seeds.

roles/foreman_provisioning/tasks/main.yml
```diff
@@ -183,44 +200,172 @@
   shell: >
     {{ foreman_provisioning_hammer }} os set-default-template --id {{ foreman_provisioning_centos73.Id }} --config-template-id {{ item.Id }}
   with_items: "{{ kickstart_templates }}"
+# CentOS 7 end
+
+# Fedora 27 start
```
xprazak2: Maybe tasks for each os could be in a separate file and we could just include/import them here?

roles/libvirt_setup/tasks/main.yml
```diff
@@ -36,6 +36,9 @@
     name: provision
     xml: '{{ lookup("template", "../templates/network.xml.j2") }}'
 
+- name: 'make provisioning network autostart on reboot'
+  file: state=link src=/etc/libvirt/qemu/networks/provision.xml path=/etc/libvirt/qemu/networks/autostart/provision.xml
```
xprazak2: Why not `virsh net-autostart provision`?

roles/foreman_provisioning/tasks/main.yml
```diff
@@ -9,12 +14,6 @@
   set_fact:
     foreman_provisioning_smart_proxy: "{{ foreman_provisioning_smart_proxy_json.stdout|from_json }}"
 
-- name: 'install libvirt gem'
```
xprazak2: You need to keep it. Without ruby-libvirt gem, proxy cannot have dns and dhcp feature for libvirt.

roles/foreman_provisioning/tasks/main.yml
```diff
@@ -9,12 +14,6 @@
...
-  yum: name=rubygem-ruby-libvirt state=present
-
-- name: 'install foreman-libvirt'
```
xprazak2: I think you need to keep this one as well, otherwise there will be no libvirt compute resource.

xprazak2:
You will need to rebase, I ran into the domian issue, which should be already fixed in 07caaedb9134360c433608906c8d3cbb6aa4c876

roles/foreman_provisioning/tasks/main.yml
ekohl: I thought running the installer with ```--enable-foreman-compute-libvirt``` would take care of that.

roles/foreman_provisioning/tasks/main.yml
ekohl: ```foreman-libvirt``` will pull this in so it's redundant.

roles/foreman_provisioning/tasks/main.yml
```diff
@@ -183,44 +200,172 @@
...
+- name: 'find Ubuntu 17.10'
+  shell: >
+    {{ foreman_provisioning_hammer }} --output json os info --name "Ubuntu 17.10" ||
+    {{ foreman_provisioning_hammer }} --output json os info --title "Ubuntu 17.10"
+  register: foreman_provisioning_ubuntu93_json
+  ignore_errors: True
+
+- name: 'get CentOS 17.10 info'
+  set_fact:
+    foreman_provisioning_ubuntu93: "{{ foreman_provisioning_ubuntu93_json.stdout|from_json }}"
```
ekohl: ubuntu93 -> ubuntu1710?

roles/foreman_provisioning/tasks/main.yml
```diff
@@ -183,44 +200,172 @@
...
+    --name Ubuntu --major 17 --minor 10 --release-name 'artful' --architectures x86_64 --family 'Debian' --media 'Ubuntu mirror' --partition-tables 'Preseed default'
+
+- name: 'find Ubuntu 17.10'
+  shell: >
+    {{ foreman_provisioning_hammer }} --output json os info --name "Ubuntu 17.10" ||
+    {{ foreman_provisioning_hammer }} --output json os info --title "Ubuntu 17.10"
+  register: foreman_provisioning_ubuntu93_json
+  ignore_errors: True
+
+- name: 'get CentOS 17.10 info'
```
ekohl: centos 17.10 -> ubuntu 17.10?

roles/libvirt_setup/tasks/main.yml
ekohl: You could also add ```autostart: yes``` to the above ```virt_net``` definition for the same effect.

roles/foreman_provisioning/tasks/main.yml
ares: I tend to agree, I opened a ticket and will add a comment to the code, this shouldn't prevent having this workaround meanwhile...

roles/libvirt_setup/tasks/main.yml
ares: sadly either didn't work, I think it's a bug in libvirt, I see the network is set to autoprovision but the symlink is missing and it does not start after HW is rebooted, I'll add extra comment

ares: also from what I read, `autostart: yes` is default

xprazak2:
I retested, failing with:

```
/Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[hostname]/features: change from ["Logs", "Pulp", "Puppet", "Puppet CA", "TFTP", "Templates"] to ["DHCP", "DNS", "Logs", "Puppet", "Puppet CA", "TFTP", "Templates"] failed:
Proxy hostname has failed to load one or more features (DHCP, DNS), check /var/log/foreman-proxy/proxy.log for configuration errors
```

iNecas:
@xprazak2 this error you're seeing is caused due to missing `rubygem-ruby-libvirt` package. after installing it, it moved forward.

After fixing that, I've hit also issue with `unknown provider`, so I needed to install `foreman-libvirt` manually in order to proceed

roles/foreman_provisioning/tasks/main.yml
ekohl: Correction, it is needed since foreman-libvirt pulls in the SCL version where this is the non-SCL version for the proxy.

roles/katello_provisioning/tasks/main.yml
```diff
@@ -102,6 +102,35 @@
...
+    --name "EPEL 7"
+    --url "http://download.fedoraproject.org/pub/epel/7/x86_64"
+    --download-policy {{ katello_provisioning_download_policy }}
+    --organization "{{ katello_provisioning_organization }}"
+  when: katello_epel_repo.stderr.find('not found') != -1
+
+- name: 'sync the epel repo'
+  shell: >
+    {{ katello_provisioning_hammer }} repository synchronize
+    --name "EPEL 7"
```
iNecas: This just ate my hard-drive: do we really need to download 12GB of data by default? I would perhaps leave it as an optional thing to do (something like `katello_provisioning_epel_sync`).

roles/foreman_provisioning/tasks/main.yml
```diff
@@ -145,82 +144,78 @@
...
+# --environment production
+# TODO compute-profile can't be specified by name until http://projects.theforeman.org/issues/21580/ so we hardcode 1
+- name: 'create hostgroup Base'
   shell: >
     {{ foreman_provisioning_hammer }} hostgroup create
-    --name 'Forklift Base'
+    --name 'Base'
     --architecture x86_64
     --domain {{ foreman_provisioning_domain }}
-    --environment production
```
iNecas: Why removing this? It means I can't use the host group directly and need to set the puppet environment manually later.

roles/foreman_provisioning/tasks/main.yml
```diff
@@ -145,82 +144,78 @@
...
+    {{ foreman_provisioning_hammer }} hostgroup info --name 'CentOS 7 Mirror'
   register: foreman_provisioning_hostgroup_centos_mirror
   ignore_errors: True
 
-- name: 'create hostgroup Forklift CentOS 7 Mirror'
+- name: 'create hostgroup CentOS 7 Mirror'
   shell: >
     {{ foreman_provisioning_hammer }} hostgroup create
-    --name 'Forklift CentOS 7 Mirror'
+    --name 'CentOS 7 Mirror'
```
iNecas: I guess we should add `--pxe-loader "PXELinux BIOS"` here, as otherwise, the provisioning doesn't work out of the box

ares: added to the base rather, it gets inherited to all children

roles/katello_provisioning/tasks/main.yml
ares: how did you test it? default policy is on_demand so unless you do it with immediate it should be fine (that is default for demo hw, perhaps you tested it with that?)

roles/foreman_provisioning/tasks/main.yml
ares: see the TODO above, can't set environment, it seems to be fixed now so removing it

ares:
Update all and rebased, I was able to deploy the whole setup easily with this version. I suggest we won't add more stuff and continue improving it via separate PRs (of course this version also works for you, seem just fine in my env)

roles/foreman_provisioning/tasks/main.yml
iNecas: :+1: interesting it was not possible to set on UI, when OS was not selected, therefore I was not suggesting that. Anyway, tested by updating the loader just on the base and it was properly propagating.

roles/katello_provisioning/tasks/main.yml
iNecas: You are right, I haven't realized this: I actually panicked, because I was low on disk space, and even the metadata parsing probably caused some spikes. So I guess we are ok now :)

iNecas:
:+1: let's fix any additional things separately. This is good enough PR to be merged. All the code introduced here works fine and enhances the provisioning setup quite a bit.

ekohl:
Thanks!

PR #616, opened 2018-02-16
Enhance openscap plugin setup

ares:

PR #619, opened 2018-02-16
Demo hw

ares:
requires:
- ~~https://github.com/theforeman/forklift/pull/615~~
- https://github.com/theforeman/forklift/pull/616 (ready for testing, not blocking this PR)
- ~~https://github.com/theforeman/forklift/pull/617~~
- ~~https://github.com/theforeman/forklift/pull/618~~

There are some TODOs when more stuff is available through API or fixed but that shouldn't block the merge as it can be already used today

### DIFF
--- a/playbooks/luna_demo_environment.yml
+++ b/playbooks/luna_demo_environment.yml
@@ -1,0 +1,79 @@
+# Manually verify your hosts has FQDN set correctly
+# check it's in /etc/hosts /etc/hostname and hostname returns it, /etc/hosts first line should be 192.168.73.1  $fqdn
+
+# TODO - add role for setting up the machine - autologin in gdm, perhaps background image etc, installation of virt-manager, vim, htop
+# TODO - refactor katello_provisioning so that it does not run installer, it should only seed stuff, installer should be customized from playbook using foreman_installer's role variables
+
+- hosts: all
+  become: true
+  roles:
+    - disable_ipv6
+    - disable_firewall
+    - selinux
+    - etc_hosts
+    - epel_repositories
+    - ostree_repositories
+    - puppet_repositories
+    - foreman_repositories
+    - katello_repositories
+    - foreman_installer
+    - libvirt_setup
+    - katello_provisioning
+    - plugins/foreman_templates
+# openscap manual steps
+# 1. edit Base/Katello CentOS 7/OpenSCAP host group and set the policy manually
+# covered by http://projects.theforeman.org/issues/22481
+    - plugins/foreman_openscap
+# discovery manual steps (to make discovery work via pxelinux):
+# 1. unlock PXELinux global default
+# 2. change ONTIMEOUT local to ONTIMEOUT discovery
+# 3. save
+# 4. click Build PXE Default button
+# covered by http://projects.theforeman.org/issues/21204/
+    - plugins/foreman_discovery
+    - plugins/foreman_bootdisk
+    - plugins/foreman_remote_execution
+    - plugins/foreman_ansible 
+  vars:
+    katello_provisioning_organization: Demo
+    katello_provisioning_location: Belgium
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+    foreman_installer_additional_packages:
+      - rubygem-ruby-libvirt
+      - foreman-libvirt
+      - foreman-ec2
+      - foreman-gce
+      - foreman-openstack
+      - foreman-ovirt
+      - foreman-rackspace
+      - foreman-vmware
+      - katello
+    foreman_installer_options: >
+      --foreman-admin-password {{ foreman_installer_admin_password }}
+      --foreman-organizations-enabled true 
+      --foreman-locations-enabled true 
+      --foreman-initial-organization {{ katello_provisioning_organization }}
+      --foreman-initial-location {{ katello_provisioning_location }} 
+      --puppet-show-diff=true
+      --foreman-proxy-dns true 
+      --foreman-proxy-dhcp true 
+      --foreman-proxy-dns-managed false 
+      --foreman-proxy-dhcp-managed false 
+      --foreman-proxy-dns-provider libvirt 
+      --foreman-proxy-dns-interface virbr1 
+      --foreman-proxy-dhcp-provider libvirt 
+      --foreman-proxy-libvirt-network provision 
+      --foreman-proxy-dhcp-interface virbr1 \
+    foreman_repositories_environment: release
+    katello_repositories_environment: release
+    selinux_state: disabled
+    # if you're rerunning this playbook to save some time you may want to disable repo sync if they were synced recently
+    katello_provisioning_sync_repos: True
+    # if you want to speed up the setup and you don't care about packages being downloaded, you can change this to on_demand
+    # valid options are on_demand, background, immediate
+    katello_provisioning_download_policy: immediate
+# MANUAL STEPS
+# 1. create a compute profile through UI (there's no API for it)
+#   define it as Small, 2 GB RAM, network type is Virtual (NAT) with network provision


ares:
updated, ready for review/testing, @adamruzicka might be handy for you

playbooks/demo_hw.yml
```diff
@@ -0,0 +1,80 @@
...
+      - foreman-openstack
+      - foreman-ovirt
+      - foreman-rackspace
+      - foreman-vmware
+      - katello
+    foreman_installer_options: >
+      --foreman-organizations-enabled true 
+      --foreman-locations-enabled true 
+      --foreman-initial-organization Demo 
+      --foreman-initial-location Belgium
```
adamruzicka: Maybe reuse `katello_provisioning_{location,organization}` here?

ares: reused :-)

ares:
@adamruzicka does it work for you now? Is there something else blocking this PR? I hope other improvements can be addressed separately.

ares:
@ehelms could this get in?

ehelms:
@ares yes - with one question from me. "demo hw" is not something that is obvious for what this is doing or a user looking at this would know what to do with this. Can we give this a more descriptive name?

ares:
how about foreman_demo_environment? foreman_presentation? luna_demo_environment?

playbooks/demo_hw.yml
```diff
@@ -0,0 +1,83 @@
+# Manually verify your hosts has FQDN set correctly
+# check it's in /etc/hosts /etc/hostname and hostname returns it, /etc/hosts first line should be 192.168.73.1  $fqdn
```
ekohl: Our etc_hosts role do this. If not, you should ensure the ansible inventory has the correct IP.

playbooks/demo_hw.yml
```diff
@@ -0,0 +1,83 @@
...
+
+# TODO - setup puppetserver so it show_diffs=true
```
ekohl: That should be `foreman-installer --puppet-show-diff=true`

playbooks/demo_hw.yml
```diff
@@ -0,0 +1,83 @@
...
+# the bug with ssh keys not being deployed
+# http://projects.theforeman.org/issues/22483/
+    - plugins/foreman_remote_execution
+    - plugins/foreman_ansible 
+  vars:
+    katello_provisioning_organization: Demo
+    katello_provisioning_location: Belgium
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "{{ '--disable-system-checks' if katello_repositories_version != '3.1' else '' }}"
```
ekohl: This can be simplified to just `--disable-system-checks` since I doubt we ever want to demo 3.1.

playbooks/demo_hw.yml
```diff
@@ -0,0 +1,83 @@
...
+    foreman_repositories_environment: release
+    katello_repositories_environment: release
+    selinux_state: disabled
+    # if you're rerunning this playbook to save some time you may want to disable repo sync if they were synced recently
+    katello_provisioning_sync_repos: True
+    # if you want to speed up the setup and you don't care about packages being downloaded, you can change this to on_demand
+    # valid options are on_demand, background, immediate
+    katello_provisioning_download_policy: immediate
+    # TODO can't install images this way, because of http://projects.theforeman.org/issues/22479
+    # foreman_discovery_installer_options: "--foreman-proxy-plugin-discovery-install-images true"
```
ekohl: This was fixed and should be in nightlies.

ehelms:
If it is the equivalent set of plugins to luna, I like that with the demo_environment

playbooks/demo_hw.yml
ares: I'm not sure if this is the way how I'm running it, but it didn't work for systems that couldn't resolve their FQDN. I ran ansible-playbook with `-i machine.example.com,` where machine.example.com was on my localhost /etc/hosts. I think this comment does not do any harm and checking hostname is properly is good idea anyway.

playbooks/luna_demo_environment.yml
```diff
@@ -0,0 +1,80 @@
...
+# openscap manual steps
+# 1. edit Base/Katello CentOS 7/OpenSCAP host group and set the policy manually
+# covered by http://projects.theforeman.org/issues/22481
+    - plugins/foreman_openscap
+# discovery manual steps (to make discovery work via pxelinux):
+# 1. unlock PXELinux global default
+# 2. change ONTIMEOUT local to ONTIMEOUT discovery
+# 3. save
+# 4. click Build PXE Default button
+# covered by http://projects.theforeman.org/issues/21204/
```
ekohl: This one is closed so is it still an issue?

playbooks/luna_demo_environment.yml
```diff
@@ -0,0 +1,80 @@
...
+      --foreman-proxy-dhcp-interface virbr1 \
+    foreman_repositories_environment: release
+    katello_repositories_environment: release
+    selinux_state: disabled
+    # if you're rerunning this playbook to save some time you may want to disable repo sync if they were synced recently
+    katello_provisioning_sync_repos: True
+    # if you want to speed up the setup and you don't care about packages being downloaded, you can change this to on_demand
+    # valid options are on_demand, background, immediate
+    katello_provisioning_download_policy: immediate
+    foreman_discovery_installer_options: "--foreman-proxy-plugin-discovery-install-images true"
```
ekohl: This is already in the role since 62581ed8214a19db6fbba54907d93dee2b8b65e1 so can be removed

playbooks/luna_demo_environment.yml
```diff
@@ -0,0 +1,80 @@
...
+# covered by http://projects.theforeman.org/issues/21204/
+    - plugins/foreman_discovery
+    - plugins/foreman_bootdisk
+    - plugins/foreman_remote_execution
+    - plugins/foreman_ansible 
+  vars:
+    katello_provisioning_organization: Demo
+    katello_provisioning_location: Belgium
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
```
ekohl: Can we move these to `foreman_installer_options`? The admin password is already set as a default.

playbooks/luna_demo_environment.yml
```diff
@@ -0,0 +1,80 @@
...
+      - rubygem-ruby-libvirt
+      - foreman-libvirt
+      - foreman-ec2
+      - foreman-gce
+      - foreman-openstack
+      - foreman-ovirt
+      - foreman-rackspace
+      - foreman-vmware
+      - katello
+    foreman_installer_options: >
```
ekohl: Does this work to specify a list?

playbooks/luna_demo_environment.yml
```diff
@@ -0,0 +1,80 @@
...
+    katello_provisioning_organization: Demo
+    katello_provisioning_location: Belgium
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--foreman-admin-password {{ foreman_installer_admin_password }}"
+      - "--puppet-show-diff=true"
+    foreman_installer_additional_packages:
+      - rubygem-ruby-libvirt
+      - foreman-libvirt
```
ekohl: These can also be set using `--foreman-compute-PROVIDER` as installer options.

playbooks/luna_demo_environment.yml
ares: as I mentioned above, the discovery setup could probably be improved already, I don't have cycles to fiddle with settings through hammer now

playbooks/luna_demo_environment.yml
ares: I think I tried these already several times but for some reason, this package must have been specified explicitly, if you insist I can test the playbook again with this option

ares:
pushed few more changes that were easy enough

playbooks/luna_demo_environment.yml
ekohl: Then the docs are out of date: https://github.com/theforeman/forklift/blob/7099a4e99b8fa5e3107f5bb471d9eb4ac180c3ab/roles/foreman_installer/tasks/install.yml#L2-L5

playbooks/luna_demo_environment.yml
```diff
@@ -0,0 +1,79 @@
...
+      --foreman-proxy-dns-managed false 
+      --foreman-proxy-dhcp-managed false 
+      --foreman-proxy-dns-provider libvirt 
+      --foreman-proxy-dns-interface virbr1 
+      --foreman-proxy-dhcp-provider libvirt 
+      --foreman-proxy-libvirt-network provision 
+      --foreman-proxy-dhcp-interface virbr1 \
+    foreman_repositories_environment: release
+    katello_repositories_environment: release
+    selinux_state: disabled
```
akofink: Is this required? What isn't compatible with selinux?

ares: This may be from the time it was broken, I don't recall tbh

ehelms:
@ares What do you want to do with this PR?

ekohl:
We merged #996 which was based on this.

ares:
Thank you

PR #623, opened 2018-02-19
Add devel box that configures multiple plugins

ehelms:
This does not currently work due to (https://community.theforeman.org/t/rename-foreman-docker-git-repository-to-foreman-docker/8391). That will either have to get resolved, or we will have to add special code in puppet-katello_devel that handles the special case of foreman_docker.

You might ask yourself why is this named "charon". Well, it sounded better than "foreman_with_a_bunch_of_plugins" and was short and to the point. Charon is a moon or natural satellite of the dwarf Planet pluto.

### DIFF
--- a/playbooks/luna-devel.yml
+++ b/playbooks/luna-devel.yml
@@ -1,0 +1,17 @@
+---
+- import_playbook: devel.yml
+  vars:
+    katello_devel_github_username: ''
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--katello-devel-github-username '{{ katello_devel_github_username }}'"
+      - "--katello-devel-enable-ostree=true"
+      - "--katello-devel-extra-plugins theforeman/foreman_remote_execution"
+      - "--katello-devel-extra-plugins theforeman/foreman_discovery"
+      - "--katello-devel-extra-plugins theforeman/foreman_ansible"
+      - "--katello-devel-extra-plugins theforeman/foreman-tasks"
+      - "--katello-devel-extra-plugins theforeman/foreman_bootdisk"
+      - "--katello-devel-extra-plugins theforeman/foreman_openscap"
+      - "--katello-devel-extra-plugins theforeman/foreman_templates"
+      - "--katello-devel-extra-plugins theforeman/foreman_hooks"
+      - "--katello-devel-extra-plugins theforeman/foreman_docker"
--- a/playbooks/luna.yml
+++ b/playbooks/luna.yml
@@ -1,0 +1,20 @@
+---
+- hosts: all
+  become: true
+  vars:
+    foreman_installer_options:
+      - "--disable-system-checks"
+      - "--foreman-admin-password {{ foreman_installer_admin_password }}"
+      - "--enable-foreman-plugin-ansible"
+      - "--enable-foreman-plugin-discovery"
+      - "--enable-foreman-plugin-bootdisk"
+      - "--enable-foreman-plugin-hooks"
+      - "--enable-foreman-plugin-openscap"
+      - "--enable-foreman-plugin-templates"
+      - "--enable-foreman-plugin-remote-execution"
+      - "--enable-foreman-proxy-plugin-remote-execution-ssh"
+      - "--enable-foreman-proxy-plugin-discovery"
+      - "--enable-foreman-proxy-plugin-ansible"
+      - "--enable-foreman-proxy-plugin-openscap"
+  roles:
+    - role: katello
--- a/roles/foreman_installer/tasks/install.yml
+++ b/roles/foreman_installer/tasks/install.yml
@@ -3,7 +3,8 @@
   shell: >
       {{ foreman_installer_command }} {{ (foreman_installer_verbose == True) | ternary("-v", "") }}
       --scenario "{{ foreman_installer_scenario }}"
-      {{ foreman_installer_options }}
+      {{ foreman_installer_options | join(" ") if foreman_installer_options is iterable else '' }}
+      {{ foreman_installer_options if foreman_installer_options is string else '' }}
       {{ foreman_installer_options_internal_use_only | join(" ") }}
   when: not foreman_installer_skip_installer
   tags:
--- a/vagrant/boxes.d/01-builtin.yaml
+++ b/vagrant/boxes.d/01-builtin.yaml
@@ -27,3 +27,8 @@ boxes:
     box: centos7
     ansible:
       playbook: 'playbooks/runcible.yml'
+
+  centos7-luna:
+    box: centos7
+    ansible:
+      playbook: 'playbooks/luna.yml'
--- a/vagrant/boxes.d/99-local.yaml.example
+++ b/vagrant/boxes.d/99-local.yaml.example
@@ -10,6 +10,15 @@ centos7-devel:
       foreman_devel_github_push_ssh: True
       katello_devel_github_username: <REPLACE ME>
 
+centos7-luna-devel:
+  box: centos7
+  memory: 9000
+  ansible:
+    playbook: 'playbooks/luna-devel.yml'
+    group: 'devel'
+    variables:
+      katello_devel_github_username: <REPLACE ME>
+
 centos7-hammer-devel:
   box: centos7
   memory: 512


boxes.d/99-local.yaml.example
```diff
@@ -10,6 +10,25 @@ centos7-devel:
...
       katello_devel_github_username: <REPLACE ME>
 
+charon:
+  box: centos7
+  memory: 9000
+  ansible:
+    playbook: 'playbooks/devel.yml'
+    group: 'devel'
+    variables:
+      katello_devel_github_username: <REPLACE ME>
```
ekohl: I did consider using some system variable or ansible variable to automatically do this, but couldn't really find a good way. Closes was delegating one to localhost, register the ```ansible_username``` fact and defaulting to that later on.

roles/foreman_installer/tasks/install.yml
```diff
@@ -3,7 +3,7 @@
   shell: >
       {{ foreman_installer_command }} {{ (foreman_installer_verbose == True) | ternary("-v", "") }}
       --scenario "{{ foreman_installer_scenario }}"
-      {{ foreman_installer_options }}
+      {{ foreman_installer_options | join(" ") }}
```
ekohl: Doesn't this break when it's a string?

playbooks/charon-devel.yml
```diff
@@ -0,0 +1,30 @@
...
+  vars:
+    katello_repositories_version: nightly
+    foreman_installer_scenario: katello-devel
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--katello-devel-github-username {{ katello_devel_github_username }}"
+      - "--katello-devel-enable-ostree=true"
+    foreman_installer_additional_packages:
+      - foreman-installer-katello-devel
+    foreman_installer_options:
```
ekohl: Should this also be on internal so users can override other settings easily in boxes.d?

ehelms: Fair point.

playbooks/charon-devel.yml
```diff
@@ -0,0 +1,30 @@
...
+      - "--katello-devel-extra-plugins theforeman/foreman_remote_execution"
+      - "--katello-devel-extra-plugins theforeman/foreman_discovery"
+      - "--katello-devel-extra-plugins theforeman/foreman_ansible"
+      - "--katello-devel-extra-plugins theforeman/foreman-tasks"
+      - "--katello-devel-extra-plugins theforeman/foreman_bootdisk"
+      - "--katello-devel-extra-plugins theforeman/foreman_openscap"
+      - "--katello-devel-extra-plugins theforeman/foreman_templates"
+      - "--katello-devel-extra-plugins theforeman/foreman_hooks"
+      - "--katello-devel-extra-plugins theforeman/foreman-docker"
```
ekohl: Should be foreman_docker now

ehelms:
This currently breaks without https://github.com/theforeman/puppet-katello_devel/pull/145

ares:
Could we reconsider the name please :-) if you want to keep space nomenclature, how about sond or probe? since this is based on non-released version, it's sond/probe in what's will be found in next release

anyway I think this is the answer for what I wanted to do in devel playbook, I'm happy to test once other @ekohl comments are addressed. I'd also like to add hammer and smart proxy devel at some point, ofc course not in this PR :-) A single playbook for complete devel env.

playbooks/charon-devel.yml
```diff
@@ -0,0 +1,30 @@
...
+      - "--disable-system-checks"
+      - "--katello-devel-github-username {{ katello_devel_github_username }}"
+      - "--katello-devel-enable-ostree=true"
+    foreman_installer_additional_packages:
+      - foreman-installer-katello-devel
+    foreman_installer_options:
+      - "--katello-devel-extra-plugins theforeman/foreman_remote_execution"
+      - "--katello-devel-extra-plugins theforeman/foreman_discovery"
+      - "--katello-devel-extra-plugins theforeman/foreman_ansible"
+      - "--katello-devel-extra-plugins theforeman/foreman-tasks"
```
rplevka: i think there is the same problem as with "foreman-docker".
using the current version, foreman installer is failing with:

```
[!] There was an error parsing `Gemfile`: Undefined local variable or method `tasks_dev' for Gemfile. Bundler cannot continue.
```
foreman-tasks.local.rb:
```
:development_group => :foreman-tasks_dev
```
enclosing the symbol in doublequotes works fine.

ehelms: https://github.com/theforeman/puppet-katello_devel/pull/145 fixes that but it has to propagate to the nightly repositories which is currently blocked by an issue being fixed in Foreman.

ehelms:
Given some input that 'charon' doesn't translate phonetically well into other languages I switched to use a different moon, Luna, which is the latin name for the Earth's moon. Happy to bikeshed this further if folks would prefer, but perhaps Luna rolls off the tongue well (plus its short and sweet).

playbooks/luna-devel.yml
```diff
@@ -0,0 +1,29 @@
+- hosts: devel
+  become: true
+  vars:
+    katello_repositories_version: nightly
+    foreman_installer_scenario: katello-devel
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--katello-devel-github-username {{ katello_devel_github_username }}"
+      - "--katello-devel-enable-ostree=true"
+      - "--katello-devel-extra-plugins theforeman/foreman_remote_execution"
```
dLobatog: Does this set up the proxies required for these features?

ares: no, I think it would be useful to add installer options so that proxy is installed with all respective plugins via rpms until we have proper dev setup of proxy with plugins too

ekohl:
@BlackSmith an easier solution is to use https://github.com/theforeman/forklift/blob/master/roles/foreman_devel/tasks/github_push_ssh.yml which does something similar but for every git repository. It's here included via the ```foreman_devel``` role

playbooks/luna-devel.yml
BlackSmith: The plugin / repository was renamed from "theforeman/foreman-docker" to "theforeman/foreman_docker".
https://github.com/theforeman/foreman_docker

BlackSmith: Sorry this is comment to line 18.

boxes.d/99-local.yaml.example
```diff
@@ -10,6 +10,15 @@ centos7-devel:
       foreman_devel_github_push_ssh: True
       katello_devel_github_username: <REPLACE ME>
 
+luna-devel:
```
akofink: Shouldn't this be 'centos7-luna-devel' to match the others?

ehelms:
[test containers]

ehelms:
[test containers]

ehelms:
[test containers]

ehelms:
OK - this is fully updated to take into account existing `devel.yml` and the updates to devel scenario handling.

ehelms:
@ares @ekohl Any others care to re-review?

playbooks/luna-devel.yml
```diff
@@ -0,0 +1,17 @@
...
+      - "--katello-devel-extra-plugins theforeman/foreman_discovery"
+      - "--katello-devel-extra-plugins theforeman/foreman_ansible"
+      - "--katello-devel-extra-plugins theforeman/foreman-tasks"
+      - "--katello-devel-extra-plugins theforeman/foreman_bootdisk"
+      - "--katello-devel-extra-plugins theforeman/foreman_openscap"
+      - "--katello-devel-extra-plugins theforeman/foreman_templates"
+      - "--katello-devel-extra-plugins theforeman/foreman_hooks"
+      - "--katello-devel-extra-plugins theforeman/foreman-docker"
```
ekohl: This is now ```foreman_docker```, right (as suggested above by @BlackSmith)?

ehelms: Updated

playbooks/luna.yml
```diff
@@ -0,0 +1,19 @@
...
+      - "--disable-system-checks"
+      - "--foreman-admin-password {{ foreman_installer_admin_password }}"
+      - "--enable-foreman-plugin-ansible"
+      - "--enable-foreman-plugin-discovery"
+      - "--enable-foreman-plugin-foreman-bootdisk"
+      - "--enable-foreman-plugin-foreman-hooks"
+      - "--enable-foreman-plugin-foreman-openscap"
+      - "--enable-foreman-plugin-foreman-templates"
+      - "--enable-foreman-plugin-remote-execution"
+      - "--enable-foreman-proxy-plugin-remote-execution-ssh"
```
ekohl: Would it make sense to include these proxy plugins in the devel scenario as well?

akofink:
Got the following error:

```
[ERROR 2018-03-22T14:28:57 verbose]  /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[luna-devel.akofink-desktop]: Could not evaluate: Proxy luna-devel.akofink-desktop cannot be retrieved: unknown error (
response 503)
```

Maybe I'm missing something? Should I still need https://github.com/theforeman/puppet-katello_devel/pull/145 in foreman_installer_module_prs?

@ekohl if you have it working, would you mind sharing the box yaml you're using?

akofink:
So, after a rebase, here's what I'm getting. It's definitely weird:

```
Execution of 'git remote add --katello-devel-enable-ostree=true https://--katello-devel-enable-ostree=true@github.com/--katello-devel-enable-ostree=true/foreman.git' returned
 129: error: unknown option `katello-devel-enable-ostree=true'
```

playbooks/luna.yml
```diff
@@ -0,0 +1,20 @@
+---
+- hosts: all
+  become: true
+  vars:
+    foreman_installer_options:
```
ekohl: This should be ```foreman_installer_options_internal_use_only```

ehelms: The Katello meta role does not allow me to override that. Hence why I opted for this given this is specifically attempting to set certain installer variables for it's purpose. The other option is I just don't use the Katello role itself and re-implement it here :/

ehelms:
@akofink I think https://github.com/theforeman/puppet-katello_devel/pull/153 will solve the weirdness

akofink:
With https://github.com/theforeman/puppet-katello_devel/pull/153: Now it seems the proxy doesn't come up.

```
[ERROR 2018-03-27T18:39:59 verbose]  /Stage[main]/Foreman_proxy::Register/Foreman_smartproxy[luna-devel.akofink-desktop]: Could not evaluate: Proxy luna-devel.akofink-desktop cannot be retrieved: unknown error (response 503)
[ERROR 2018-03-27T18:39:59 verbose] /usr/share/foreman-installer/modules/foreman/lib/puppet/provider/foreman_smartproxy/rest_v3.rb:7:in `proxy'
[ERROR 2018-03-27T18:39:59 verbose] /usr/share/foreman-installer/modules/foreman/lib/puppet/provider/foreman_smartproxy/rest_v3.rb:13:in `id'
[ERROR 2018-03-27T18:39:59 verbose] /usr/share/foreman-installer/modules/foreman/lib/puppet/provider/foreman_smartproxy/rest_v3.rb:17:in `exists?'
...
```

Even if this is unrelated to the changes in this PR, it's preventing me from fully testing. @ehelms does your box come up without errors right now? If so, what does your boxes.yaml look like? Koji yes/no?

ekohl:
@akofink there is a race condition in the config. https://github.com/theforeman/forklift/commit/3bd25a9315995b832f5b684e4bed02356703453e isn't in katello-installer. If you run the devel scenario, then the installer (whether it installs it or updates it) will overwrite it again. I think that if you rerun the devel setup it should work.

https://github.com/Katello/katello-installer/commit/dd307b0b7c9bc541d801785f78e1024e87b75947 & https://github.com/theforeman/foreman-packaging/pull/2281 need to make it in.

akofink:
So far, rerunning produces the same error. I'm a bit stretched for time at the moment, but I'm willing to retest when the installer change makes it in.

playbooks/luna.yml
```diff
@@ -0,0 +1,20 @@
...
+      - "--enable-foreman-plugin-discovery"
+      - "--enable-foreman-plugin-bootdisk"
+      - "--enable-foreman-plugin-hooks"
+      - "--enable-foreman-plugin-openscap"
+      - "--enable-foreman-plugin-templates"
+      - "--enable-foreman-plugin-remote-execution"
+      - "--enable-foreman-proxy-plugin-remote-execution-ssh"
+      - "--enable-foreman-proxy-plugin-discovery"
+      - "--enable-foreman-proxy-plugin-ansible"
+      - "--enable-foreman-proxy-plugin-openscap"
```
ares: I may be missing something, but doesn't it need extra records in https://github.com/theforeman/forklift/blob/master/roles/foreman_installer_devel_scenario/files/katello-devel-answers.yaml#L32 ?

ehelms: I think you are right. Surprised the installer does not complain.

ehelms: Oh wait, no, this uses the standard Katello scenario. Note this is not the same as luna-devel, this does a nightly production install with all of these enabled.

roles/foreman_installer/tasks/install.yml
akofink: Yes it does. It separates each character in the string with a `" "`

akofink: I'd vote to put this in a separate PR (I like the change to using an array of stings vs. one big ole string). We'd need to change any places in forklift that are using the string of options.

playbooks/luna.yml
akofink: Could we get a nightly production luna box added to the `boxes.d/99-local.yaml.example`, or (bonus points) enable luna box configuration with `vagrant/config/versions.yaml`.

roles/foreman_installer/tasks/install.yml
ehelms: :/  From grepping, this new playbook is literally the only thing using this option and if possible, I'd prefer not to write the luna options as a giant ugly string only to then to then re-write them back to an array.

ehelms:
@akofink I added a luna box by default.  @ekohl You have changes requested, can you re-review?

roles/foreman_installer/tasks/install.yml
ekohl: That's because this option was always intended as a *user* option. You'll find this in `boxes.d/*` which is not in git. You're right that an array is better, but just changing it will break users configs.

playbooks/luna.yml
ekohl: In the `luna-devel` playbook you import another playbook. Would that work here as well?

akofink:
I tested the centos7-luna box, and I'm getting [this error](https://paste.fedoraproject.org/paste/paZ9vVFpHJRP4ghrXAd-dQ), even when I try to reprovision:

```
[ WARN 2018-04-10T19:54:07 verbose]  Compiled catalog for centos7-luna.akofink-desktop in environment production in 5.06 seconds  
[ INFO 2018-04-10T19:54:10 verbose]  Applying configuration version '1523390042'                                            
[ WARN 2018-04-10T19:54:42 verbose]  /Stage[main]/Foreman_proxy::Plugin::Remote_execution::Ssh/File[/usr/share/foreman-proxy/.ssh]/ensure: ensure changed 'link' to 'directory'                                                               
[ WARN 2018-04-10T19:54:42 verbose]  /Stage[main]/Foreman_proxy::Plugin::Remote_execution::Ssh/Exec[generate_ssh_key]/returns: executed successfully
[ERROR 2018-04-10T19:54:44 verbose]  Execution of '/bin/yum -d 0 -e 0 -y install rubygem-smart_proxy_ansible' returned 1: mv: '/usr/share/foreman-proxy/.ansible.cfg' and '/etc/foreman-proxy/ansible.cfg' are the same file
[ERROR 2018-04-10T19:54:44 verbose]  %pre(rubygem-smart_proxy_ansible-2.0.2-2.fm1_18.el7.noarch) scriptlet failed, exit status 1 
[ERROR 2018-04-10T19:54:44 verbose] Error in PREIN scriptlet in rpm package rubygem-smart_proxy_ansible-2.0.2-2.fm1_18.el7.noarch
```

ehelms:
@ekohl Any further comments or is this ready to be merged and iterated on?

roles/foreman_installer/tasks/install.yml
ehelms: I will make this change, but not because I agree with it. This repository is unversioned, this is relatively unused parameter and this just adds technical debt cruft rather than moving towards a better solution for users that will be easy to adapt to. So before I make this change, the question is when can we remove this backwards compatibility layer? 1 month? 2 months?

akofink: I'm fine with making the change and breaking current box configs if you make sure to notify the forklift users (i.e. community post?).

Didn't you face the same dilemma when you moved the vagrant directory? Isn't [this](https://github.com/theforeman/forklift/blob/master/Vagrantfile#L9) tech debt? And when should we remove it? ;)

ekohl: @ehelms I'd also be fine with a hard error instructing users to move to a list rather than a string. Forcing users to upgrade generally makes the time to remove it shorter.

akofink: I like that idea, @ekohl. Maybe we should also make an issue on forklift to list the tech debt we need to eventually remove.

ehelms:
Added the backwards compatible change.

ekohl:
Not sure if I'm blind, but I don't see the backwards compatible changes.

akofink:
yeah weird. @ehelms did you revert that change?

ehelms:
@ekohl @akofink I did! Cause the code snippet did not work. 

```
{{ foreman_installer_options | join(" ")  if list(foreman_installer_options) }}
{{ foreman_installer_options if string(foreman_installer_options) }}
```

Threw errors about `list` so I reverted since that was easier than trying to investigate. I originally assumed this code had already been vetted.

ekohl:
@ehelms does (pseudo-code)`fail: when: string(foreman_install_options)` work?

akofink:
> I originally assumed this code had already been vetted.

@ehelms I did say "something like this" :man_shrugging:

ehelms:
@ekohl @akofink This should now achieve what you want

PR #624, opened 2018-02-20
katello_client: cleanup, AK support, extra repos support

evgeni:


roles/katello_client/tasks/main.yml
```diff
@@ -1,16 +1,35 @@
...
+    state: "present"
+    org_id: "{{ katello_client_organization }}"
+    activationkey: "{{ katello_client_activationkey }}"
+  when:
+    - katello_client_activationkey != ""
+
+- name: 'Enable extra repositories'
+  command: subscription-manager repos --enable {{ item }}
+  with_items:
+    "{{ katello_client_extra_repos }}"
```
ehelms: This line is a bit of a hanging chad here. Did you intend to have a `-` or shoudl it be on the line above it?

ehelms:
I think this change looks great. The small comment and one follow up question.

This is still the old style of installing the full katello-agent. Should this install katello-host-tools and optionally install katello-agent based on a flag?

roles/katello_client/tasks/main.yml
evgeni: either works, I think. on the line above is probably more readable.

evgeni:
@ehelms updated

roles/katello_client/tasks/main.yml
```diff
@@ -1,18 +1,42 @@
...
+  when:
+    - katello_client_activationkey != ""
+
+- name: 'Enable extra repositories'
+  command: subscription-manager repos --enable {{ item }}
+  with_items: "{{ katello_client_extra_repos }}"
+
+- name: 'Install katello-host-tools-fact-plugin'
+  yum:
+    name: "katello-host-tools-fact-plugin"
```
ehelms: I am curious the opt here for katello-host-tools-fact-plugin instead of the base package of katello-host-tools (which requires the fact plugin) ?

Do you think adding an additional option for katello-host-tools-tracer would be useful?

evgeni: that's just me being silly, k-h-t is of course the right package here.

will add a tracer option too

evgeni:
and updated again!

PR #627, opened 2018-02-24
Refactor Ruby scl role to be generic

ehelms:


roles/ruby_scl/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+    - "{{ ruby_scl_version }}"
+    - "{{ ruby_scl_version }}-ruby-devel"
+
+- name: 'Enable SCL at login'
+  blockinfile:
+    dest: /etc/profile.d/enable-{{ ruby_scl_version }}.sh
+    create: yes
+    block: |
+      #!/bin/bash
+      source /opt/rh/{{ ruby_scl_version }}/enable
```
ekohl: The modern way is ```source scl_source enable {{ ruby-scl_version }}``` which also sets ```X_SCLS``` for you

ehelms: Good catch!

ehelms:
Updated

roles/ruby_scl/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+    - "{{ ruby_scl_version }}-ruby-devel"
+
+- name: 'Enable SCL at login'
+  blockinfile:
+    dest: /etc/profile.d/enable-{{ ruby_scl_version }}.sh
+    create: yes
+    block: |
+      #!/bin/bash
+      source scl_source enable {{ ruby_scl_version }}
+
```
ekohl: Nit: could you remove the empty line?

akofink:
You missed the `become: true` parts :(

PR #629, opened 2018-02-26
katello_client: don't fail when katello-host-tools are not available

evgeni:


ekohl:
In which case is that not available?

evgeni:
when the user created a CV w/o the katello client bits, which will still work fine as a "dumb" consumer

PR #635, opened 2018-03-02
Avoid using vagrant insecure SSH keys

ekohl:


ekohl:
Agreed there may be dragons

PR #640, opened 2018-03-07
Add SLES boxes

johnpmitsch:
Not sure if these are the right arguments, but the boxes spin up fine :100:

### DIFF
--- a/boxes.d/00-suse.yaml
+++ b/boxes.d/00-suse.yaml
@@ -1,0 +1,12 @@
+# using elastic's image since it is sp3, which is not offered by SUSE yet
+sles12sp3:
+  box_name: elastic/sles-12-x86_64
+  pty: true
+
+sles12sp2:
+  box_name: suse/sles12sp2
+  pty: true
+
+sles12sp1:
+  box_name: suse/sles12sp1
+  pty: true


boxes.d/00-suse.yaml
```diff
@@ -0,0 +1,13 @@
+# using elastic's image since it is sp3, which is not offered by SUSE yet
+sles12:
```
parthaa: may be say sles12sp3

boxes.d/00-suse.yaml
```diff
@@ -0,0 +1,13 @@
...
+  box_name: elastic/sles-12-x86_64
+  shell: 'echo TEST'
+  pty: true
+  libvirt: https://app.vagrantup.com/elastic/boxes/sles-12-x86_64
+
+sles11:
```
parthaa: may be say sles11sp3

boxes.d/00-suse.yaml
jturel: agree!

boxes.d/00-suse.yaml
parthaa: Got an issue installing this box. 
```
==> suse11: Setting hostname...
==> suse11: Configuring and enabling network interfaces...
==> suse11: Rsyncing folder: /home/paji/vagrant/katello-deploy/ => /vagrant
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

find /vagrant '!' -type l -a '(' ! -user vagrant -or ! -group vagrant ')' -exec chown vagrant:vagrant '{}' +

Stdout from the command:



Stderr from the command:

find: `vagrant' is not the name of an existing group
```

boxes.d/00-suse.yaml
```diff
@@ -0,0 +1,13 @@
+# using elastic's image since it is sp3, which is not offered by SUSE yet
+sles12:
+  box_name: elastic/sles-12-x86_64
+  shell: 'echo TEST'
```
ekohl: We don't use the shell provider anymore and I think we ripped out the support.

johnpmitsch:
@ehelms @evgeni Are there any workaround for the following (seen when spinning up sles11 box)
```
Vagrant assumes that this means the command failed!

find /vagrant '!' -type l -a '(' ! -user vagrant -or ! -group vagrant ')' -exec chown vagrant:vagrant '{}' +

Stdout from the command:
find: `vagrant' is not the name of an existing group
```

boxes.d/00-suse.yaml
```diff
@@ -0,0 +1,8 @@
+# using elastic's image since it is sp3, which is not offered by SUSE yet
+sles12sp3:
+  box_name: elastic/sles-12-x86_64
```
evgeni: I'd prefer not to use non-official boxes.

mmoll: as related note: also the JeOS SLES 11SP3 image is not intended for general usage.

It's a long standing problem without real solution. In the past I built SLES baseboxes with packer and distributed these over internal, non-internet-reachable HTTP servers for legal reasons.

johnpmitsch: unfortunately, there is no SUSE supported box for SLES12-sp3. The good news is the suse supported SLES12-sp2 and SLES12-sp1 spin up fine.

Here is my proposal:
- Add SLES12 sp3 image from elastic until suse releases one, then switch to that.
- Use SUSE supported  SLES12-sp2 and SLES12-sp1 currently (updated in this PR)
- Investigate "monkey-patching" SUSEs SLES11-sp3 box by adding a vagrant user and releasing a new box to vagrant cloud, then using that box (this would be done outside of this PR)

johnpmitsch: @mmoll what is the JeOS image? Is that what elastic's image is based on?

mmoll: yeah, but that's gone from this PR now.

ehelms:
@johnpmitsch what is generating that line? I think it means the box itself is lacking a vagrant group when they created it so we'd have to do that ourselves assuming its failing inside the Ansible run and not the Vagrant spin up.

boxes.d/00-suse.yaml
```diff
@@ -0,0 +1,12 @@
...
+  pty: true
+
+sles12sp2:
+  box_name: suse/sles12sp2
+  pty: true
+
+sles12sp1:
+  box_name: suse/sles12sp1
+  pty: true
```
jlsherrill: do these sles boxes actually work?  :/  

I'm getting:

```
URL: ["https://atlas.hashicorp.com/suse/sles12sp2"]
Error: The requested URL returned error: 404 Not Found
```

I get the same thing with the sles11 box i pointed to as well :(

johnpmitsch: the slese12sp2 box spins up fine for me, but vagrant has already downloaded the image so I can't verify that the image downloads ok. I also tried `rm -rf /var/lib/libvirt/images/suse-VAGRANTSLASH-sles12sp2_vagrant_box_image_0.0.1.img` and respun up the box, that worked fine as well. Maybe there were network issues at the time? If you still hit it we can have someone verify (who hasn't spun up the box before)

re SLES11 box, I [commented here](https://github.com/theforeman/forklift/pull/640#issuecomment-371508641) about some issues I ran into spinning it up, so I left that one out of the PR

ehelms:
@johnpmitsch @jlsherrill Any updates here?

boxes.d/00-suse.yaml
johnpmitsch: @jlsherrill were you able to spin up the sles12 boxes?

johnpmitsch:
@ehelms I can remove the SLES12sp3 box if we don't want to use a third-party box, but I know devs will end up using it regardless since we do need to test a lot on 12sp3.

We also need someone can confirm 12sp2 and 12sp1 spin up fine as @jlsherrill saw some trouble with those, but I can't reproduce locally.

johnpmitsch:
closing since SLES boxes aren't a priority anymore and can be added as needed manually by developers who need them. If anyone wants to take this commit and re-open, they are welcome to

PR #646, opened 2018-03-10
Move vagrant into vagrant directory

ehelms:


.gitignore
```diff
@@ -1,4 +1,4 @@
-.vagrant
+*.vagrant*
```
ekohl: Is this change needed? The directory is still named .vagrant.

.gitignore
```diff
@@ -7,15 +7,13 @@ foreman-installer*
 puppet-*
 boxes.yaml
 .idea
-scripts
 repo
-Gemfile.lock
+vagrant/Gemfile.lock
```
ekohl: Maybe this should be moved to ```vagrant/.gitignore``` along with the other vagrant specific ignores? Or at least grouped in this file.

akofink: +1 to `vagrant/.gitignore`

.gitignore
```diff
@@ -7,15 +7,13 @@ foreman-installer*
...
 *.retry
 *.pyc
 user_playbooks
 playbooks/galaxy_roles
 requirements.yml
-.tmp_boxes
```
akofink: I noticed we remove this and don't re-add it anywhere. Are we no longer using it?

ekohl: This was what we used before ```boxes.d/80-tmp-*``` and is indeed no longer used.

ehelms:
I've updated this based on all comments and included the updated migration.

ekohl:
Moving them might be done like the local boxes configuration.

akofink:
It could be, but I don't think it should be. :)

ekohl:
The gemfile.lock is only used if you're running tests, so not that important. The settings file is used when creating hosts so not moving it would create an issue for users.

akofink:
I don't have strong feelings. :)

ehelms:
Updated to migrate the settings.yaml file

Vagrantfile
```diff
@@ -4,23 +4,37 @@
 
 VAGRANTFILE_DIR = File.dirname(__FILE__)
 
-require "#{VAGRANTFILE_DIR}/lib/forklift"
+require "#{VAGRANTFILE_DIR}/vagrant/lib/forklift"
```
ekohl: It might make sense to introduce ```VAGRANT_DIR```

PR #647, opened 2018-03-13
add centos7 fips enabled vm

amitkarsale:
The box will generate a VM with FIPS mode enabled. To spin the machine following command needs to be executed :  `vagrant up centos7-fips`

to verify if FIPS enabled in the newly generated machine execute : 

`cat /proc/sys/crypto/fips_enabled`

it should return 1 if FIPS enabled else 0

### DIFF
--- a/roles/fips/tasks/main.yml
+++ b/roles/fips/tasks/main.yml
@@ -1,0 +1,29 @@
+---
+- name: 'Install dracut-fips'
+  package: 
+    name: dracut-fips
+    state: present
+
+- name: 'Run dracut to rebuild initramfs'
+  command: dracut --force
+
+- name: 'Get boot_uuid'
+  command: 'findmnt -no uuid /boot'
+  register: result
+
+- name: 'Edit kernel command-line to include the fips=1 and boot=UUID=XXXX or boot=LABEL=XXXX or boot=/dev/DEVICE argument'
+  shell: 'grubby --update-kernel=DEFAULT --args="fips=1 boot=UUID={{ result.stdout }}"'
+
+- name: reboot vm
+  shell: sleep 2 && shutdown -r +1 "Ansible reboot"
+  async: 0
+  poll: 0
+  ignore_errors: true
+
+- name: Waiting for reboot
+  local_action: wait_for 
+                host={{ ansible_ssh_host }}
+                state=started
+
+- name: 'Verify FIPS enabled - If failed means the machine is not FIPS enabled'
+  shell: cat /proc/sys/crypto/fips_enabled | grep 1
--- a/playbooks/fips.yml
+++ b/playbooks/fips.yml
@@ -1,0 +1,6 @@
+---
+- hosts: all
+  become : yes
+  gather_facts: no
+  roles:
+    - fips
--- a/vagrant/boxes.d/01-builtin.yaml
+++ b/vagrant/boxes.d/01-builtin.yaml
@@ -23,6 +23,13 @@ boxes:
         - 'playbooks/katello_provisioning.yml'
       group: 'server'
 
+  centos7-fips:
+    box:   'centos7'
+    memory: 8096
+    ansible:
+      playbook: 'playbooks/fips.yml'
+      group: 'fips'
+
   runcible:
     box: centos7
     ansible:


boxes.d/00-centos.yaml
```diff
@@ -19,3 +19,11 @@ boxes:
     box_name: centos7-atomic
     pty: true
     libvirt: http://cloud.centos.org/centos/7/atomic/images/CentOS-Atomic-Host-7-Vagrant-Libvirt.box
+
+  centos7-fips:
```
ekohl: I think this should live in ```boxes.d/01-builtin.yaml```

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: 'Install dracut-fips'
+  action: yum name=dracut-fips state=installed
```
ekohl: We tend to space it out a bit with newlines.

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
```
amitkarsale: line breaks are added.

roles/fips/tasks/main.yml
amitkarsale: line breaks are added.

roles/fips/tasks/main.yml
ekohl: I'm unfamiliar with action. Is there a reason plain ```yum``` doesn't work?

ekohl:
I wonder if the playbook could be made idempotent. I tend to apply playbooks multiple times and it rebooting every time is somewhat tedious.

roles/fips/defaults/main.yml
```diff
@@ -0,0 +1 @@
+---
```
ehelms: If an empty file, I'd rather we just omit it all together.

roles/fips/tasks/main.yml
ehelms: Or can we use `package` module here? I also tend to prefer that we use syntax like:

```
package:
  name: dracut-fips
  state: present
```

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  command: grubby --update-kernel=DEFAULT --args=fips=1
+
+- name: 'Get uuid'
+  command: 'findmnt -no uuid /boot'
+  register: result
+
+- name: 'Edit kernel command-line to include the boot=UUID=XXXX or boot=LABEL=XXXX or boot=/dev/DEVICE argument'
+  command: "grubby --update-kernel=DEFAULT --args=boot=UUID={{result.stdout}}"
+
+- name: Restart server
```
ehelms: Should we add a step to wait for the server to come back up before exiting? We could then add in the check to output whether FIPs was properly enabled or not rather than having it as a manual user step.

roles/fips/tasks/main.yml
amitkarsale: I just started learning Anisble and this is my first playbook. Thanks for commenting on that.
`Package` module makes more sense than `action` and `yum` module.

roles/fips/tasks/main.yml
amitkarsale: have tried `wait_for` module over here but didn't find any success, its going timeout to reach the machine after reboot, any help over this @ehelms

ehelms: I think this is a good case for a `handler` in Ansible parlance, and then attaching that handler to the tasks above so that if they change the handler to reboot is initiated. If nothing changes, no reboot.

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- name: 'Edit kernel command-line to include the fips=1 argument'
+  command: grubby --update-kernel=DEFAULT --args=fips=1
+
+- name: 'Get uuid'
+  command: 'findmnt -no uuid /boot'
+  register: result
+
+- name: 'Edit kernel command-line to include the boot=UUID=XXXX or boot=LABEL=XXXX or boot=/dev/DEVICE argument'
+  command: "grubby --update-kernel=DEFAULT --args=boot=UUID={{result.stdout}}"
```
ehelms: A formatting nicety here would be nice `{{ result.stdout }}`

roles/fips/tasks/main.yml
ehelms: Have you tried this pattern? https://www.linux.com/blog/ansible-reboot-server-play-book-and-wait-it-come-back

amitkarsale:
any changes suggested @ehelms @ekohl or is it good to merge..?

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: 'Install dracut-fips'
+  package: 
+    name: dracut-fips
+    state: present
+
+- name: 'Run dracut to rebuild initramfs'
+  command: dracut --force
+
+- name: 'Edit kernel command-line to include the fips=1 argument'
+  command: grubby --update-kernel=DEFAULT --args=fips=1
```
ekohl: Is it possible to make this idempotent? Will this be persisted when a kernel is updated?

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: 'Get uuid'
+  command: 'findmnt -no uuid /boot'
+  register: result
```
ekohl: Maybe a better name would be ```boot_uuid```?

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: 'Edit kernel command-line to include the boot=UUID=XXXX or boot=LABEL=XXXX or boot=/dev/DEVICE argument'
+  command: "grubby --update-kernel=DEFAULT --args=boot=UUID={{ result.stdout }}"
```
ekohl: Should this be combined with the other grubby command?

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: reboot vm
+  shell: sleep 2 && shutdown -r +1 "Ansible reboot"
+  async: 0
+  poll: 0
+  ignore_errors: true
+
+- name: Waiting for reboot
+  local_action: wait_for host={{ ansible_ssh_host }} state=started
```
ekohl: I think we typically split this in multiple lines.

roles/fips/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: 'Verify FIPS enabled - 1 : Enabled ; 0 : Disabled '
```
ekohl: Perhaps this needs to be compared and the role failed if it doesn't match. When you enable multiple roles this will be easy to miss.

roles/fips/tasks/main.yml
amitkarsale: @ekohl I've tried on updating the kernel but the idempotency is not persisted. I've tried on setting the current kernel as default using `grubby --set-default <kernal_path>` but this doesn't make sense as updating the kernel and not using it pointless.

Do you have any thoughts of how we can achieve it.

amitkarsale:
@ekohl I've tried on updating the kernel but the idempotency is not persisted. I've tried on setting the current kernel as default using grubby --set-default <kernal_path> but this doesn't make sense as updating the kernel and not using it pointless.

Do you have any thoughts of how we can achieve it.

amitkarsale:
@ekohl @ehelms have tested the host with z-stream update and the FIPS flags is persisted. Do we have any fine-tuning left for it ?

ehelms:
@amitkarsale This will at least need a rebase.

playbooks/centos7_fips.yml
```diff
@@ -0,0 +1,6 @@
+---
```
ehelms: The playbook can just be named `fips.yml` I think.

amitkarsale: Done with the rebase and file name change.

ehelms:
@ekohl Take a final glance at this?

ekohl:
Thanks!

PR #648, opened 2018-03-13
New role for setup the vagrant user

BlackSmith:
This is necessary for installation from scratch (without vagrant).


ehelms:
Can you point me at the code that is tied tightly to vagrant user and requiring this sort of work around? I'd like to look to see if we should fix that code rather than introduce creating a vagrant user.

BlackSmith:
If users use the vagrant, then they have this user as default in their images. But in situation, when we use playbooks alone (devel, luna, ...) this user is missing. I know, that we can change the vagrant user by foreman options and my solution is dynamic as well. But vagrant name is as good as any other name.

ares:
:+1: fixes the problem for me, otherwise I was getting

```
[ WARN 2018-03-19T07:29:22 verbose]  Compiled catalog for fedora-build02.... in environment production in 2.62 seconds
[ INFO 2018-03-19T07:29:23 verbose]  Applying configuration version '1521458960'
[ WARN 2018-03-19T07:29:35 verbose]  /Stage[main]/Katello_devel::Setup/Katello_devel::Bundle[install --without mysql:mysql2 --retry 3 --jobs 3]/Exec[bundle-install --without mysql:mysql2 --retry 3 --jobs 3]/returns: Ruby ruby-2.4.1 is not installed.
[ERROR 2018-03-19T07:29:35 verbose]  'rvm 2.4 do bundle install --without mysql:mysql2 --retry 3 --jobs 3' returned 2 instead of one of [0]
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/errors.rb:157:in `fail'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type/exec.rb:164:in `sync'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction/resource_harness.rb:236:in `sync'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction/resource_harness.rb:134:in `sync_if_needed'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction/resource_harness.rb:88:in `block in perform_changes'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction/resource_harness.rb:87:in `each'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction/resource_harness.rb:87:in `perform_changes'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction/resource_harness.rb:21:in `evaluate'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction.rb:257:in `apply'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction.rb:277:in `eval_resource'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction.rb:181:in `call'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction.rb:181:in `block (2 levels) in evaluate'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util.rb:517:in `block in thinmark'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/2.4.0/benchmark.rb:308:in `realtime'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util.rb:516:in `thinmark'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction.rb:181:in `block in evaluate'
[ERROR 2018-03-19T07:29:35 verbose] /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/graph/relationship_graph.rb:121:in `traverse'
```

pgagne:
When I tried to just run the playbooks against a beaker machine I had a lot of issues that i think were related to the fact the playbooks assume there is a user called vagrant. Including putting stuff in /home/vagrant. All this despite the fact that i set `ansible_user:  root` in my inventory vars.

IMHO I would rather see us fix the paths in the playbook to be more generic and not have to create another user called vagrant, which might be confusing to some new users.

ekohl:
@pgagne that makes sense, but in any case you want an unprivileged user for a devel scenario so would ```unprivileged_user``` be a good name?

Other than that I think the authorized keys should be flexible and use a parameter. It makes sense to to copy ```~/.ssh/.id_rsa.pub``` by default since that's likely to be present, unlike ```authorized_keys```.

pgagne:
Yes `unprivileged_user` would be good. 

We also need to go through a find absolute references to the vagrant user and home dir and update them, probably in a separate pull request. 

Here is a search to get started: https://github.com/theforeman/forklift/search?utf8=%E2%9C%93&q=%2Fhome%2Fvagrant&type=

I would like to do this, but am tied up at the moment, so if someone else wants to dive in more that would be great.

ehelms:
@ekohl @pgagne Are you requesting changes to this PR or is it good to go as is?

BlackSmith:
@pgagne, @ekohl  I have fixed the problem with unprivileged user. 
Either way, the playbooks need unprivileged user. The name is not so important, if we can change them.
The main reason for this role is preparing of environment for other roles (which have to run as unprivileged user). From my point of view the file authorized_keys is necessary (if it does not exist).

@ekohl Why we need ssh keys there? It is not problem to do it. The question is, should we generate new ones or copy? Should we generate/copy only one pair or all keys and which kind (rsa, dsa, ed...)?

@pgagne The fix of static declaration of username/home folder can be the next step.

ekohl:
> @ekohl Why we need ssh keys there? It is not problem to do it. The question is, should we generate new ones or copy? Should we generate/copy only one pair or all keys and which kind (rsa, dsa, ed...)?

My comment wasn't about `authorized_keys` inside the VM but on the host you're running ansible on. On my laptop I don't have `authorized_keys` but do have `id_rsa.pub` so I think that's a better default. If you make it a parameter you can have the best of both worlds (or use other key types).

BlackSmith:
@ekohl Oh, my mistake. 
I have added new option vagrant_user_import_ssh_pub_key. If this is set to pub key file, it is added into authorized_keys. In other case, it copies authorized_keys from root (if exist).

jhutar:
Hello. Are we basically done here?

BlackSmith:
@ekohl  Done

roles/unpriviledged_user/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+unpriviledged_user_username: vagrant
+# unpriviledged_user_import_ssh_pub_key:
```
ehelms: Given this is commented out, can it be dropped?

BlackSmith: I define all variables which are use by script (my habit.).  It is easier for next using. I can see all possible options (especialy correct names or some comments for it) on the one place and I don't have to read whole source code. In this case, the variable unpriviledged_user_import_ssh_pub_key undefined in default.
If you have a problem with this, I can remove this.

ehelms: I'd prefer it removed or set to a default value. The point of defaults is just that, to set variables and give them a default in case no user supplies them.

BlackSmith:
Done.

PR #649, opened 2018-03-13
Add a few new features to foreman_devel

BlackSmith:
New features:
- the checking of user's plugin forks
- the checking of registration via SSM
- adding a new record to /etc/hosts, if the machine is behind NAT (OpenStack instances)

### DIFF
--- a/roles/foreman_devel/tasks/check_forks_of_repos.yml
+++ b/roles/foreman_devel/tasks/check_forks_of_repos.yml
@@ -1,0 +1,14 @@
+---
+- name: "Check forks of user's repository"
+  uri:
+    url: "https://github.com/{{ katello_devel_github_username }}/{{ item.split('/') | last }}"
+  with_items: "{{ foreman_installer_options_internal_use_only }}"
+  when: item | search("katello-devel-extra-plugins")
+
+
+- name: "Check forks of user's katello & foreman repository"
+  uri:
+    url: "https://github.com/{{ katello_devel_github_username }}/{{ item }}"
+  with_items:
+    - katello
+    - foreman
--- a/roles/foreman_devel/tasks/main.yml
+++ b/roles/foreman_devel/tasks/main.yml
@@ -5,3 +5,10 @@
 - name: setup reviewing script
   include_role: 
     name: reviewer
+
+- include_tasks: check_forks_of_repos.yml
+  when: katello_devel_github_username is defined and foreman_installer_options_internal_use_only is defined
+
+- include_tasks: rhel.yml
+  when: ansible_distribution == "RedHat"
+
--- a/roles/foreman_devel/tasks/main.yml
+++ b/roles/foreman_devel/tasks/main.yml
@@ -1,14 +1,13 @@
 ---
-- include_tasks: github_push_ssh.yml
-  when: foreman_devel_github_push_ssh
-
-- name: setup reviewing script
-  include_role: 
-    name: reviewer
-
 - include_tasks: check_forks_of_repos.yml
   when: katello_devel_github_username is defined and foreman_installer_options_internal_use_only is defined
 
 - include_tasks: rhel.yml
   when: ansible_distribution == "RedHat"
 
+- include_tasks: github_push_ssh.yml
+  when: foreman_devel_github_push_ssh
+
+- name: setup reviewing script
+  include_role: 
+    name: reviewer
--- a/roles/foreman_devel/tasks/rhel.yml
+++ b/roles/foreman_devel/tasks/rhel.yml
@@ -1,0 +1,22 @@
+---
+- name: "Check the registration"
+  shell: "subscription-manager identity > /dev/null"
+  register: sresult
+  become: true
+  ignore_errors: True
+
+- name: "Stop installation; the system is not registered"
+  fail:
+    msg: "Please register the system via subscription-manager before continuing"
+  when: "sresult.rc != 0"
+
+- name: "Enable only RHEL7 base + optional + extras repos"
+  shell: "{{ item }}"
+  become: true
+  with_items:
+    - subscription-manager repos --disable='*'
+    - >
+      subscription-manager repos 
+      --enable="rhel-7-server-rpms"
+      --enable="rhel-7-server-extras-rpms"
+      --enable="rhel-7-server-optional-rpms"
--- a/roles/foreman_devel/tasks/rhel.yml
+++ b/roles/foreman_devel/tasks/rhel.yml
@@ -10,13 +10,17 @@
     msg: "Please register the system via subscription-manager before continuing"
   when: "sresult.rc != 0"
 
+- name: "Disable all repos"
+  rhsm_repository:
+    name: "*"
+    state: "disabled"
+
 - name: "Enable only RHEL7 base + optional + extras repos"
-  shell: "{{ item }}"
+  rhsm_repository:
+    name: "{{ item }}"
+    state: "present"
   become: true
   with_items:
-    - subscription-manager repos --disable='*'
-    - >
-      subscription-manager repos 
-      --enable="rhel-7-server-rpms"
-      --enable="rhel-7-server-extras-rpms"
-      --enable="rhel-7-server-optional-rpms"
+      - rhel-7-server-rpms
+      - rhel-7-server-extras-rpms
+      - rhel-7-server-optional-rpms
--- a/roles/reviewer/tasks/main.yml
+++ b/roles/reviewer/tasks/main.yml
@@ -1,3 +1,8 @@
+- name: "Install git"
+  package:
+    name: 'git'
+    state: 'present'
+
 - name: Install reviewer tool from git
   git:
     repo: "{{ reviewer_install_repo }}"


roles/foreman_devel/tasks/check_forks_of_repos.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: "Check forks of user's repository"
+  uri:
+    url: "https://github.com/{{ katello_devel_github_username }}/{{ item.split('/') | last }}"
+    no_log: True
+  with_items: "{{ foreman_installer_options_internal_use_only }}"
```
ekohl: This looks like a coupling between roles and I'm not sure that's a good thing. Also note that we have the normal installer options users can specify but that's not an array so harder to use in this way.

roles/foreman_devel/tasks/main.yml
```diff
@@ -1,3 +1,27 @@
 ---
 - include_tasks: github_push_ssh.yml
   when: foreman_devel_github_push_ssh
+
+- include_tasks: check_forks_of_repos.yml
+  when: katello_devel_github_username is defined and foreman_installer_options_internal_use_only is defined
+
+- name: "Check situation, if the machine is behind NAT"
+  # for example OpenStack instances
+  # This is potencional problem, if user has the same private IP in own lan.
```
ekohl: s/potencional/potential/

roles/foreman_devel/tasks/main.yml
```diff
@@ -1,3 +1,27 @@
...
+
+- name: "Check situation, if the machine is behind NAT"
+  # for example OpenStack instances
+  # This is potencional problem, if user has the same private IP in own lan. 
+  shell: "ping -c 1 {{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}"
+  delegate_to: localhost
+  ignore_errors: True
+  register: rping
+
+- name: "Add OS hostname to hosts file"
```
ekohl: This looks a lot like our ```etc_hosts``` role

roles/foreman_devel/tasks/check_forks_of_repos.yml
```diff
@@ -0,0 +1,16 @@
+---
```
ekohl: What is this intended to do? It doesn't log and it looks like there's no further related actions to it.

roles/foreman_devel/tasks/check_forks_of_repos.yml
BlackSmith: Yes, I understand. This is based on https://github.com/theforeman/forklift/pull/623.
This is really important for newone (as am I), who does not have all forks of gits and the install proces fails in the middle. The better way is checking all requirements in the beginning.

roles/foreman_devel/tasks/main.yml
BlackSmith: Fixed

roles/foreman_devel/tasks/check_forks_of_repos.yml
BlackSmith: I'm not sure, what do you mean ? pls more info

ehelms: I think @ekohl is asking what is the intent of this set of tasks? To bomb out if forks don't exist and warn the user?

roles/foreman_devel/tasks/main.yml
ehelms: I think this could just be done, without the check starting at line 8. The `lineinfile` module is meant to only make changes if a file is not present. So seems like there would be no harm in always ensuring there is an entry in `/etc/hosts` ?

roles/foreman_devel/tasks/main.yml
```diff
@@ -1,3 +1,27 @@
...
+  lineinfile:
+    dest: /etc/hosts
+    regexp: ".*{{ ansible_fqdn }}.*$"
+    line: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }} {{ ansible_fqdn }}"
+    state: present
+  become: true
+  when: "rping.rc != 0"
+
+- include_tasks: rhel.yml
+  when: ansible_distribution == "RedHat"
```
ehelms: Doesn't CentOS fall under this distribution family? In which case, the actions in the task wouldn't make as much sense.

roles/foreman_devel/tasks/check_forks_of_repos.yml
BlackSmith: Yes, @ehelms  is right. It is very important that it is testing before the main installation. From my experience, the installation process in not idempotent.

roles/foreman_devel/tasks/main.yml
BlackSmith: @ehelms No problem, I'm removing it.

roles/foreman_devel/tasks/main.yml
BlackSmith: Centos has got these values:
  "ansible_distribution": "CentOS",
  "ansible_distribution_file_variety": "RedHat", 
  "ansible_distribution_major_version": "7", 
  "ansible_distribution_release": "Core", 
  "ansible_distribution_version": "7.2.1511",

roles/foreman_devel/tasks/check_forks_of_repos.yml
ekohl: @BlackSmith then we should look into why it's not idempotent.

BlackSmith: @ekol I fully agree, but It is now  over my abilities. :(

ekohl: @BlackSmith if you have specific examples I could have a look.

roles/foreman_devel/tasks/main.yml
BlackSmith: done

roles/foreman_devel/tasks/check_forks_of_repos.yml
BlackSmith: @ekohl What I remember, one of my problem was missing optional or extras repo in RHEL7. Next one is missing of some plugin or katello forks.  The problem with puppet-katello_devel (fixed in pull request 145), problem with IP 169.254.169.254, ...
I'm not sure which one was fatal. Because after few bad tries with continuing, I started to kill old machines and I tried it again with new one.
It is true that before your help (with katello_repositories_environment: staging), we used a little different scenario/playbook. There was for example sometimes problem with "npm install || node ./script/npm_install_plugins.js".  (Maybe you remember https://github.com/BlackSmith/forklift/blob/master/playbooks/full-lunar-devel.yml). Maybe many of this problems are stupid and they are caused only with bad environment, attributes, etc. But it is true, this project has no good documentation and many things are implicite.

ekohl: You're right and we should work on that. It's always been a case where someone solved their own problems. Within the open source community I don't think people really use RHEL7 but opt for CentOS7 which is always easier to use. I tend to ask these kind of questions to get some sense of how it's used and the reasoning so I can try to aim at a unified structure.

roles/foreman_devel/tasks/main.yml
```diff
@@ -1,3 +1,24 @@
...
+    regexp: ".*{{ ansible_fqdn }}.*$"
+    line: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }} {{ ansible_fqdn }}"
+    state: present
+  become: true
+
+- include_tasks: rhel.yml
+  when: ansible_distribution == "RedHat"
+
+- name: "Install systemd-devel"
+  # This is required by journald-native gem
```
ehelms: @ekohl is this related to the structured logging that went in?

roles/foreman_devel/tasks/main.yml
```diff
@@ -1,3 +1,24 @@
...
+- name: "Add OS hostname to hosts file" 
+  lineinfile:
+    dest: /etc/hosts
+    regexp: ".*{{ ansible_fqdn }}.*$"
+    line: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }} {{ ansible_fqdn }}"
+    state: present
+  become: true
+
+- include_tasks: rhel.yml
+  when: ansible_distribution == "RedHat"
```
ehelms: We'll want to test this against CentOS to ensure it doesn't break.

roles/foreman_devel/tasks/main.yml
ekohl: Yes. I'd prefer this to be added to https://github.com/theforeman/puppet-katello_devel/blob/e52881cfccffa16de434e80bd2af4a2ad23b9b1b/manifests/install.pp#L4 so it all lives in one place.

BlackSmith: @ekohl  OK, no problem. Should I add it there in this pull request or you create a new pull requests for it?

ekohl: https://github.com/theforeman/puppet-katello_devel/pull/151

roles/foreman_devel/tasks/main.yml
```diff
@@ -1,3 +1,18 @@
 ---
 - include_tasks: github_push_ssh.yml
   when: foreman_devel_github_push_ssh
+
+- include_tasks: check_forks_of_repos.yml
+  when: katello_devel_github_username is defined and foreman_installer_options_internal_use_only is defined
+
+- name: "Add OS hostname to hosts file"
```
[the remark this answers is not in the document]
BlackSmith: If I understand this 
https://github.com/theforeman/forklift/blob/2e706dedd792af67cefe31feeb125c3d33fcc044/roles/etc_hosts/tasks/main.yml#L1-L12 
it is vagrant only stuff. 
The variables like 'ansible_host' are not defined by ansible, they are defined by dynamic inventory for vagrant 
https://github.com/theforeman/forklift/blob/2e706dedd792af67cefe31feeb125c3d33fcc044/inventories/vagrant.py#L53-L56. 

It does not solve my problem (the IP address and DNS A/PTR record are different).

ekohl:
@jhutar that's exactly what I meant. We implement an ansible inventory that reads from vagrant, but it's still the ansible inventory format.

BlackSmith:
I see! OK, this solution is not so automatic as mine, but I can remove it.

jhutar:
Hello @ekohl again :-) Does it look OK now?

jhutar:
Or @ehelms, does this look OK now? This is quite useful for QE working on the Foreman/Katello project.

roles/foreman_devel/tasks/rhel.yml
```diff
@@ -0,0 +1,22 @@
...
+  register: sresult
+  become: true
+  ignore_errors: True
+
+- name: "Stop installation; the system is not registered"
+  fail:
+    msg: "Please register the system via subscription-manager before continuing"
+  when: "sresult.rc != 0"
+
+- name: "Enable only RHEL7 base + optional + extras repos"
```
ekohl: I think [rhsm_repository](https://docs.ansible.com/ansible/2.5/modules/rhsm_repository_module.html) is cleaner

roles/foreman_devel/tasks/github_push_ssh.yml
```diff
@@ -9,3 +9,4 @@
     name: 'url."git@github.com:".pushInsteadOf'
     value: 'https://github.com/'
     scope: system
+  when: foreman_devel_github_push_ssh
```
ekohl: I'd prefer to keep this conditional in main. IMHO it's better if the `reviewer` role ensures git as well. It's a standalone role and it should work standalone.

BlackSmith:
Hello @kohl, 
I have rebased it and rewritten it with rhsm_repository.
However I have to change the ordering of tasks. New task "reviewer" has got broken dependence. If variable  foreman_devel_github_push_ssh is not true then git package is not installed and reviewer fails.

ehelms:
@ekohl Appears ready for re-review

ekohl:
Thanks!

PR #659, opened 2018-03-16
Add puppetlabs-vcsrepo install

ehelms:


roles/foreman_installer_devel_scenario/tasks/main.yml
```diff
@@ -23,6 +23,9 @@
     dest: /usr/share/katello-installer-base/modules/katello_devel
     update: yes
 
+- name: 'Install puppetlabs-vcsrepo'
+  command: puppet module install --target-dir /usr/share/katello-installer-base/modules puppetlabs-vcrepo
```
ekohl: Should we ensure puppet is installed before doing this?

roles/foreman_installer_devel_scenario/tasks/main.yml
```diff
@@ -23,6 +23,14 @@
     dest: /usr/share/katello-installer-base/modules/katello_devel
     update: yes
 
+- name: 'Install puppet-agent'
+  package:
+    name: puppet-agent
```
ehelms: @ekohl Added

PR #662, opened 2018-03-19
Cert for webpack on chrome

pgagne:
I was having an issue where it seems JS wasn't loading and UI elements could not be used. It seemed to be caused by ERR_CERT_AUTHORITY_INVALID issues. 

This was running chrome, so I didn't try this at first.

I followed these instruction and it seems to have fixed my issue.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -59,9 +59,9 @@ cd foreman
 bundle exec foreman start
 ```
 
-### Webpack dev server & Firefox
+### Webpack dev server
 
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -59,9 +59,9 @@ cd foreman
 bundle exec foreman start
 ```
 
-### Webpack dev server & Firefox
+### Webpack dev server
 
-When using the dev server on Firefox you need to accept the self-signed certificate. You should go to `https://centos7-devel.example.com:3808` (or your equivalent) and add an exception for this certificate. Chrome does not throw any warnings if you have accepted the same certificate already for `https://centos7-devel.example.com`.
+When using the dev server on Firefox and Chrome you need to accept the self-signed certificate. You should go to `https://centos7-devel.example.com:3808` (or your equivalent) and add an exception for this certificate.
```
jturel: How about 'when using the dev server you may need to accept...' so we can avoid listing every browser imaginable?

ekohl:
Thanks!

PR #665, opened 2018-03-21
document default credentials and how to change them

evgeni:


README.md
```diff
@@ -51,6 +51,10 @@ vagrant up centos7-devel
 
 You can find more thorough guides in the [docs folder](https://github.com/theforeman/forklift/tree/master/docs).
 
+### Credentials
+
+By default `forklift` deploys Foreman with `admin`/`changeme` as username and password, please change this in production installs (either after the install, or by setting `foreman_installer_admin_password`).
```
ekohl: Might be good to mention that rerunning the installer does *not* update the password. If you've seeded, then you need to do it inside the application.

ehelms: I think you might want `this on production installs`

evgeni: updated. I did not want to bother the user with seeding etc, so I just wrote "during the initial deployment", tell me what you think :)

PR #669, opened 2018-03-22
Add flag foreman_installer_skip to setup repositories only and quit

lpramuk:
Similar flag to foreman_installer_skip_installer but skipping both packages and installer
Useful when you want just setup repos and proceed manually what to install/upgrade.

E.g:
You got Nightly and run ansible to get Koji repos but don't want to apply package upgrade/install at all. 

Not sure about naming, I'm open to suggestions: 

1. foreman_installer_skip    <- chose this one
2. foreman_installer_skip_all
3. foreman_installer_no_packages

### DIFF
--- a/roles/foreman_installer/defaults/main.yml
+++ b/roles/foreman_installer/defaults/main.yml
@@ -1,6 +1,7 @@
 ---
 foreman_installer_additional_packages: []
 foreman_installer_admin_password: changeme
+foreman_installer_skip: False
 foreman_installer_skip_installer: False
 foreman_installer_verbose: True
 foreman_installer_scenario: foreman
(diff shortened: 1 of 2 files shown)


ekohl:
I'm not sure this is the right approach. I'd recommend including the repository roles manually (which we do in a lot of places already). Perhaps a new meta role to include all the repositories would make sense.

lpramuk:
all meta role katello fits except its last role foreman_installer to have DRY I just did this PR
Ofc I can WET and WET and WET - the same meta role except last role and no parametrization.

Not familiar with ansible (yet?) to choose whether WET or DRY is better ...

ehelms:
I think I tend to agree with @ekohl that a separate playbook or meta role makes sense if the feature request is "I want to setup all required repositories but do nothing else". This would in general be useful in multiple scenarios.

ekohl:
Closing this because a new meta role would be a better path forward.

PR #670, opened 2018-03-22
Fix the centos7-hammer-devel box to `become` less

akofink:
Related to https://github.com/theforeman/forklift/pull/627 and https://github.com/theforeman/forklift/pull/668.

We want to install a hammer_devel environment in the `"{{ remote_user }}"`'s home directory, not root's.

### DIFF
--- a/playbooks/hammer_devel.yml
+++ b/playbooks/hammer_devel.yml
@@ -1,8 +1,8 @@
 ---
 - hosts: all
-  become: yes
   roles:
     - etc_hosts
     - git
-    - ruby_scl
+    - role: ruby_scl
+      become: yes
     - hammer_devel


playbooks/hammer_devel.yml
```diff
@@ -1,8 +1,7 @@
 ---
 - hosts: all
-  become: yes
   roles:
     - etc_hosts
     - git
-    - ruby_scl
+    - { role: ruby_scl, become: yes }
```
ehelms: Can this be:

```
- role: ruby_scl
  become: yes
```

I find the hash syntax a bit jarring in yaml.

akofink: Yes. Yes it can. :)

akofink:
Updated

akofink:
Would love to get this in so centos7-hammer-devel isn't broken. :)

ekohl:
Would it be a best practice to use ```become: true``` in the role instead? Here my ansible knowledge is limited.

akofink:
@ekohl That's how it was originally implemented until @ehelms changed it. Perhaps @ehelms could explain his reasoning and offer an opinion?

akofink:
Unless removing the `become` was a mistake, in which case, I'm not opposed to re-adding it on the role itself. @ehelms?

PR #672, opened 2018-03-27
redhat_subscription module requires activationkey with org_id

ehelms:


ekohl:
I'm not exactly sure what to make of this PR. The title suggests it needs an `org_id` but here you remove it. Does that mean the `org_id` is in the AK?

ehelms:
Well, it has to do with the specific task -- "Register client with subscription-manager as user". This is registering with username/password which can't have an org_id

PR #673, opened 2018-04-04
Restore backward compatibility with 1.16

xprazak2:
Some of the tasks rely on latest features in nightly which breaks backward compatibility. The role should not fail when used for 1.16 or 1.17

### DIFF
--- a/roles/foreman_provisioning/defaults/main.yml
+++ b/roles/foreman_provisioning/defaults/main.yml
@@ -11,3 +11,4 @@ foreman_provisioning_dhcp_end: 192.168.73.254
 foreman_provisioning_network: 192.168.73.0
 foreman_provisioning_installer_options: []
 foreman_provisioning_domain: "{{ ansible_domain }}"
+foreman_provisioning_foreman_version: "{{ foreman_repositories_version | default('nightly') }}"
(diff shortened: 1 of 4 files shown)


ekohl:
The annoying part about this is that you can't rely on variables from another role. When you just run this role, it has different behavior. I'm not sure what the ansible best practices are in this case.

ehelms:
This role could:

 1) Depend on the foreman_repositories role
 2) Set a fact based on parsing the Foreman version found by looking at the Foreman version file installed or by querying an API

roles/foreman_provisioning/defaults/main.yml
```diff
@@ -11,3 +11,4 @@ foreman_provisioning_dhcp_end: 192.168.73.254
 foreman_provisioning_network: 192.168.73.0
 foreman_provisioning_installer_options: []
 foreman_provisioning_domain: "{{ ansible_domain }}"
+foreman_provisioning_foreman_version: "{{ foreman_repositories_version if foreman_repositories_version is defined else 'nightly' }}"
```
ekohl: Isn't `{{ foreman_repositories_version|default('nightly') }}` equivalent?

xprazak2: I learn something new every day :smile:, updated.

PR #674, opened 2018-04-09
Fix hostname for OpenStack instances

BlackSmith:
This method of build hosts file is inadequate for OpenStack. 
https://github.com/theforeman/forklift/blob/master/roles/etc_hosts/tasks/main.yml#L2-L12

The original method create a fictive hostname for vagrnat instances. In OpenStack we need a little different view, we already have the real fqdn, but we are behind NAT. DNS server returns  our public IP (address of our router)  for our hostname.  This new task is run only if we are in a cloud (OpenStack).

### DIFF
--- a/roles/etc_hosts/tasks/main.yml
+++ b/roles/etc_hosts/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-- name: "Build hosts file"
+- name: "Build hosts file (vagrant)"
   lineinfile:
     dest: /etc/hosts
     regexp: ".*{{ item.replace('.', '-') }}$"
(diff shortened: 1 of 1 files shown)


ekohl:
This is interesting. We use this in rackspace but with public IPs. However, on Rackspace the 169-IP address that's standard is actually dropped (rather than rejected) so I this fact wouldn't resolve and not change anything for this.

BlackSmith:
I'm not sure how is Rackspace configured (I have no access there).  Is there the public IP setup on instance? If the 169-IP addresses are from 169.254.0.0/16, it's "link local" and yes, it is dropped by first router (the same as private ranges of IP). 

If the public IPs is assigned directly to instances, it can be that. 
But my experience with OpenStack is the instances have got only private IPs and public one is assigned on NAT. Our OpenStack uses neutron, maybe it can be difference between our OpenStack and RackSpace.

BlackSmith:
I find out a bad thing,  facts facter_* is present only on machines, where is foreman installed. :( 
It is missing on clean machines. I have to investigation it or find a different way how to detect the situation, the instance is behind NAT.

ekohl:
I think that if you explicitly install facter that's enough for those facts to be present. It's packaged in EPEL, but that could be an issue with pure RHEL systems.

ehelms:
@BlackSmith Planning to come back to this PR?

BlackSmith:
@ehelms Yes, you right. This PR is a little out of date.  I can not to solve this now, maybe in future.

PR #675, opened 2018-04-09
Adjust bats-puppet test to look for hostname instead of the last host…

zjhuntin:
… in the list

### DIFF
--- a/bats/fb-test-puppet.bats
+++ b/bats/fb-test-puppet.bats
@@ -61,8 +61,7 @@ fi
 }
 
 @test "import ntp puppet class" {
-  id=$(hammer --csv proxy list | tail -n1 | cut -d, -f1)
-  hammer proxy import-classes --id $id
+  hammer proxy import-classes --name $(hostname -f)
   count=$(hammer --csv puppet-class list --search 'name = ntp' | wc -l)
   [ $count -gt 1 ]
 }


bats/fb-test-puppet.bats
```diff
@@ -61,7 +61,7 @@ fi
 }
 
 @test "import ntp puppet class" {
-  id=$(hammer --csv proxy list | tail -n1 | cut -d, -f1)
+  id=$(hammer --csv proxy list | grep $HOSTNAME | cut -d, -f1)
```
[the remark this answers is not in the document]
ehelms: Oh yea, that makes sense.

bats/fb-test-puppet.bats
```diff
@@ -61,7 +61,7 @@ fi
 }
 
 @test "import ntp puppet class" {
-  id=$(hammer --csv proxy list | tail -n1 | cut -d, -f1)
+  id=$(hammer --csv proxy list --name $HOSTNAME | cut -d, -f1)
```
ekohl: Everywhere else we use `$(hostname -f)`. Is `$HOSTNAME` also defined and should we rely on that instead?

zjhuntin: Sure, I think that makes the most sense.

ekohl: I'd don't care which one we use, but I'd like it to be consistent. Could you update it?

zjhuntin:
@ehelms @ekohl updated, figured out it was --search, not name we needed, and now it's consistent using $(hostname -f)

bats/fb-test-puppet.bats
```diff
@@ -61,7 +61,7 @@ fi
 }
 
 @test "import ntp puppet class" {
-  id=$(hammer --csv proxy list | tail -n1 | cut -d, -f1)
+  id=$(hammer --csv proxy list --search $(hostname -f) | tail -n1 | cut -d, -f1)
```
ekohl: I'd expect you should be able to use `hamer proxy info --name $(hostname -f)` and not need to use tail. You might even be able to simplify it all to `hammer proxy import-classes --name $(hostname -f)`.

zjhuntin: Totally, right that works great.

PR #676, opened 2018-04-09
[WIP] Add ansible bats test

ekohl:
This is not yet complete (needs installer work) and the foreman-installer call is not in line with the other tests, but I need a place to store this.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,6 +2,8 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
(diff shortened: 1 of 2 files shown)


ehelms:
@ekohl Is there a plan for including this work?

ehelms:
Thoughts on revisiting this? or do things like smoker make this closeable as a future enhancement there instead?

ekohl:
At this point in time I have no time for this. Smoker takes a different approach (only test what's available), but I do hope to cover plugins end-to-end at some point.

PR #681, opened 2018-04-11
making module_pr work with puppet-foreman

pondrejk:
Trying to use foreman_installer_module_prs to test PRs from puppet-foreman repo fails with 
```
remote: Repository not found.
fatal: repository 'https://github.com/theforeman/puppet-puppet-foreman/' not found
```
This pr removes the `puppet-` prefix if it is supplied in the project name in foreman_installer_module_prs, if there is a more elegant way how to do this in an ansible loop, I'd welcome a suggestion :)


ekohl:
Typically we call it with `theforeman/foreman/123` which avoids the issue. I'd love if we could deal with this in a more generic way. If we could remove `*-` the we could also use things like `puppetlabs-` and test `puppetlabs/postgresql/950`. That means we should also fix that ternary to determine the location though.

PR #684, opened 2018-04-19
test pipeline for a loadbalanced proxy setup

evgeni:
this adds a new pipeline that can be used to test basic proxy loadbalancing for katello by using an haproxy setup. it's not a full-blown setup as proposed in https://community.theforeman.org/t/user-survey-supporting-ha-smart-proxies/8832, just something that can be done with todays code already.

# what does this playbook do?
Installs a Katello 3.6, two proxies and places an HAproxy infront of these two, balancing all the services between both proxies. Then a client is added to test that setup

# current caveats

1. PuppetCA is not really load-balanceable and only runs on the first proxy, being a SPOF. Clients need to talk to a specific port on the HAProxy to reach the CA.
1. Making the second proxy a CA-client to the first one is a bit cumbersome, I wish the installer would handle that for us
1. Pulp generates different repo metadata on different proxies, so clients need to stick to their backend or the'll need to re-fetch the metadata all the time
1. SELinux is permissive on the HAProxy box

### DIFF
--- a/pipelines/pipeline_katello_lbproxy.yml
+++ b/pipelines/pipeline_katello_lbproxy.yml
@@ -14,6 +14,9 @@
       haproxy01:
         box: centos7
         memory: 3072
+      client01:
+        box: centos7
+        memory: 3072
   roles:
     - forklift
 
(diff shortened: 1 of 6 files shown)


pipelines/pipeline_katello_lbproxy.yml
```diff
@@ -0,0 +1,238 @@
...
+        dest: /etc/hosts
+        line: "{{ hostvars['haproxy01'].ansible_default_ipv4.address }} {{ lbname }}"
+    - name: add puppet autosign entry
+      lineinfile:
+        path: /etc/puppetlabs/puppet/autosign.conf
+        line: "{{ ansible_fqdn }}"
+      delegate_to: proxy01
+    - name: fetch bootstrap.py
+      get_url:
+        url: https://raw.githubusercontent.com/Katello/katello-client-bootstrap/puppet-server/bootstrap.py
```
beav: 404, should probably be https://github.com/Katello/katello-client-bootstrap/blob/master/bootstrap.py

pipelines/pipeline_katello_lbproxy.yml
```diff
@@ -0,0 +1,238 @@
+- hosts: localhost
+  vars:
+    forklift_name: pipeline-katello-lbproxy
+    forklift_boxes:
+      katello01:
+        box: centos7
+        memory: 6144
```
beav: got OOM with 6144, retrying with 8192

pipelines/pipeline_katello_lbproxy.yml
evgeni: right, that's what you get when you use an own branch when testing, but publishing with the  patch merged and not checkig :)

pipelines/pipeline_katello_lbproxy.yml
evgeni: interesting, as 6144 is also what we use in our other pipelines. changed to 8192 (and lowered haproxy and client to 2048)

beav:
ACK, thanks @evgeni !

At some point it may be good to break some of the setup out into bats tests, and then also test the client (basically what was done by hand). I'll record this as a todo but it's not related to this PR.

PR #686, opened 2018-04-26
Add a google provider with examples

akofink:
I was recently given a large credit to use for a year, and this is what I've been using to make forklift work with GCE.

Note: Rather than specifying an image, Google uses the concept of image families to ensure users are using the most up to date image.

### DIFF
--- a/vagrant/boxes.d/00-centos.yaml
+++ b/vagrant/boxes.d/00-centos.yaml
@@ -2,6 +2,8 @@ boxes:
   centos6:
     box_name:   'centos/6'
     image_name: !ruby/regexp '/CentOS 6.*PV/'
+    google_options:
+      image_family: 'centos-6'
     pty:        true
     scenarios:
       - foreman
@@ -10,6 +12,8 @@ boxes:
   centos7:
     box_name:   'centos/7'
     image_name: !ruby/regexp '/CentOS 7.*PV/'
+    google_options:
+      image_family: 'centos-7'
     pty:        true
     scenarios:
       - foreman
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -85,6 +85,7 @@ module Forklift
         configure_virtualbox(machine, box, networks)
         configure_rackspace(machine, box)
         configure_openstack_provider(machine, box)
+        configure_google_provider(machine, box)
         configure_synced_folders(machine, box)
         configure_sshfs(config, box)
         configure_nfs(config, box)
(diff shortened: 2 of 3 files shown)


vagrant/boxes.d/99-local.yaml.example
```diff
@@ -59,3 +59,12 @@ centos7-dynflow-devel:
     variables:
       dynflow_devel_github_username: '<REPLACE ME>'
       dynflow_devel_github_fork_remote_name: 'origin'
+
+google-centos7-devel:
+  box: google-centos7 # always use this box_name for google
+  image_family: 'centos-7'
```
akofink: Rather than specifying an image, Google uses the concept of image families to ensure users are using the most up to date image in an image family.

ekohl:
Would a section in the README with the plugin installation instructions be a good idea?

akofink:
Other providers aren't documented there, so I don't think it'd be the right place (though I do need to update the Readme with the google_options). I did provide a link to the vagrant-google plugin which contains installation instructions. I'd rather let them document installation there and link to it.

vagrant/boxes.d/00-centos.yaml
```diff
@@ -19,3 +19,8 @@ boxes:
     box_name: centos7-atomic
     pty: true
     libvirt: http://cloud.centos.org/centos/7/atomic/images/CentOS-Atomic-Host-7-Vagrant-Libvirt.box
+
+  google-centos7:
+    box_name: 'google/gce'
+    google_options:
+      image_family: 'centos-7'
```
ehelms: I am not seeing why this box needs its own definition and name? Google is just a provider, so you should be able to specify this on the main `centos7` box.

akofink: The main centos7 box only supports 4 providers? https://app.vagrantup.com/centos/boxes/7

ekohl: Some provides do not use box names. openstack also uses an image name. You already override the box in the library so that can be ignored.

akofink: I'm pretty sure the google provider requires you to set box_name to 'google/gce' for any box and use image_name to set the os.

akofink: Oh, I guess in the past, we've just used overrides? Is that the right way to go?

ekohl: Isn't that why you have `override.vm.box = 'google/gce'` in `configure_google_provider`?

akofink:
I've tested this using centos7-devel and a plain centos6 box with GCE.

PR #689, opened 2018-05-04
Hard fail when foreman_installer_options is a string

ekohl:


akofink:
The Readme has some examples of strings with the devel box.

PR #708, opened 2018-05-08
Force destroy to ignore y/n prompt

akofink:
I honestly don't know when this broke, but I think recent versions of Vagrant ask you interactively.


ekohl:
Thanks!

PR #709, opened 2018-05-09
allow haproxy to bind any port in the lbproxy setup

evgeni:
using the haproxy_connect_any sebool


beav:
change is straighforward, APJ

evgeni:
@ekohl the role is included, at a different step: 
https://github.com/evgeni/forklift/blob/7e14ba59eb19447ba52fd0c593ccc2baaac5243b/pipelines/pipeline_katello_lbproxy.yml#L31

I just needed to override it once for the haproxy box before. I don't anymore.

PR #710, opened 2018-05-09
Update 3.4-3.5 upgrade pipeline for 3.5.2

akofink:
The rkerberos is a workaround and related to epel repos being broken right now for CentOS 7.4

The other changes come from the [Puppet 4-5 upgrade docs](http://projects.theforeman.org/projects/foreman/wiki/Upgrading_from_Puppet_4_to_5).

Thanks to @stbenjam for the help figuring this out.

### DIFF
--- a/pipelines/pipeline_katello_34_to_35_upgrade.yml
+++ b/pipelines/pipeline_katello_34_to_35_upgrade.yml
@@ -1,3 +1,4 @@
+# Bring up boxes
 - hosts: localhost
   vars:
     forklift_name: pipeline-katello-3.4-3.5
@@ -11,6 +12,7 @@
   roles:
     - forklift
 
+# Install Katello 3.4, Puppet 4
 - hosts:
   - pipeline-katello-3.4-3.5-centos7
   - pipeline-proxy-3.4-3.5-centos7
@@ -67,6 +69,7 @@
     - foreman_proxy_content
     - foreman_installer
 
+# Ensure bats passes on 3.4 to generate some data
 - hosts: pipeline-katello-3.4-3.5-centos7
   become: true
   vars:
@@ -76,9 +79,21 @@
   roles:
     - bats
 
+# Hack in case walrus is still on the machine
+- hosts:
+  - pipeline-katello-3.4-3.5-centos7
+  become: yes
+  tasks:
+    - name: 'Remove walrus package'
+      yum:
+        name: walrus
+        state: absent
+
+# Update to Katello 3.5
 - hosts: pipeline-katello-3.4-3.5-centos7
   become: true
   vars:
+    puppet_repositories_version: 4
     katello_repositories_version: 3.5
     foreman_repositories_version: 1.16
     foreman_installer_upgrade: True
@@ -96,6 +111,7 @@
 - hosts: pipeline-proxy-3.4-3.5-centos7
   become: true
   vars:
+    puppet_repositories_version: 4
     katello_repositories_version: 3.5
     foreman_repositories_version: 1.16
     foreman_installer_upgrade: True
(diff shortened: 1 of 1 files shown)


pipelines/pipeline_katello_34_to_35_upgrade.yml
```diff
@@ -87,6 +116,11 @@
     foreman_installer_scenario: katello
     foreman_installer_options_internal_use_only:
       - "--disable-system-checks"
+      - "--reset-puppet-server-version"
```
ekohl: Any reason you can't use `--upgrade-puppet` here (and below)?

pipelines/pipeline_katello_34_to_35_upgrade.yml
```diff
@@ -32,6 +32,17 @@
     - haveged
     - disable_firewall
 
+- hosts:
```
ekohl: https://github.com/theforeman/forklift/commit/9fd00b0b9c07acfa688ae243e05e1a2cc8c53497 should make this redundant but it's included. Any idea why?

akofink: I can try without, but I was hitting an error (maybe 2 days ago before the PR was merged ;)

pipelines/pipeline_katello_34_to_35_upgrade.yml
akofink: I was following [this wiki](http://projects.theforeman.org/projects/foreman/wiki/Upgrading_from_Puppet_4_to_5#Prepare-And-Run-The-Installer), which doesn't mention `--upgrade-puppet` at all. Would you mind updating the wiki to the "@ekohl recommended way"? ;) In the meantime, I'll go ahead and test it out.

akofink: @ekohl I tried it, and apparently it's unsupported during an upgrade:

```
Concurrent use of --upgrade and --upgrade-puppet is not supported. Please run --upgrade first, then --upgrade-puppet.
```

So I could leave it as is and have only one installer run for the upgrade (and puppet upgrade) or have two installer runs just to use the 'more convenient' option. That doesn't sound more convenient. Maybe I'm missing something?

akofink: And, here's what I get when trying to use the installer with `--upgrade-puppet` before re-running it with `--upgrade`:

```
TASK [foreman_installer : Run installer] *****************************************************************************************************************************************************************************************************
Thursday 10 May 2018  12:01:49 -0400 (0:00:00.061)       0:01:28.680 **********
fatal: [pipeline-katello-3.4-3.5-centos7]: FAILED! => {
    "changed": true,
    "cmd": "foreman-installer -v --scenario \"katello\"  --disable-system-checks --upgrade-puppet",
    "delta": "0:00:01.087437",
    "end": "2018-05-10 16:01:51.006276",
    "rc": 1,
    "start": "2018-05-10 16:01:49.918839"
}

STDOUT:

Puppet already installed and upgraded. Skipping.


MSG:

non-zero return code

fatal: [pipeline-proxy-3.4-3.5-centos7]: FAILED! => {
    "changed": true,
    "cmd": "foreman-installer -v --scenario \"katello\"  --disable-system-checks --upgrade-puppet",
    "delta": "0:00:02.131133",
    "end": "2018-05-10 16:01:52.079101",
    "rc": 27,
    "start": "2018-05-10 16:01:49.947968"
}

STDOUT:

[ INFO 2018-05-10 16:01:52 verbose] Executing hooks in group pre_migrations
[ INFO 2018-05-10 16:01:52 verbose] All hooks in group pre_migrations finished
[ERROR 2018-05-10 16:01:52 verbose] You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Use --force to override. You can use --compare-scenarios to see the differences
[ INFO 2018-05-10 16:01:52 verbose] Log was be written to /var/log/foreman-installer/katello.log
[ INFO 2018-05-10 16:01:52 verbose] Logs flushed


MSG:

non-zero return code
```

Not sure why it thinks this is a different scenario...

akofink:
Updated to be a bit more methodical about what exactly we're updating at any given step. Now, we install 3.4/P4, run bats, upgrade to 3.5/P4, run bats, then upgrade to P5, and run bats again.

pipelines/pipeline_katello_34_to_35_upgrade.yml
```diff
@@ -110,6 +126,69 @@
...
+  vars:
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--reset-puppet-server-jvm-extra-args"
+      - "--reset-puppet-server-puppetserver-version"
+      - "--reset-puppet-server-puppetserver-metrics"
+      - "--reset-puppet-server-puppetserver-experimental"
+  roles:
+    - foreman_installer
```
zjhuntin: Is this part just to upgrade puppet?

akofink: No, this part resets some Foreman variables about the puppet server. I think it discovers the Puppet version and stuff if they're not set. The actual puppet upgrade happens on [line 160](https://github.com/theforeman/forklift/pull/710/files#diff-8b9d90c21ca1de611ed04bd30b9685d1R160), so by this point, if you ran `puppet --version`, it'd say `5.5.1`.

akofink: I'm getting these steps from [this wiki](http://projects.theforeman.org/projects/foreman/wiki/Upgrading_from_Puppet_4_to_5#Prepare-And-Run-The-Installer).

akofink:
@ekohl This one's ready for merge.

akofink:
`--upgrade-puppet` is meant for Puppet 3 to Puppet 4 upgrade, not Puppet 4 to 5, as mentioned in previous comments.

akofink:
Apparently the help text is incorrect, and --upgrade-puppet *should* work for P4-P5, though @ekohl mentioned there may be a bug, as --upgrade-puppet doesn't reset all the variables that the wiki lists.

ehelms:
@akofink This needs a rebase

akofink:
Ready for merge here.

ehelms:
Still shows needing a rebase @akofink

akofink:
Really? It shows "Merge pull request" for me.

ehelms:
@akofink What does it show if you select 'Rebase and merge' ?

akofink:
We should disable 'Create a merge commit' on this and all other repos under our control.

PR #711, opened 2018-05-16
WIP: Run development environment using Docker and Ansible

ripcurld:
cc @ehelms @ohadlevy @carbonin @ekohl 

Would like to get your feedback. This is to run a simple development environment using Docker.

Thanks.


containers/roles/development/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+    name: "{{image_name}}"
+    tag: "{{image_tag}}"
+    path: images/develop/foreman
+    buildargs:
+      hostUID: "{{user_id}}"
+  tags:
+    - build
+
+# reset the bundle configs to download gems as specified
+- name: reset bundle configurations in {{foreman_src}}/.bunlde/config
```
ekohl: s/bunlde/bundle

containers/README.md
```diff
@@ -1,3 +1,108 @@
+# Foreman on Docker
+
+The following use case introduces a way for developers to run `Foreman` using `Docker`.
+
+## Prerequiesets
```
ekohl: Prerequisites

containers/roles/development/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+    cleanup: yes
+    detach: false
+    name: "database_setup"
+    links: "{{None if adapter == 'sqlite3' else foreman_database_name + ':' + db_host}}"
+    image: "{{image_name}}:{{image_tag}}"
+    command: "{{item}}"
+  loop:
+    - bundle exec rake db:migrate
+    - bundle exec rake db:seed
+    - bundle exec rake permissions:reset
```
ekohl: Why is this needed? Seeding should already set up a user + password. You also probably want to make sure you set the env vars to get an initial installation. See https://github.com/theforeman/puppet-foreman/blob/127925156119ec4a76bbb33537a2f1168da095ba/manifests/database.pp#L17-L25 for inspiration.

containers/roles/development/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+
+# reset the bundle configs to download gems as specified
+- name: reset bundle configurations in {{foreman_src}}/.bunlde/config
+  file:
+    state: absent
+    path: "{{foreman_src}}/.bundle/config"
+  tags:
+    - install
+
+- name: install npm and gem packages
```
carbonin: Why not include this step in the build process? What is the use case for separate build and install steps?

carbonin:
I agree with @ekohl on this. It seems a bit complex. Especially seeing as we're not utilizing much of the isolation that containers provide (all the prerequisites are installed and built on the host rather than the image).

Is there a particular development use case that this simplifies over running the app locally?

carbonin:
Also, I think multiple source-based images would be confusing. This PR and https://github.com/theforeman/forklift/pull/703 feel like they are trying to accomplish the same thing.

akofink:
I think forklift has made us a bit spoiled, but can you make it work with a single command? ;)

ripcurld:
@akofink thanks for your review. You can run this as one command `ansible-playbook development.yml` and it will run perfectly.

I will add that part to the `README.md` file, though.

Thanks.

ekohl:
It's a lot of steps for what could essentially be:
```bash
sudo dnf -y install ruby{,-devel,gems} rubygem-{nokogiri,bundler,unf_ext,rdoc} \
                   libvirt-devel sqlite-devel libxml2-devel libcurl-devel systemd-devel \
                   redhat-rpm-config nodejs git gcc-c++ make hostname
git clone https://github.com/theforeman/foreman
cd foreman
cp config/settings.yaml.example config/settings.yaml
cp config/database.yml.example config/database.yml
bundle install --without mysql mysql2 postgresql
npm install
bundle exec rake db:migrate db:seed
bundle exec foreman start
```

Perhaps my aversion also comes from the fact that I think developer environments *should* differ. IMHO we want a flexible and robust tool that works on my platforms. Since we deploy to many operating systems (RHEL/CentOS, Debian, Ubuntu, BSD) I want this to be reflected in developer setups. Always working in the same development environment leads to a mono-culture and less robust systems.

I also expect Foreman developers to be comfortable with managing their own systems. The target audience is sysadmins. You have to know your audience and the tools they use.

More of an implementation detail, but I do feel that the current addition to the containers directory brings even more confusion to an already very widely spread repository with a huge scope. Perhaps this should live in its own directory as a sort of standalone tool. Perhaps we need `containers/openshift` and `containers/docker` or something.

Given those factors and how thinly spread I already am I'd like to distance myself from this. I'm not going to block any merges, but I simply don't have the time to maintain it.

ehelms:
@ripcurld Thanks for the contribution. I think @ekohl has summed up some of my concerns with this but I'll express a few others that I think are key:

 1) This only works for the simple case of just Foreman. But our deployments these days are quite complex to the point of needing to orchestrate multiple containers. I see no value in inventing our own orchestration. 
 2) If our target is to run on Kubernetes/Openshift we should stick to it as a platform. I think we are safe to assume that any version of the stack should run there and not re-invent the wheel. How would you handle this when all 13 or so services need to be run for a development environment?
 3) A goal of containers is have development and production be as minimally distant from one another as possible. I think this increases the divide.

I'd consider the following ideas to pivot towards a development environment:

 * Looking into using `oc rsync` to sync local source into a running Openshift container running with `RAILS_ENV=development` environment variable set
 * Consider how `https://github.com/theforeman/forklift/pull/703` could be used

PR #716, opened 2018-05-25
Move qpid to Dockerfile

ehelms:


ehelms:
[test containers]

carbonin:
Was there supposed to be a qpid Dockerfile in here also? Or are you pulling from a different image?

ehelms:
@carbonin Dunno where that got off to put good catch, I put in the missing Dockerfile

carbonin:
Can you also remove the `qpid` role? (Same concern as https://github.com/theforeman/forklift/pull/717#issuecomment-392817713)

ehelms:
@carbonin Check and done.

PR #720, opened 2018-05-29
Fixes #695: Move Pulp to Dockerfiles

### LINKED ISSUE #695: Move pulp, pulp-celerybeat, pulp-resource-manager, pulp-worker  to Dockerfile

ehelms:


ehelms:
@carbonin @ohadlevy Care to review? This is a bigger one but I'm down to just three PRs to get rid of a-c.

containers/images/pulp-celerybeat/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM quay.io/foreman/pulp:latest
+
+COPY container-assets/entrypoint.sh /usr/bin/entrypoint.sh
+COPY container-assets/start_pulp_celerybeat.sh /usr/bin/start_pulp_celerybeat.sh
+
+USER apache
+
+CMD ["/usr/bin/start_pulp_celerybeat.sh"]
+ENTRYPOINT ["/usr/bin/entrypoint.sh"]
```
carbonin: Are any of these process based where they would need a better init here?

containers/images/pulp-resource-manager/container-assets/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+/usr/bin/wait_on_database_migration.py
```
ohadlevy: what are the chances of a race condition if you run multiple pods?

containers/images/pulp-streamer/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+    yum -y install epel-release && \
+		yum -y install python-pulp-streamer && \
+		yum clean all
+
+COPY container-assets/streamer.conf /etc/pulp/streamer.conf
+COPY container-assets/entrypoint.sh /usr/bin/entrypoint.sh
+COPY container-assets/start_pulp_streamer.sh /usr/bin/start_pulp_streamer.sh
+
+ADD https://raw.githubusercontent.com/ehelms/pulp/127d099833ec8d3c757db2b28b5fd14a2b6d7077/streamer/usr/share/pulp/wsgi/streamer.tac /usr/share/pulp/wsgi/streamer.tac
```
ohadlevy: why fetch from github?

containers/images/pulp-celerybeat/Dockerfile
ehelms: Can you expand for me on 'better init' ?

containers/images/pulp-resource-manager/container-assets/entrypoint.sh
ehelms: Race condition for what to happen? We run multiple pulp-workers which has this same wait without issue.

containers/images/pulp-streamer/Dockerfile
ehelms: Cause my patch has not been accepted into Pulp yet :/ I do have a PR open for it though.

containers/images/pulp-celerybeat/Dockerfile
carbonin: [dumb-init](https://github.com/Yelp/dumb-init) or [tini](https://github.com/krallin/tini) specifically something that handles signals and zombie reaping properly.

carbonin: I'm just not sure which of these containers would have child processes which exit and would need to  be `wait()`ed on.

carbonin: Right, if this is just a translation then this is fine for now.

I'll comment on the issue around our experience with dumb-init.

containers/roles/squid/tasks/main.yml
```diff
@@ -1,11 +1,4 @@
 ---
-- name: 'Install squid'
```
carbonin: Can this whole file be removed?

ehelms:
@ohadlevy Final review is yours

PR #721, opened 2018-05-29
Remove unbuilt containers from container.yml

ehelms:


carbonin:
Why is the proxy-register container also being removed here?
Is that sharing an image with a different service?

ehelms:
The foreman-proxy-register image is no longer used. Instead a lifecycle hook is being used.

PR #722, opened 2018-05-29
Fixes #699 - Move content-server to Dockerfile

### LINKED ISSUE #699: Move content-server to a Dockerfile

ehelms:


containers/images/content-server/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM quay.io/foreman/pulp
+
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+    yum -y install epel-release && \
+    yum -y install python-pulp-streamer && \
```
ohadlevy: you can also rm -rf /var/cache/yum

containers/images/content-server/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN ln -s /etc/pki/pulp/certs/pulp.crt /etc/pki/tls/certs/localhost.crt
+RUN ln -s /etc/pki/pulp/private/pulp.key /etc/pki/tls/private/localhost.key
+
+RUN chown apache:apache /var/lib/pulp
+
+RUN sed -i 's/127.0.0.1/squid/g' /etc/httpd/conf.d/pulp_streamer.conf
+
+COPY container-assets/content_wsgi.conf /etc/httpd/conf.d/content_wsgi.conf
+
+CMD ["httpd", "-D", "FOREGROUND"]
```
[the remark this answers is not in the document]
carbonin: IIRC httpd does the permission and user change when you start it up, so it should run as root.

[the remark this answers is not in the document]
ohadlevy: right, its needed in order to bind to port 80/443, thanks.

containers/images/content-server/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM quay.io/foreman/pulp
+
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+    yum -y install epel-release && \
+    yum -y install python-pulp-streamer && \
+		yum clean all
+
+RUN rm -f /etc/httpd/conf.d/pulp.conf /etc/httpd/conf.d/pulp_puppet.conf
```
ripcurld: Each Dockerfile instruction like `RUN` creates a new layer and from what I understood - avoiding adding more layers to the image is a best practice.

Thus, maybe you can merge these next lines into:

```
RUN rm -f /etc/httpd/conf.d/pulp.conf /etc/httpd/conf.d/pulp_puppet.conf /etc/pki/tls/certs/localhost.crt /etc/pki/tls/private/localhost.key && \
    ln -s /var/lib/pulp/published /var/www/pub && \
    ln -s /etc/pki/pulp/certs/pulp.crt /etc/pki/tls/certs/localhost.crt && \
    ln -s /etc/pki/pulp/private/pulp.key /etc/pki/tls/private/localhost.key && \
    chown apache:apache /var/lib/pulp && \
    sed -i 's/127.0.0.1/squid/g' /etc/httpd/conf.d/pulp_streamer.conf
```
Unless there are caching or other concerns that made you decide to do that.

containers/images/content-server/Dockerfile
ehelms: I tested that and it didn't seem to do anything more than `yum clean all`.

containers/images/content-server/Dockerfile
ehelms: It does reduce layers, however, I think it makes readability worse. I think we can explore image sizes across the board after the initial round to weigh size reduction vs. understandbility.

ehelms:
@ohadlevy @carbonin please re-review at your convience

containers/container.yml
```diff
@@ -76,10 +76,6 @@ services:
     from: centos:7
     roles:
       - squid
-  content-server:
-    from: foreman-pulp-base:latest
-    roles:
-      - content-server
```
carbonin: Same as before, can we remove this role?

ehelms: With fire -- done

carbonin: :fire: :fire: :scissors: :+1:

PR #723, opened 2018-05-29
Fixes #694: Move foreman-proxy to Dockerfile

### LINKED ISSUE #694: Move foreman-proxy to Dockerfile

ehelms:


containers/images/foreman-proxy/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+COPY container-assets/logs.yml /etc/foreman-proxy/settings.d/logs.yml
+COPY container-assets/pulp.yml /etc/foreman-proxy/settings.d/pulp.yml
+COPY container-assets/puppet.yml /etc/foreman-proxy/settings.d/puppet.yml
+COPY container-assets/puppet_proxy_puppet_api.yml /etc/foreman-proxy/settings.d/puppet_proxy_puppet_api.yml
+COPY container-assets/settings.yml /etc/foreman-proxy/settings.yml
+
+RUN ln -s /etc/foreman-proxy/settings.d /usr/share/foreman-proxy/config/settings.d
+
+CMD ["/usr/share/foreman-proxy/bin/smart-proxy"]
```
ripcurld: Why not using `ENTRYPOINT` here?

ehelms: I thought as I understood things, command should be what your container actually runs and it is more easily overriden if we need to for example just get inside the container for inspection? I find them both kinda nuanced to be honest.

ripcurld: > it is more easily overriden if we need to for example just get inside the container for inspection

IIRC to get inside a container, `docker exec` is used and from my experience it doesn't use `ENTRYPOINT` or `CMD`.

Although, there is a slight difference between them, it does change the user experience.
So I guess if `smart-proxy` doesn't get arguments then `CMD` is good enough.

PR #724, opened 2018-05-29
Fixes #693 - Move foreman and dynflow to Dockerfiles

### LINKED ISSUE #693: Move foreman & dynflow containers to Dockerfiles

ehelms:


containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    yum -y install tfm-rubygem-katello tfm-rubygem-foreman_ansible tfm-rubygem-foreman_discovery tfm-rubygem-foreman_hooks && \
+    yum -y install tfm-rubygem-foreman_memcache tfm-rubygem-foreman_openscap tfm-rubygem-foreman_remote_execution tfm-rubygem-foreman_templates && \
+    yum clean all
+
+COPY container-assets/settings.yaml /etc/foreman/settings.yaml
+COPY container-assets/katello.yaml /etc/foreman/plugins/katello.yaml
+COPY container-assets/foreman_memcache.yaml /etc/foreman/plugins/foreman_memcache.yaml
+
+# Generate apipie:cache:index
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
```
ohadlevy: its already defined in line 3

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+    yum -y install epel-release && \
+    yum -y install http://yum.theforeman.org/releases/nightly/el7/x86_64/foreman-release.rpm && \
+    yum -y install https://fedorapeople.org/groups/katello/releases/yum/nightly/katello/el7/x86_64/katello-repos-latest.rpm && \
+    yum -y install foreman-release-scl && \
+    yum -y install qpid-tools && \
+    yum -y install foreman tfm-rubygem-puma && \
+    yum -y install tfm-rubygem-katello tfm-rubygem-foreman_ansible tfm-rubygem-foreman_discovery tfm-rubygem-foreman_hooks && \
+    yum -y install tfm-rubygem-foreman_memcache tfm-rubygem-foreman_openscap tfm-rubygem-foreman_remote_execution tfm-rubygem-foreman_templates && \
```
ohadlevy: you can squash multiple yum executions in to a fewer ones, it would make image creation faster. (e.g. )
```
yum -y install qpid-tools foreman tfm-rubygem-puma ....
```
Probably better to expose those as an ARG so people can add additional packages if they rebuild it (with an additional plugin?)

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+COPY container-assets/foreman_memcache.yaml /etc/foreman/plugins/foreman_memcache.yaml
+
+# Generate apipie:cache:index
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+    yum -y install foreman-sqlite && \
+		yum clean all
+
+COPY container-assets/build/database.yml /etc/foreman/database.yml
+
+RUN foreman-rake db:create
```
ohadlevy: you can run it all in one line (db:create db:migrate apipie:cache:index ...)

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN foreman-rake db:migrate
+RUN foreman-rake apipie:cache:index
```
ohadlevy: does it make sense to limit the language support for apipie (so its both faster and smaller by default?)

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
```
ohadlevy: duplicated from lines 3

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    yum -y remove foreman-sqlite && \
+		yum clean all
```
ohadlevy: you should really clear the cache too by rm it

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+COPY container-assets/settings.yaml /etc/foreman/settings.yaml
+COPY container-assets/katello.yaml /etc/foreman/plugins/katello.yaml
+COPY container-assets/foreman_memcache.yaml /etc/foreman/plugins/foreman_memcache.yaml
+
+# Generate apipie:cache:index
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+    yum -y install foreman-sqlite && \
+		yum clean all
+
+COPY container-assets/build/database.yml /etc/foreman/database.yml
```
ohadlevy: instead of coping a temp database.yml file, you can use in memory sqlite, e.g.
```sh
DATABASE_URL=sqlite3::memory: bundle exec rake ${TASKS}
```
(abased on https://github.com/theforeman/forklift/pull/703/files#diff-e73050ae9f2baca99ce3cc574adee190R21)

containers/images/foreman/Dockerfile
ehelms: Fair, I'll update some of them, but I don't want to lose the readability and the implications that each is a unique step to be done in a particular order. Sometimes that is my gripe with these "ugly" ways you are forced to write Dockerfiles rather than the build tooling being able to optimize.

As for making the package list an ARG, I'd like to translate as much as possible things directly first. To reduce the amount of churn in conversion. Then come back through and add in features.

containers/images/foreman/Dockerfile
ehelms: I have seen issues with running create and migrate in the same step.

containers/images/foreman/Dockerfile
ehelms: And re-enable all languages in the future? This seems like a project level question.

containers/images/foreman/Dockerfile
ehelms: When I deleted the cache it had no effect on image size in my local testing.

ehelms:
Addressed all the comments and made a few additional changes to make things a bit more efficient.

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    yum -y install foreman tfm-rubygem-puma \
+    							 tfm-rubygem-katello tfm-rubygem-foreman_ansible tfm-rubygem-foreman_discovery tfm-rubygem-foreman_hooks \
+    							 tfm-rubygem-foreman_openscap tfm-rubygem-foreman_remote_execution tfm-rubygem-foreman_templates && \
+    yum clean all
+
+COPY container-assets/settings.yaml /etc/foreman/settings.yaml
+COPY container-assets/katello.yaml /etc/foreman/plugins/katello.yaml
+
+# Generate apipie:cache:index
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
```
ohadlevy: ```tsflags=nodocs``` should already be in yum.conf at this stage.

containers/images/foreman/Dockerfile
ohadlevy: this is what yum prints:
```
Maybe you want: rm -rf /var/cache/yum, to also free up space taken by orphaned data from disabled or removed repos
```

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    yum -y install foreman-sqlite && \
+		yum clean all
+
+RUN DATABASE_URL=sqlite3::memory: foreman-rake db:create db:migrate apipie:cache:index
+
+RUN yum -y remove foreman-sqlite && \
```
ohadlevy: note this doesnt remove all dependent packages that get installed (e.g. tfm-ror51-rubygem-sqlite3) either add that as well or consider using something like yum history (https://github.com/theforeman/forklift/pull/703/files#diff-e73050ae9f2baca99ce3cc574adee190R27)

ehelms:
Addressed comments

containers/images/dynflow/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM quay.io/foreman/foreman:latest
+
+COPY container-assets/entrypoint.sh /usr/bin/entrypoint.sh
+COPY container-assets/wait_on_migrations.py /usr/bin/wait_on_migrations.py
+
+ENTRYPOINT ["/usr/bin/entrypoint.sh"]
+CMD ["/usr/sbin/dynflowd", "--foreman-root=/usr/share/foreman", "run"]
```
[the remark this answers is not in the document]
ehelms: There is a USER in the Foreman base image which should cover it I think?

ehelms:
@carbonin @ohadlevy Since this is the last set to go in, I used it to completely scrub references to ansible-container, updated the documentation and tooling playbooks for pure Dockerfile image builds. There are a few updates to other Dockerfiles in here now that I found while testing the tooling playbooks.

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM centos:7
+
+RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
+    yum -y install epel-release \
+    							 http://yum.theforeman.org/releases/nightly/el7/x86_64/foreman-release.rpm \
+    							 https://fedorapeople.org/groups/katello/releases/yum/nightly/katello/el7/x86_64/katello-repos-latest.rpm && \
+    yum -y install foreman-release-scl && \
```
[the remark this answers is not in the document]
ehelms: The `foreman-release-scl` is provided by the the Foreman repository configured in the line above. This RPM itself configures a set of repositories so it needs to be a separate command before installing RPMs.

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+		yum clean all
+
+RUN DATABASE_URL=sqlite3::memory: foreman-rake db:create db:migrate apipie:cache:index
+
+RUN yum -y remove foreman-sqlite tfm-ror51-rubygem-sqlite3 && \
+		yum clean all
+# end generate apipie:cache:index
+
+RUN yum -y install tfm-rubygem-foreman_memcache && \
+		yum clean all
```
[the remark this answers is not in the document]
ehelms: Yes and no. I think some work needs to be done on `foreman_memcache` first. If this is included the build takes a lot longer and spits out a lot more debug info than is needed because it tries to connect to a memcache server that does not exist.

ehelms:
This has a few changes across images to ensure everything is working properly.

ehelms:
@ohadlevy @carbonin OK this is fully ready for a review. There are a lot of changes in here, some mistakes in other images found but you'll notice it is green now based on the changes here in.

containers/images/content-server/Dockerfile
```diff
@@ -13,9 +13,11 @@ RUN ln -s /etc/pki/pulp/certs/pulp.crt /etc/pki/tls/certs/localhost.crt
...
 
 RUN chown apache:apache /var/lib/pulp
+RUN mkdir /var/www/html/pub && chown apache:apache -R /var/www/html
 
 RUN sed -i 's/127.0.0.1/squid/g' /etc/httpd/conf.d/pulp_streamer.conf
 
 COPY container-assets/content_wsgi.conf /etc/httpd/conf.d/content_wsgi.conf
 
 CMD ["httpd", "-D", "FOREGROUND"]
+ENTRYPOINT []
```
carbonin: Can you move `ENTRYPOINT` above `CMD`?

I think it reads better as technically `CMD` is supposed to be the arguments for `ENTRYPOINT`

carbonin: Also, is there any way to make `ENTRYPOINT` more generic in the pulp image rather than having to blank it here?

containers/images/foreman/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+RUN yum -y install tfm-rubygem-foreman_memcache && \
+    yum clean all
+
+COPY container-assets/foreman_memcache.yaml /etc/foreman/plugins/foreman_memcache.yaml
+COPY container-assets/database.yml /etc/foreman/database.yml
+COPY container-assets/entrypoint.sh /usr/bin/entrypoint.sh
+COPY container-assets/start_foreman.sh /usr/bin/start_foreman.sh
+COPY container-assets/wait_on_postgres.py /usr/bin/wait_on_postgres.py
+COPY container-assets/create_qpid_queue.sh /usr/bin/create_qpid_queue.sh
+#COPY container-assets/puma.rb /usr/share/foreman/bundler.d/puma.rb
```
carbonin: Does this need to stay here?

containers/images/pulp-resource-manager/Dockerfile
```diff
@@ -4,9 +4,9 @@ RUN mkdir /var/lib/pulp/celery
...
 
 COPY container-assets/entrypoint.sh /usr/bin/entrypoint.sh
-COPY container-assets/start_celery_beat.sh /usr/bin/start_celery_beat.sh
+COPY container-assets/start_pulp_resource_manager.sh /usr/bin/start_pulp_resource_manager.sh
 
 USER apache
 
-CMD ["/usr/bin/start_celery_beat.sh"]
+CMD ["/usr/bin/start_pulp_resource_manager.sh"]
 ENTRYPOINT ["/usr/bin/entrypoint.sh"]
```
carbonin: Same, can we keep `ENTRYPOINT` above `CMD`?

containers/images/content-server/Dockerfile
ehelms: The tough part is Pulp's entrypoint is doing things like database operations before handing off to the final run command. I could move all that into the CMD itself but I had been following most examples I found that put things like database operations into the entrypoint.

ehelms:
Updated @carbonin -- I found a few places that entrypoint was after CMD and updated all of them

ehelms:
[test containers]

PR #725, opened 2018-05-30
Fix the roles path in ansible.cfg

rahulbajaj0509:


ansible.cfg
```diff
@@ -9,3 +9,5 @@ retry_files_save_path = $HOME/.ansible/retry-files
 remote_user = vagrant
 host_key_checking = False
 force_color = 1
+roles_path= ./roles
```
ekohl: On line 7 we already have a `roles_path`. Doesn't that work for you? It'd be good to have just 1 definition.

rahulbajaj0509:
When tried to run the `vagrant up centos7-devel` command on Ubuntu, it gives me an error like:
```

centos7-devel: Running ansible-playbook...
Using /home/rahul/github/forklift/ansible.cfg as config file
ERROR! the role 'etc_hosts' was not found in /home/rahul/github/forklift/playbooks/roles:/home/rahul/github/forklift/$PWD/playbooks/galaxy_roles:/home/rahul/github/forklift/$PWD/user_playbooks/roles:/home/rahul/github/forklift/$PWD/roles:/home/rahul/github/forklift/playbooks

The error appears to have been in '/home/rahul/github/forklift/playbooks/devel.yml': line 14, column 7, but may

  roles:
    - etc_hosts
      ^ here

```

This patch makes sure, that the roles path is identical to Ansible.

ansible.cfg
rahulbajaj0509: @ekohl if you see the error in the comment, ansible is not able to resolve $PWD variable in Ubuntu, although this works completely fine in fedora. Waiting for the ansible folks reply on this one.

the path being resolved is like : `/home/rabajaj/forklift/$PWD/roles`, trying to figure out why $PWD is being intact here in the path.

ehelms:
@ekohl Final review on you

PR #727, opened 2018-05-31
Split libvirt tftp configuration into another task

zjhuntin:


playbooks/provision_test_box.yml
```diff
@@ -0,0 +1,6 @@
+---
+- hosts: all
+  become: true
+  roles:
+    - role: libvirt_install
+    - role: vagrant_install
```
ehelms: I think the vagrant install should be it own PR.

roles/libvirt_configure/tasks/main.yml
```diff
@@ -63,9 +32,6 @@
 - name: 'copy syslinux files to /var/lib/tftpboot'
   shell: cp /usr/share/syslinux/{pxelinux.0,menu.c32,chain.c32} /var/lib/tftpboot
 
-- name: 'chown /var/lib/tftpboot for foreman-proxy user'
```
ehelms: Why remove this?

ehelms:
The naming of the two roles does not leave me with a sense of obvious as to what they do. I would think of it more like the following:

Option 1:

 * libvirt role
 * variable for libvirt_tftp that configures the tftp part optionally

Option 2:

 * libvirt role
 * libvirt_tftp role

roles/libvirt_configure/tasks/main.yml
zjhuntin: Wasn't meaning to, I removed it temporarily to get it working on another machine. I'll replace it.

playbooks/katello_provisioning.yml
```diff
@@ -1,6 +1,8 @@
 - hosts: all
   become: true
+  vars:
+    tftp: true
```
ehelms: Best practice is to scope variables based on role. For example, this would become `libvirt_tftp`. Also, we should set a default in defaults/main.yml

zjhuntin:
@ehelms Updated!

ehelms:
@zjhuntin Needs a rebase please

zjhuntin:
@ehelms Rebased

PR #728, opened 2018-06-01
Add vagrant role

zjhuntin:


roles/vagrant/default/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+vagrant_libvirt: false
```
ehelms: I tend to prefer this true by default since its the most common case.

roles/vagrant/tasks/vagrant_libvirt.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: 'install vagrant-libvirt requirements'
+  yum:
+    name: "{{item}}"
```
ehelms: Spaces around the argument

PR #730, opened 2018-06-05
Katello 3.7 pipeline bits

jturel:
This appears to be working until bats tests fail

### DIFF
--- a/pipelines/pipeline_katello_36_to_37_upgrade.yml
+++ b/pipelines/pipeline_katello_36_to_37_upgrade.yml
@@ -1,0 +1,156 @@
+- hosts: localhost
+  vars:
+    forklift_name: pipeline-katello-3.6-3.7
+    forklift_boxes:
+      pipeline-katello-3.6-3.7-centos7:
+        box: centos7
+        memory: 6144
+      pipeline-proxy-3.6-3.7-centos7:
+        box: centos7
+        memory: 3072
+  roles:
+    - forklift
+
+- hosts:
+  - pipeline-katello-3.6-3.7-centos7
+  - pipeline-proxy-3.6-3.7-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 4
+    katello_repositories_version: 3.6
+    foreman_repositories_version: 1.17
+    foreman_installer_skip_installer: true
+  roles:
+    - umask
+    - selinux
+    - etc_hosts
+    - epel_repositories
+    - puppet_repositories
+    - foreman_repositories
+    - katello_repositories
+    - update_os_packages
+    - haveged
+    - disable_firewall
+
+- hosts: pipeline-katello-3.6-3.7-centos7
+  become: yes
+  vars:
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--foreman-admin-password {{ foreman_installer_admin_password }}"
+      - "--disable-system-checks"
+    foreman_installer_additional_packages:
+      - katello
+  roles:
+    - foreman_installer
+
+- hosts: pipeline-proxy-3.6-3.7-centos7
+  become: yes
+  vars:
+    foreman_proxy_content_server: pipeline-katello-3.6-3.7-centos7
+    foreman_installer_scenario: foreman-proxy-content
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--foreman-proxy-trusted-hosts {{ server_fqdn.stdout }}"
+      - "--foreman-proxy-trusted-hosts {{ ansible_nodename }}"
+      - "--foreman-proxy-foreman-base-url https://{{ server_fqdn.stdout }}"
+      - "--foreman-proxy-register-in-foreman true"
+      - "--foreman-proxy-oauth-consumer-key {{ oauth_consumer_key.stdout }}"
+      - "--foreman-proxy-oauth-consumer-secret {{ oauth_consumer_secret.stdout }}"
+      - "--foreman-proxy-content-certs-tar {{ foreman_proxy_content_certs_tar }}"
+      - "--foreman-proxy-content-parent-fqdn {{ server_fqdn.stdout }}"
+      - "--puppet-server-foreman-url https://{{ server_fqdn.stdout }}"
+    foreman_installer_additional_packages:
+      - foreman-installer-katello
+  roles:
+    - foreman_proxy_content
+    - foreman_installer
+
+- hosts: pipeline-katello-3.6-3.7-centos7
+  become: true
+  vars:
+    bats_tests:
+      - fb-test-katello.bats
+      - fb-content-katello.bats
+  roles:
+    - bats
+
+- hosts: pipeline-katello-3.6-3.7-centos7
+  become: true
+  vars:
+    katello_repositories_version: 3.7
+    foreman_repositories_version: 1.18
+    foreman_installer_upgrade: True
+    foreman_repositories_environment: staging
+    katello_repositories_environment: staging
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+  roles:
+    - foreman_repositories
+    - katello_repositories
+    - foreman_installer
+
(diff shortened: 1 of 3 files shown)


pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -0,0 +1,107 @@
...
+        memory: 3072
+  roles:
+    - forklift
+
+- hosts:
+  - pipeline-katello-3.6-3.7-centos7
+  - pipeline-proxy-3.6-3.7-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 5555
```
akofink: This should be 4 or 5, depending if you want to test the puppet upgrade.

pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -0,0 +1,107 @@
...
+  become: true
+  vars:
+    katello_repositories_version: 3.7
+    foreman_repositories_version: 1.18
+    foreman_installer_upgrade: True
+    foreman_repositories_environment: staging
+    katello_repositories_environment: staging
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--upgrade-puppet"
```
akofink: Looks like you're upgrading puppet? So you might want to try to explicitly test both Katello versions with both Puppet versions, similar to the [3.5 upgrade pipeline](https://github.com/theforeman/forklift/pull/710/files) (not yet merged). Just note that --upgrade-puppet is broken in 3.5, but it should work for you.

ekohl: I thought we removed `--upgrade-puppet` in 3.7 and included it in `--upgrade`

jturel: Thanks. I removed `--upgrade-puppet`

akofink: @ekohl Is there now no way to upgrade Katello but stay on P4?

pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -0,0 +1,106 @@
...
+        memory: 3072
+  roles:
+    - forklift
+
+- hosts:
+  - pipeline-katello-3.6-3.7-centos7
+  - pipeline-proxy-3.6-3.7-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 5
```
akofink: I'd encourage you to test the puppet upgrade rather than starting with P5.

pipelines/pipeline_katello_36_to_37_upgrade.yml
ekohl: Manage your yum repositories so only P4 RPMs are available or pin them otherwise. We've modified `--upgrade` to just ensure we work with the currently installed Puppet version.

pipelines/pipeline_katello_36_to_37_upgrade.yml
jturel: take a look ... I think I've approximated what you did with 3.5

pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -0,0 +1,168 @@
...
+    katello_repositories_version: 3.7
+    foreman_repositories_version: 1.18
+    foreman_installer_upgrade: True
+    foreman_repositories_environment: staging
+    katello_repositories_environment: staging
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+  roles:
+    - puppet_repositories
```
akofink: You technically don't need this here, but it doesn't hurt.

pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -0,0 +1,168 @@
...
+    puppet_repositories_version: 5
+  roles:
+    - puppet_repositories
+
+- hosts:
+  - pipeline-katello-3.6-3.7-centos7
+  - pipeline-proxy-3.6-3.7-centos7
+  become: true
+  tasks:
+    - name: 'Update packages'
```
akofink: I don't know if the installer will/should do this for you.

pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -0,0 +1,168 @@
...
+      yum:
+        name: '*'
+        state: latest
+    - name: 'Reinstall puppet-agent-oauth'
+      command: yum reinstall -y puppet-agent-oauth
```
akofink: This was a workaround for an issue in 3.5. I don't know if you need it here.

pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -0,0 +1,168 @@
...
+- hosts: pipeline-katello-3.6-3.7-centos7
+  become: true
+  vars:
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
+      - "--disable-system-checks"
+      - "--reset-puppet-server-jvm-extra-args"
+      - "--reset-puppet-server-puppetserver-version"
+      - "--reset-puppet-server-puppetserver-metrics"
+      - "--reset-puppet-server-puppetserver-experimental"
```
akofink: The installer should do this for you (I think).

pipelines/pipeline_katello_36_to_37_upgrade.yml
ehelms: Agreed, it should be droppable

pipelines/pipeline_katello_36_to_37_upgrade.yml
ehelms: It won't, so its needed.

pipelines/pipeline_katello_36_to_37_upgrade.yml
jturel: Do you mean only line 140?

akofink: I thought `--upgrade` covers all these relevant `--reset-puppet-*` options, but I could be wrong.

jturel:
@akofink updated!

pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -0,0 +1,156 @@
...
+    - name: 'Update packages'
+      yum:
+        name: '*'
+        state: latest
+
+- hosts: pipeline-katello-3.6-3.7-centos7
+  become: true
+  vars:
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
```
akofink: Wait, the installer just handles the puppet upgrade automagically? You don't need `foreman_installer_upgrade: True`?

jturel: No idea. I see this when the installer runs:

```
[ WARN 2018-06-11T19:12:20 verbose] -:puppet_version: 4.10.12                                                                                     
[ WARN 2018-06-11T19:12:20 verbose] +:puppet_version: 5.5.2 
```

What should I be looking for?

jturel: Trying with that upgrade flag right now

jturel: I'm going to wait until the upgrade related issues are fixed before looking into this more.

ehelms:
@jturel @akofink Mind if we merge this as is so that folks can spin up the RC ?

jturel:
@ehelms I'm OK with that. Do you want me to change away from the staging repos first?

pipelines/pipeline_katello_36_to_37_upgrade.yml
ekohl: I'd prefer to include the `update_os_packages` role so it also works on Debian.

ehelms:
Nope! pipelines are designed to test staging :)

PR #732, opened 2018-06-05
Add SSH key creation for Robottelo role to run on same host

ehelms:


roles/robottelo/tasks/ssh.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: 'Check if SSH key exists'
+  stat:
+    path: "{{ robottelo_ssh_key }}"
+  register: st
+
+- name: 'Create ssh keys'
+  command: "ssh-keygen -b 2048 -t rsa -f {{ robottelo_ssh_key }} -q -N ''"
+  when: st.stat.exists == false
```
ekohl: `command` also has a `creates:` that should avoid the need for a separate stat

roles/robottelo/tasks/ssh.yml
```diff
@@ -0,0 +1,15 @@
...
+
+- name: 'Add ssh key to authorized_keys'
```
ekohl: Won't this break vagrant SSH login or will that continue to work because that's the `vagrant user?

ehelms:
Updated - thanks @ekohl

roles/robottelo/tasks/main.yml
```diff
@@ -47,18 +49,24 @@
...
     dest: "{{ robottelo_directory }}/robottelo.properties"
     regexp: "^hostname="
-    line: "hostname={{ ansible_fqdn }}"
+    line: "hostname={{ robottelo_hostname }}"
     state: present
     backrefs: yes
 
-- name: Set robottelo logging
+- name: Set robottelo SSH
   lineinfile:
```
ekohl: Any reason this can't use `replace` here in as well like in logging below? Same for the above hostname.

ehelms: No reason it can't use replace. I tend to think of lineinfile as change one single line vs. replace which is designed to replace all occurrences. If you prefer use of replace I can update.

roles/robottelo/tasks/ssh.yml
```diff
@@ -0,0 +1,11 @@
...
+- name: 'Create ssh keys'
+  command: "ssh-keygen -b 2048 -t rsa -f {{ robottelo_ssh_key }} -q -N ''"
+  args:
+    creates: "{{ robottelo_ssh_key }}"
+
+- name: 'Add ssh key to authorized_keys'
+  copy:
+    remote_src: true
+    src: "{{ robottelo_ssh_key }}"
+    dest: /root/.ssh/authorized_keys
```
akofink: Doesn't this `cp {{ robottelo_ssh_key }} /root/.ssh/authorized_keys` ? Don't we actually want to do `cat {{ robottelo_ssh_key }}.pub >> /root/.ssh/authorized_keys` ?

akofink: What I'm saying is, authorized_keys isn't for private keys, and you can't just overwrite it, you have to append or lose all your other authorized keys.

ekohl: Oh you're right, the `src` should be `{{ robottelo_ssh_key }}.pub`. Good catch.

akofink: We also shouldn't replace the file but simply ensure our public key is in there: https://github.com/theforeman/forklift/pull/754

PR #736, opened 2018-06-07
Move to using K8s/Openshift modules from Ansible core

ehelms:
The modules we were using have been officially deprecated in favor of these modules in Ansible core as of 2.5.

### DIFF
--- a/containers/deploy/candlepin/deployments.yml
+++ b/containers/deploy/candlepin/deployments.yml
@@ -1,20 +1,21 @@
 ---
 - name: Scale Candlepin to 0
-  openshift_v1_deployment_config:
+  openshift_raw:
       name: candlepin
       namespace: "{{ project_name }}"
       replicas: 0
+      kind: DeploymentConfig
   tags:
     - stop
     - restart
 
 - name: Create Candlepin deployment
-  openshift_v1_deployment_config:
+  openshift_raw:
       state: "{{ deployment_state }}"
       force: false
-      resource_definition:
+      definition:
           apiVersion: v1
-          kind: deployment_config
+          kind: DeploymentConfig
           metadata:
               name: candlepin
               labels:
@@ -22,6 +23,11 @@
                   service: candlepin
               namespace: "{{ project_name }}"
           spec:
+              replicas: 1
+              strategy:
+                rollingParams:
+                  timeoutSeconds: 900
+                type: Rolling
               template:
                   metadata:
                       labels:
@@ -63,11 +69,6 @@
                                   key: ca.crt
                               secretName: ca
                           name: ca
-              replicas: 1
-              strategy:
-                  type: Rolling
-                  rollingParams:
-                      timeoutSeconds: 900
   tags:
     - start
     - restart
--- a/containers/deploy/foreman-proxy/deployments.yml
+++ b/containers/deploy/foreman-proxy/deployments.yml
@@ -1,6 +1,6 @@
 ---
 - name: Scale Foreman Proxy to 0
-  openshift_v1_deployment_config:
+  openshift_raw:
       name: foreman-proxy
       namespace: "{{ project_name }}"
       replicas: 0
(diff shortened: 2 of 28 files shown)


ehelms:
[test containers]

ehelms:
@carbonin Any chance you could glance at this?

ehelms:
[test containers]

containers/deploy/candlepin/deployments.yml
```diff
@@ -1,27 +1,38 @@
...
           spec:
+              replicas: 1
+              strategy:
+                resources: {}
+                rollingParams:
+                  intervalSeconds: 1
+                  maxSurge: 25%
+                  maxUnavailable: 25%
+                  timeoutSeconds: 900
+                  updatePeriodSeconds: 1
```
carbonin: Are these things we need because of the new library or separate enhancements?

containers/images/foreman/Dockerfile
```diff
@@ -34,7 +34,6 @@ COPY container-assets/start_foreman.sh /usr/bin/start_foreman.sh
 COPY container-assets/wait_on_postgres.py /usr/bin/wait_on_postgres.py
 COPY container-assets/create_qpid_queue.sh /usr/bin/create_qpid_queue.sh
 
-RUN yum -y reinstall http://koji.katello.org/kojifiles/work/tasks/1420/101420/tfm-rubygem-puma-3.11.4-1.el7.x86_64.rpm
```
carbonin: How is this related to the new ansible module?

containers/tools/deploy.yml
```diff
@@ -12,21 +12,12 @@
...
-    - name: 'Install python openshift'
-      command: pip install git+https://github.com/ehelms/openshift-restclient-python.git@fix-destination-ca
-
-    - name: 'Install kubernetes modules'
-      command: 'ansible-galaxy install -r requirements.yml'
-      args:
-        chdir: "{{ working_dir }}"
+    - name: 'Install openshift'
+      pip:
+        name: openshift
```
carbonin: The ansible module [docs](https://docs.ansible.com/ansible/latest/modules/openshift_raw_module.html#openshift-raw-module) list a rather strict version requirement for this package (`== 0.4.3`). Can we (and do we want to) enforce that here?

containers/deploy/candlepin/deployments.yml
ehelms: They were me testing something and copy from Openshift itself. I can drop them.

containers/images/foreman/Dockerfile
ehelms: It's not, but builds were breaking without it. I can split to it's down PR.

containers/tools/deploy.yml
ehelms: I noticed that too, and I've had no issues with newer versions. Further, newer versions handle some more advanced options better (such as the certificate_ca_file) so I am inclined to ride latest.

containers/images/foreman/Dockerfile
carbonin: I think just at least another commit with some details would be good.

containers/deploy/foreman-proxy/deployments.yml
```diff
@@ -9,12 +9,12 @@
     - restart
 
 - name: Create Foreman Proxy deployment
-  openshift_v1_deployment_config:
+  openshift_raw:
       state: "{{ deployment_state }}"
       force: false
       resource_definition:
```
carbonin: Does this need to change?

carbonin: Now that I'm looking for it. It looks like this was missed in quite a few places. Does this need to be changed to `definition` everywhere?

ehelms: Not required, this still works and is supported from what I can tell from the test run.

carbonin: Okay, then I think we should probably be consistent rather than using `definition` in some places and `resource_definition` in others. I'm fine either way, but I think it's strange to have both.

containers/deploy/candlepin/deployments.yml
```diff
@@ -1,27 +1,34 @@
...
           metadata:
               name: candlepin
               labels:
                   app: foreman
                   service: candlepin
               namespace: "{{ project_name }}"
           spec:
+              replicas: 1
+              strategy:
+                resources: {}
```
carbonin: Not sure we need this either.

PR #739, opened 2018-06-13
More complete luna

ares:
adds respective hammer plugins and virt_who_configure plugin which is in fact katello plugin and is not exposed in foreman-installer


playbooks/luna.yml
```diff
@@ -16,5 +16,14 @@
...
       - "--enable-foreman-proxy-plugin-openscap"
+      - "--enable-foreman-cli-discovery"
+      - "--enable-foreman-cli-openscap"
+      - "--enable-foreman-cli-remote-execution"
+      - "--enable-foreman-cli-tasks"
+      - "--enable-foreman-cli-templates"
+    foreman_installer_additional_packages:
+      - tfm-rubygem-foreman_virt_who_configure
+      - tfm-rubygem-hammer_cli_foreman_virt_who_configure
+    foreman_repositories_version: 1.18
```
akofink: Shouldn't this be nightly?

ares: oops sorry, that should have not been part of the commit

ekohl:
https://github.com/theforeman/puppet-foreman/pull/642 is the start. When that's merged it also needs a PR to the installer scenarios to expose them.

PR #740, opened 2018-06-13
Use downcase hostname for VM domain

PaulSD:
To prevent katello-installer failures:
https://github.com/Katello/katello-installer/blob/master/checks/hostname.rb#L31


PaulSD:
I don't believe the failing test is actually related to this change.

ekohl:
You're right, looks like a new rubocop rule.

PR #741, opened 2018-06-14
Add bats task to copy results to local host.

zjhuntin:


roles/bats/defaults/main.yaml
```diff
@@ -1,10 +1,12 @@
 ---
+bats_copy_results: true
```
ehelms: Would fetch be better than copy?

zjhuntin: Yeah, I think that makes sense.

PR #746, opened 2018-06-18
Re-factor deploy tasks into roles

ehelms:


ehelms:
[test containers]

containers/foreman.yml
```diff
@@ -8,7 +8,7 @@
     pulp_worker_count: 2
     registry: quay.io/foreman
     foreman_providers: false
-  tasks:
+  pre_tasks:
```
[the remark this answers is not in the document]
ehelms: Correct!

containers/deploy/qpid/tasks/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+- import_tasks: services.yml
```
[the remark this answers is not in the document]
ehelms: They are different in:

 *  include_tasks is dynamic (you can actually use a variable so it defined at execution time)
 * import_tasks is static

ehelms:
Ensure you have Ansible 2.5, then it should be just a:

```
ansible-playbook foreman.yml --tags start
```

ehelms:
[test containers]

ehelms:
I updated this to re-think things a bit. With an eye towards potential migrations later, I moved all the definitions of objects to templates and just simply import them into tasks. This also splits the entities out a bit more such as moving mongodb to it's own role. I also dropped all of the start/restart/destroy logic for now. I found that is was more confusing than usual in most of the current workflows and I'd rather add those back in after settling on a set of base deployment configuration.

@bdunne @carbonin If you find any time, having you look over these changes would be appreciated.

containers/foreman.yml
```diff
@@ -7,70 +7,24 @@
     cluster_up: false
     pulp_worker_count: 2
     registry: quay.io/foreman
-    foreman_providers: false
```
carbonin: Does this PR just remove this? To be clear, that's fine with me as it was really just for the POC.

ehelms: Aye, it does. Cleaning up as part of this change.

containers/deploy/mongodb/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Create MongoDB persistent volume claim
+  k8s_raw:
+    state: "{{ deployment_state }}"
+    definition: "{{ lookup('template', 'templates/' + item + '.pvc.yaml') | from_yaml }}"
```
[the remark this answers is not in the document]
ehelms: I debated that, but I liked the general pattern and it means I might can abstract this pattern out for use.

ehelms:
@carbonin Any thoughts on what you'd like to see naming wise? "Manage Foreman service" type of naming?

carbonin:
I think "Manage" is okay, maybe "Alter" would work too, but I would also be fine without a verb. I think just listing the object we're altering would work e.g `Foreman Deployment` `Memcached Service`

carbonin:
"Create" is just plain incorrect half of the time so as long as that's not the case anymore I'm good.

ehelms:
@carbonin I just adapted the simple case of stating the object

containers/deploy/pulp/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: pulp persistent volume claim
+  k8s_raw:
+    state: "{{ deployment_state }}"
+    definition: "{{ lookup('template', 'templates/' + item + '.pvc.yaml') | from_yaml }}"
+  with_items:
+    - pulp-data
+    - puppet-data
```
fabianvf: Would this PVC logically fit better in the puppet role?

containers/deploy/service-accounts/tasks/main.yml
```diff
@@ -2,26 +2,18 @@
...
-  tags:
-    - start
 
 - name: Login as developer
   command: oc login -u developer -p a
   when: minishift or cluster_up
-  tags:
-    - start
 
 - name: Ensure on project
```
fabianvf: This should no longer be necessary, the namespace field in the k8s_raw objects will take precedence

fabianvf: * assuming there are no oc commands being run

containers/deploy/service-accounts/tasks/main.yml
```diff
@@ -31,10 +23,6 @@
       metadata:
         name: anyuid
         namespace: "{{ project_name }}"
-  tags:
-    - start
 
 - name: Allow anyuid service account to create secrets
   command: oc adm policy add-role-to-user edit -z anyuid
```
fabianvf: This can be done with the Ansible module by creating a resource with the following structure:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: anyuid-secrets
  namespace: foreman
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: edit
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: ServiceAccount
  name: anyuid
```

containers/deploy/service-accounts/tasks/main.yml
```diff
@@ -2,26 +2,18 @@
 - name: Login as system admin
   command: oc login -u system:admin
   when: minishift or cluster_up
-  tags:
-    - start
 
 - name: Add anyuid scc to anyuid service account
```
fabianvf: Probably worth it to separate these from the lower-permissioned changes done as the developer user, and document them as requirements since many/most users probably can't log in as system:admin

containers/deploy/pulp/tasks/main.yml
ehelms: Its a tough yes and no. Just because I don't have puppet doesn't meant I can't manage puppet content and vice-versa. For example, you can have Pulp manage Puppet content act like a Puppet forge. Or you can have it put puppet modules on disk where the Puppet server expects them to be. TLDR; I don't have a good answer.

ehelms:
@fabianvf Thanks for the review! I did a chunk of re-working the service-account part over here (https://github.com/theforeman/forklift/pull/763/files#diff-0ee52f26ecb2a1088795282509749b28) and its one of the parts I think worthwhile for me to do as a separate PR to re-work all of that part. If you don't mind, I'll merge this as is and ping you on a follow up PR to look over changes to that area.

containers/deploy/pulp/tasks/main.yml
fabianvf: makes sense, main reason I asked was that the puppet role will fail if this pvc has not been created, which makes this task a dependency of the puppet module.

fabianvf: Or rather, the task won't fail, but the deployment will

PR #752, opened 2018-06-21
Run minishift locally only

akofink:
If you follow the [docs](https://github.com/theforeman/forklift/tree/master/containers#start-minishift), this playbook would run against all your running forklift vms as well as localhost, which I don't think is intended.


ehelms:
I'd suggest running with -l localhost. I'd be fine with a docs update to add that to commands or adding a default inventory to the `containers/` directory with localhost.

PR #754, opened 2018-06-22
Append pub key instead of overwrite with private

akofink:
We don't ever want to overwrite `authorized_keys`, and we definitely don't want private keys in there.

### DIFF
--- a/roles/robottelo/tasks/ssh.yml
+++ b/roles/robottelo/tasks/ssh.yml
@@ -4,8 +4,11 @@
   args:
     creates: "{{ robottelo_ssh_key }}"
 
+- name: 'Get public key'
+  command: cat {{ robottelo_ssh_key }}.pub
+  register: robottelo_public_key
+
 - name: 'Add ssh key to authorized_keys'
-  copy:
-    remote_src: true
-    src: "{{ robottelo_ssh_key }}"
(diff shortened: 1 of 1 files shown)


roles/robottelo/tasks/ssh.yml
```diff
@@ -4,8 +4,12 @@
...
 
+- name: 'Get public key'
+  command: cat {{ robottelo_ssh_key }}.pub
+  register: robottelo_public_key
+
 - name: 'Add ssh key to authorized_keys'
-  copy:
-    remote_src: true
-    src: "{{ robottelo_ssh_key }}"
+  lineinfile:
```
ekohl: If you've already registered, then why not use the `authorized_key` module?

akofink:
Updated to use authorized_key module.

akofink:
Actually, wait to merge - I think we should be able to do this with `lookup('file', {{ robottelo_ssh_key + '.pub'}})`

ekohl:
> Actually, wait to merge - I think we should be able to do this with lookup('file', {{ robottelo_ssh_key + '.pub'}})

I tried that, but that takes it from the host. In the vagrant context that means `mymachine` rather than (say) `centos7-katello-bats-ci.mymachine.example.com`.

akofink:
Should be good to go.

PR #758, opened 2018-06-26
Add a playbook to gather bats_results and a sosreport from any box

zjhuntin:


playbooks/debug_box.yml
```diff
@@ -0,0 +1,22 @@
+---
+- hosts: all
+  become: true
+  roles:
+    - sos_report
+
+- hosts: all
+  become: true
```
ehelms: You don't necessarily need to split these into two plays in one file, you can have:

```
roles:
  -
tasks:
  -
```

playbooks/debug_box.yml
```diff
@@ -0,0 +1,22 @@
...
+  vars:
+    bats_output_dir: '/root/bats_results'
+    bats_remote_dir: '/tmp/debug'
+  tasks:
+    - name: "Find bats files"
+      find:
+        paths: "{{ bats_output_dir }}"
+        patterns: "*.tap"
+      register: bats_results
+    - name: "Copy bats results"
```
ehelms: Empty line between tasks please.

playbooks/debug_box.yml
```diff
@@ -0,0 +1,22 @@
+---
```
ehelms: I'm thinking the playbook name could be a bit more descriptive. This is also doing two things that feel disjoint. Running a sos_report and fetching bats results. Maybe `gather_debug` ? `debug` ? or `collect_debug` ?

PR #760, opened 2018-06-26
Update install-forklift playbook with all it's dependencies

zjhuntin:


playbooks/setup_forklift.yml
```diff
@@ -7,15 +7,23 @@
...
 - hosts: all
   become: true
   tasks:
-    - name: install git
+    - name: 'Install pip'
+      package:
+        name: 'python2-pip'
+        state: 'present'
+    - name: 'Install ansible'
+      pip:
```
ekohl: Any reason to install this via pip rather than `package`? It's in EPEL

PR #767, opened 2018-06-29
Update to use newly released Ansible 2.6 k8s module

fabianvf:

PR #768, opened 2018-06-29
Allow user to enable specific container services

fabianvf:
depends on #767


ehelms:
@fabianvf This can be rebased now

fabianvf:
@ehelms done

containers/deploy/foreman/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+
```
ehelms: Nit - empty line not needed here

containers/foreman.yml
```diff
@@ -8,6 +8,12 @@
     pulp_worker_count: 2
     registry: quay.io/foreman
     deployment_state: present
+    enabled_services:
+      - postgres
+      - puppet
+      - pulp
+      - candlepin
+      - qpid
```
ehelms: Can actually drop qpid here since Candlepin and Pulp require it.

containers/deploy/puppet/tasks/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: puppet persistent volume claim
+  k8s_raw:
+    state: "{{ deployment_state }}"
+    definition: "{{ lookup('template', 'templates/' + item + '.pvc.yaml') | from_yaml }}"
+  with_items:
+    - puppet-data
```
ehelms: This move means that Pulp now needs to require the Puppet role because Pulp mounts this volume. Which feels a bit odd to be honest, but that would at least keep it consistent for now.

containers/foreman.yml
fabianvf: should I remove it from foreman/defaults/main.yml as well then? Without qpid in this list it won't be included by foreman.

fabianvf:
@ehelms responded

ehelms:
[test containers]

containers/README.md
```diff
@@ -63,6 +63,13 @@ Once services are up and the health check returns OK for all services the applic
 
 Using the entry under HOST/PORT you can now browse to `https://<hostname>` and access the Foreman web UI.
 
+
+#### Enabling specific services
+
+By default the deployment playbook will deploy Foreman, PostgreSQL, Puppet, Pulp, Candlepin, Qpid. You can specify which of these components you want to deploy with the `enabled_services` variable. For example, to deploy only Foreman and PostgreSQL, run:
+
+    ansible-playbook foreman.yml -e enabled_services=postgres
```
ehelms: A slightly better example might be Foreman and only Puppet? I think the assumption by default is Foreman means you get Postgres so the application works ou tof the box.

ehelms:
Looks like I caused a need for rebase.

fabianvf:
@ehelms rebased

PR #769, opened 2018-06-29
Add tooling to allow build-tag-push to Openshift registry

ehelms:


containers/tools/build-tag-push-openshift.yaml
```diff
@@ -0,0 +1,28 @@
+---
+- hosts: localhost
+  gather_facts: no
+  vars:
+    registry: 172.30.1.1:5000
```
ekohl: Is this always the same? There's a `registry` command which suggests we should set it via a fact instead of hardcoded.

ehelms:
@ekohl Fair point, updated

PR #774, opened 2018-07-03
Support multiple deployments within the same cluster

ehelms:


containers/foreman.yml
```diff
@@ -18,10 +18,10 @@
       register: minishift_ip
       when: minishift == true
     - set_fact:
-        application_hostname: "foreman.{{ minishift_ip.stdout }}.nip.io"
+        application_hostname: "{{ project_name }}.{{ minishift_ip.stdout }}.nip.io"
```
ekohl: Should this be lower cased just to be sure?

ehelms: Addressed here since project also needs ot be lowercase when creating it -- https://github.com/theforeman/forklift/pull/774/files#diff-f27844b8d159df7c0a792ba92a045fdfR14

PR #777, opened 2018-07-11
Enable remote execution option for foreman-installer

rahulbajaj0509:


roles/foreman_installer_devel_scenario/files/katello-devel-answers.yaml
```diff
@@ -32,3 +32,5 @@ puppet:
 foreman_proxy::plugin::pulp:
   enabled: true
   pulpnode_enabled: false
+foreman_proxy::plugin::remote_execution::ssh:
```
ekohl: I'd set the default to false and perhaps add all plugins `foreman` supports. That way users can easily install the proxy side of a feature but stays small by default.

PR #791, opened 2018-07-22
Fix katello 3.7 upgrade pipeline

jturel:
This fixes the 3.6 to 3.7 upgrade pipeline except for the final fb-proxy.bats test after upgrading the proxy to 3.7+puppet5. I think I'm following the correct steps to perform the upgrades based on documentation.

The reason the last fb-proxy fails is because the pulp db has pending migrations. Running pulp-mange-db by hand fixes it, and running fb-proxy.bats by hand is then successful. Maybe we're missing something in the installer?

To reproduce:
```
ansible-playbook pipelines/pipeline_katello_36_to_37_upgrade.yml -e "forklift_state=up"
```

and sit tight ...

@ehelms @chris1984 could this be related to the mongo upgrade or maybe I'm performing the upgrade incorrectly?

### DIFF
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -203,6 +203,11 @@ EOF
   tPackageExists walrus-5.21
 }
 
+# it seems walrus lingers around making subsequent runs fail, so lets test package removal!
+@test "package remove (katello-agent)" {
+  timeout 300 hammer host package remove --host $(hostname -f) --packages walrus
+}
+
 @test "add puppet module to content view" {
   repo_id=$(hammer repository list --organization="${ORGANIZATION}" \
     | grep Puppet | cut -d\| -f1 | egrep -i '[0-9]+')
--- a/pipelines/pipeline_katello_36_to_37_upgrade.yml
+++ b/pipelines/pipeline_katello_36_to_37_upgrade.yml
@@ -31,6 +31,7 @@
     - update_os_packages
     - haveged
     - disable_firewall
+    - foreman_installer
 
 - hosts: pipeline-katello-3.6-3.7-centos7
   become: yes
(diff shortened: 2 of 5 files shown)


bats/fb-content-katello.bats
```diff
@@ -203,6 +203,11 @@ EOF
   tPackageExists walrus-5.21
 }
 
+# it seems walrus lingers around making subsequent runs fail, so lets test package removal!
+@test "package remove (katello-agent)" {
+  timeout 300 hammer host package remove --host $(hostname -f) --packages walrus || true
```
ekohl: If `|| true`, will this test ever fail?

pipelines/pipeline_katello_36_to_37_upgrade.yml
```diff
@@ -103,44 +105,44 @@
...
-# Upgrade to Puppet 5
+# Upgrade server+proxy to Puppet 5
 # http://projects.theforeman.org/projects/foreman/wiki/Upgrading_from_Puppet_4_to_5#Prepare-And-Run-The-Installer
 - hosts:
   - pipeline-katello-3.6-3.7-centos7
   - pipeline-proxy-3.6-3.7-centos7
   become: true
   vars:
     puppet_repositories_version: 5
+    katello_repositories_version: 3.7
```
ekohl: This shouldn't be needed because it was previously set up.

ekohl: Shouldn't we do the Puppet 5 upgrade as part of the 3.6 -> 3.7 katello server as well then? We can do the same below with the proxy upgrade. This whole step was a workaround for an issue there was in the 3.5 migration. This should be fixed now.

jturel:
@ekohl thanks for the feedback- I made some updates. mind taking another look?

PR #793, opened 2018-07-23
Move hg param for puppet

xprazak2:
Add a parameter for puppet to hostgroup
after the hostgroup has been created, not before.


ekohl:
Thanks!

PR #794, opened 2018-07-24
Correct directory naem + handle package state deprecation

ekohl:


ehelms:
Typo in the commit message `naem`

PR #795, opened 2018-07-24
[WIP] Create a katello upgrade pipeline

zjhuntin:


pipelines/pipeline_upgrade_nightly.yml
```diff
@@ -0,0 +1,72 @@
...
+  roles:
+    - forklift
+
+- hosts: pipeline-upgrade-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 4
+    katello_repositories_version: 3.6
+    foreman_repositories_version: 1.17
+    foreman_installer_skip_installer: true
```
ekohl: This looks redundant because you don't include the `foreman_installer` role. Since it's just a single system you can however. There's no benefit to splitting it here. We do that on multiple hosts so we can parallelize the base OS install.

zjhuntin: Gotcha, thanks for the comment!

zjhuntin:
Sure, I think that makes sense, I guess I didn't because it feels redundant with the katello-release-nightly test, but still probably better to see it works in this case as well.

pipelines/pipeline_upgrade_nightly.yml
```diff
@@ -0,0 +1,72 @@
+- hosts: localhost
+  vars:
+    forklift_name: pipeline-katello-3.6-3.7
```
ehelms: 3.7 to nightly?

ehelms:
ACK -- let's run the tests. If you look at th existing upgrade pipelines they run those tests
I would mirror those jobs.

pipelines/pipeline_upgrade_nightly.yml
```diff
@@ -0,0 +1,82 @@
...
+  roles:
+    - foreman_installer
+
+- hosts: pipeline-upgrade-centos7
+  become: true
+  vars:
+    katello_repositories_version: 3.7
+    foreman_repositories_version: 1.18
+    foreman_installer_upgrade: True
+    foreman_repositories_environment: staging
```
ekohl: 3.7 is GA so we can use release.

pipelines/pipeline_upgrade_nightly.yml
```diff
@@ -0,0 +1,82 @@
...
+    - katello_repositories
+    - update_os_packages
+    - haveged
+    - disable_firewall
+
+- hosts: pipeline-upgrade-centos7
+  become: yes
+  vars:
+    foreman_installer_scenario: katello
+    foreman_installer_options_internal_use_only:
```
ekohl: Perhaps we should add https://github.com/theforeman/forklift/blob/52da2e3270c694c4e407760f9df6782636bad6d1/pipelines/pipeline_foreman_nightly.yml#L19-L21 to reduce memory usage a bit?

ehelms:
Is it intentional to get rid of the standard pipeline for nightly for Katello? Is the upgrade not a separate pipeline?

zjhuntin:
@ekohl @ehelms Okay, I think this one is updated to the point of being good to go as well.

pipelines/pipeline_upgrade_nightly.yml
zjhuntin: I think I added this properly to the end of this.

pipelines/pipeline_katello_upgrade_nightly.yml
```diff
@@ -0,0 +1,85 @@
...
+        box: centos7
+        memory: 6144
+  roles:
+    - forklift
+
+- hosts: pipeline-upgrade-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 4
+    katello_repositories_version: 3.6
```
ehelms: Should this be 3.8?

pipelines/pipeline_katello_upgrade_nightly.yml
```diff
@@ -0,0 +1,85 @@
...
+      - "--puppet-server-jvm-max-heap-size 1G"
+    foreman_installer_additional_packages:
+      - katello
+  roles:
+    - foreman_installer
+
+- hosts: pipeline-upgrade-centos7
+  become: true
+  vars:
+    katello_repositories_version: 3.7
```
ehelms: This should be nightly right?

pipelines/pipeline_katello_upgrade_nightly.yml
```diff
@@ -0,0 +1,85 @@
...
+    - etc_hosts
+    - epel_repositories
+    - puppet_repositories
+    - foreman_repositories
+    - katello_repositories
+    - update_os_packages
+    - haveged
+    - disable_firewall
+
+- hosts: pipeline-upgrade-centos7
```
ekohl: Any reason this step isn't done in the host definition above?

pipelines/pipeline_katello_upgrade_nightly.yml
```diff
@@ -0,0 +1,85 @@
...
+      pipeline-upgrade-centos7:
+        box: centos7
+        memory: 6144
+  roles:
+    - forklift
+
+- hosts: pipeline-upgrade-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 4
```
ekohl: I'd prefer Puppet 5 unless you upgrade it in the next step.

zjhuntin: Hmmm, upgrading it might be a good plan for it. I'll change it so that it does that.

PR #796, opened 2018-07-25
Add Katello 3.8 pipeline

ehelms:


zjhuntin:
ACK

ehelms:
@zjhuntin Mind expanding on your concern?

playbooks/setup_forklift.yml
```diff
@@ -22,5 +25,6 @@
         state: 'present'
     - name: 'Clone Forklift'
       git:
-        repo: 'http://github.com/theforeman/forklift'
+        repo: "http://github.com/{{ github_org }}/forklift"
```
ekohl: I'd just use `forklift_url` so you're not tied to github. You can also change it to https://

ehelms: Where else would I get than from github?

ekohl: I imagine there could be instances where you have a local mirror. I also think `forklift_url` is a better naming than `github_org`.

evgeni: We have already something similar in the `bats` role:

https://github.com/theforeman/forklift/blob/96a04c3e180ee8794dfa7df30602ae6126c0caa0/roles/bats/tasks/main.yml#L33-L38

should that be moved out to an own role instead?

evgeni: (and strictly speaking, this does not belong into this PR anyways ;))

ehelms:
Updated and switch to forklift_url

playbooks/setup_forklift.yml
```diff
@@ -11,16 +11,22 @@
 
 - hosts: all
   become: true
+  vars:
+    forklift_url: "http://github.com/theforeman/forklift"
```
ekohl: Could you switch this to https to avoid the redirect?

pipelines/pipeline_katello_38.yml
```diff
@@ -0,0 +1,83 @@
...
+    - forklift
+
+- hosts:
+  - pipeline-katello-3.8-centos7
+  - pipeline-proxy-3.8-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 4
+    katello_repositories_version: 3.8
+    foreman_repositories_version: 1.18
```
ekohl: 3.8 is 1.19

ehelms:
Updated

evgeni:
IMHO this looks good. I'd role-out the forklift stuff in another PR.

@ekohl as you were raising concerns, please have a final ack :)

pipelines/pipeline_katello_38.yml
```diff
@@ -0,0 +1,83 @@
...
+        memory: 3072
+  roles:
+    - forklift
+
+- hosts:
+  - pipeline-katello-3.8-centos7
+  - pipeline-proxy-3.8-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 4
```
evgeni: Oh, btw, should this be 5 these days?

ehelms:
@evgeni Good catch, updated

PR #797, opened 2018-07-26
Fix the disabling of epel

xprazak2:
When we use url in 'name',
disabling the epel repo with yum module
does not work. We need to change the name
to get desired effects.

### DIFF
--- a/roles/epel_repositories/defaults/main.yml
+++ b/roles/epel_repositories/defaults/main.yml
@@ -1,2 +1,4 @@
 ---
 epel_repositories_state: present
+epel_repositories_url: https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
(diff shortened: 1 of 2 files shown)


roles/epel_repositories/defaults/main.yml
```diff
@@ -1,2 +1,4 @@
 ---
 epel_repositories_state: present
+epel_repositories_url: http://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
```
ekohl: Would you mind changing this to https while you're at it?

roles/epel_repositories/defaults/main.yml
```diff
@@ -1,2 +1,4 @@
...
+epel_repositories_name: "{{ epel_repositories_url if epel_repositories_state == 'present' else 'epel-release' }}"
```
ekohl: CentOS extras also carries `epel-release` but this makes it work on RHEL as well so :+1:

xprazak2:
I changed http to https.

PR #803, opened 2018-07-26
Add starting a dev server instructions

johnpmitsch:
This is an area there is a lot of confusion around. This
commit documents the various ways to start a dev server,
what the servers that we run are, and the limitations of
each way.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -10,7 +10,6 @@ This covers how to setup and configure a development environment using the Forkl
  * [Hammer Development](#hammer-development)
  * [Capsule Development](#capsule-development)
  * [Client Development](#client-development)
- * [Webpack](#webpack)
  * [Dynflow](#dynflow-development)
 
 ## Development Environment Deployment
@@ -58,7 +57,36 @@ cd foreman
 bundle exec foreman start
 ```
 
-The Foreman UI is then accessible on `https://centos7-devel.<hostname>.example.com/`, where `<hostname>` is the shortname of your workstation.  While using the webpack dev server, you must use HTTPS and the hostname, as well as visiting `https://centos7-devel.<hostname>.example.com:3808` to accept the self-signed certificate.
+### Starting the Development Server
+
+Our backend requires a rails server to be running. We also use a webpack server for the more "modern" parts of our front-end UI.
+[Webpack](https://webpack.js.org/) is a way of bundling up multiple front-end files and assets to a small amount of files to send to the browser. 
+The files that webpack handles are located in `webpack/` directory found in Foreman, Katello,
+and plugin root directories. If you are editing any files in `webpack/` and want to have your changes refresh automatically, you will need a webpack server running.
+
+Because we are using a webpack server in conjunction with a rails server, there are different ways of starting a server depending on your needs and preferences. The following are instructions for starting the server using a base `centos7-devel` box as a starting point.
+
+#### Run a rails and webpack server together using `foreman start`
+- Run `bundle exec foreman start` in `~/foreman`
+- Navigate to `https://centos7-devel.<hostname>.example.com/` where `<hostname>` is the shortname of your hypervisor (machine your VM is running on).
+- Accept the self-signed certs at `https://centos7-devel.<hostname>.example.com:3808`.
+- Everything should be set for you to run `bundle exec foreman start` to start your dev server as needed.
+
+NOTE: The `foreman` in `foreman start` is actually [this gem](https://github.com/ddollar/foreman) and not our `foreman`. It
+doesn't allow STDIN, which means it doesn't support local debugging. To use a debugger with `foreman start`, you can use
+remote debugging with a tool like [pry-remote](https://github.com/Mon-Ouie/pry-remote). There is a helpful blog post about
+remote debugging [here](http://blog.honeybadger.io/remote-debugging-with-byebug-rails-and-pow/) with more information.
+
+#### Run the rails server without a webpack server running
+If you don't need to do any development in the `webpack/` directories, you can turn the webpack server off and run only the rails server if desired.
+
+- Run `foreman-installer --katello-devel-webpack-dev-server false` to disable the webpack server. You can also set this installer option
+in your boxes.yaml entry to configure this on box creation.
+- Start the rails server with `bundle exec foreman start rails` in `~/foreman`
+  - Alternatively, you can start the rails server with `bundle exec rails s`
+
+#### Additional info
+- `foreman start` can be run with `rails` or `webpack` to start just that server: i.e. `bundle exec foreman start rails`. This can be used to start servers separately.
 
 ### Reviewing Pull Requests
 


docs/development.md
```diff
@@ -55,10 +54,42 @@ The box can now be accessed via ssh and the Rails server started directly (this
...
+[Webpack](https://webpack.js.org/) is a way of bundling up multiple front-end files and assets to a small amount of files to send to the browser. 
+The files that webpack handles are located in `webpack/` directory found in Foreman, Katello,
+and plugin root directories. If you are editing any files in `webpack/`, and want to have your changes refresh automatically, you will need a webpack server running.
+
+Because we are using the webpack server in conjunction with a rails server, there are different ways of starting a server depending on your needs and preferences. The following are the
+instructions for starting the server using a base `centos7-devel` box as a starting point.
+
+#### Run a rails and webpack server together using `foreman start`
+- run `foreman start` in `~/foreman`
+- navigate to `https://centos7-devel.<hostname>.example.com/` (or use your prefered way of navigating your dev server).
```
stbenjam: Can we still say "where `<hostname>` is the shortname of your workstation"? It might not be obvious what `<hostname>` means.

Also "preferred way of navigating" is a little confusing, with the `--public centos7-devel.<hostname>.example.com` in our default config, this is the *only* way it will work unless we make `--disable-host-check` the default.

docs/development.md
```diff
@@ -55,10 +54,42 @@ The box can now be accessed via ssh and the Rails server started directly (this
...
+  - If you are doing remote development, you will have to forward port 3808 from devel VM to your hypervisor. i.e. in forklift directory,
+    run `vagrant ssh centos7-devel -- -L 0.0.0.0:3808:localhost:3808`
+- Everything should be set for you to run `foreman start` to start your dev server as needed.
+
+NOTE: the `foreman` in `foreman start` is actually [this gem](https://github.com/ddollar/foreman) and not our `foreman`. It
+doesn't allow STDIN, which means it doesn't support debuggers like `byebug` and `pry` very well. To support debuggers, run the rails
+server and webpack separately as described below.
+
+
+#### Run the rails server without a webpack server running
```
ekohl: There is an installer option for this: `--katello-devel-webpack-dev-server false`. It'll even do the initial `webpack:compile` for you.

docs/development.md
```diff
@@ -55,10 +54,42 @@ The box can now be accessed via ssh and the Rails server started directly (this
...
+If you don't need to do any development in the `webpack/` directories, you can turn the wepback server off and just run the rails server if desired.
+
+- Change ':webpack_dev_server:` to `false` in `~/foreman/config/settings.yaml`
+- run `rake webpack:compile` in `~/foreman`, this bundles the assets that the webpack server would manage.
+- start the rails server with `rails s` in `~/foreman`
+  - use `rails s -b 0.0.0.0` if you are doing remote development
```
ekohl: I'd advise to standardize on foreman: `bundle exec foreman start rails`

docs/development.md
```diff
@@ -55,10 +54,42 @@ The box can now be accessed via ssh and the Rails server started directly (this
...
+#### Run a rails and webpack server together using `foreman start`
+- run `foreman start` in `~/foreman`
+- navigate to `https://centos7-devel.<hostname>.example.com/` (or use your prefered way of navigating your dev server).
+- accept the self-signed certs at `https://centos7-devel.<hostname>.example.com:3808`.
+  - If you are doing remote development, you will have to forward port 3808 from devel VM to your hypervisor. i.e. in forklift directory,
+    run `vagrant ssh centos7-devel -- -L 0.0.0.0:3808:localhost:3808`
+- Everything should be set for you to run `foreman start` to start your dev server as needed.
+
+NOTE: the `foreman` in `foreman start` is actually [this gem](https://github.com/ddollar/foreman) and not our `foreman`. It
+doesn't allow STDIN, which means it doesn't support debuggers like `byebug` and `pry` very well. To support debuggers, run the rails
```
stbenjam: Can we promote remote debugging as the default instead of saying it doesn't support debuggers? Remote debugging works fine with foreman gem.

 - https://github.com/Mon-Ouie/pry-remote
 - http://blog.honeybadger.io/remote-debugging-with-byebug-rails-and-pow/

docs/development.md
```diff
@@ -55,10 +54,42 @@ The box can now be accessed via ssh and the Rails server started directly (this
...
+server and webpack separately as described below.
+
+
+#### Run the rails server without a webpack server running
+If you don't need to do any development in the `webpack/` directories, you can turn the wepback server off and just run the rails server if desired.
+
+- Change ':webpack_dev_server:` to `false` in `~/foreman/config/settings.yaml`
+- run `rake webpack:compile` in `~/foreman`, this bundles the assets that the webpack server would manage.
```
stbenjam: We shouldn't be recommending manual file changes.  Running the installer with `--katello-devel-webpack-dev-server=false` will do all the right things.

docs/development.md
```diff
@@ -55,10 +54,42 @@ The box can now be accessed via ssh and the Rails server started directly (this
...
+- start the rails server with `rails s` in `~/foreman`
+  - use `rails s -b 0.0.0.0` if you are doing remote development
+
+
+#### Run a rails server and webpack server separately
+- You can use `./scripts/foreman-start-dev` in `~/foreman` to start a rails and webpack server together that supports STDIN (read: this supports debuggers). Currently, this doesn't work well with remote development.
```
ekohl: I'm not a fan of yet another way to start it. It's equivalent to `bundle exec foreman start webpack & bundle exec foreman start rails`. Because the background process gives noise it might be even better to mention the steps separately so users can execute them in two terminals.

docs/development.md
```diff
@@ -55,10 +54,42 @@ The box can now be accessed via ssh and the Rails server started directly (this
...
+server and webpack separately as described below.
+
+
+#### Run the rails server without a webpack server running
+If you don't need to do any development in the `webpack/` directories, you can turn the wepback server off and just run the rails server if desired.
+
+- Change ':webpack_dev_server:` to `false` in `~/foreman/config/settings.yaml`
+- run `rake webpack:compile` in `~/foreman`, this bundles the assets that the webpack server would manage.
+- start the rails server with `rails s` in `~/foreman`
+  - use `rails s -b 0.0.0.0` if you are doing remote development
```
stbenjam: Wouldn't it make sense to just have users do `source .env; rails s` instead, in case .env later contains things other than -b needed to start rails?

docs/development.md
ekohl: Also note that the domain can be overridden. Either at the box level or global settings.

docs/development.md
stbenjam: Oh, @ekohl's suggestion is better: `bundle exec foreman start rails`

docs/development.md
johnpmitsch: I didn't realize you can do `foreman start webpack` and `foreman start rails`, I'll test that out and updated

docs/development.md
johnpmitsch: I can definitely suggest remote debugging here, but I want to be sure that users are aware that `foreman start` wont work with local debuggers. I'll update to say local debugging is not supported and suggest using remote debugging

docs/development.md
johnpmitsch: >Can we still say "where <hostname> is the shortname of your workstation"? It might not be obvious what <hostname> means.

Sure thing

>Also "preferred way of navigating" is a little confusing, with the --public centos7-devel.<hostname>.example.com in our default config, this is the only way it will work unless we make --disable-host-check the default.

many of us use reverse proxies, which does work with these steps, but you navigate to \<myhypervisor\>:\<port\>, I was trying to support that workflow but maybe its not worth mentioning

stbenjam: > many of us use reverse proxies, which does work with these steps, but you navigate to <myhypervisor>:<port>, I was trying to support that workflow but maybe its not worth mentioning

It probably is, since you're not the only one who uses it, but maybe could go in a separate section., and maybe as a separate PR? Since we need to forward 3808 too in 
https://github.com/theforeman/forklift/blob/master/vagrant/lib/forklift/box_distributor.rb#L244, and change the --public entry, maybe with an installer flag.

johnpmitsch:
@stbenjam @ekohl updated in a separate commit. A couple things to clarify:

- I kept the part about forwarding port 3808 as that is pretty important information for a reverse proxy setup. I also clarified that this may be helpful in a reverse proxy setup. I know this is being updated in forklift itself, but my goal is to document the process as it is today. We can update the documentation as improvements to forklift are added.

- I gave the alternative of `bundle exec rails s` as a way of starting the rails server. I know we have moved to preferring `foreman start`, but I would like to give options to devs as they have different preferences.

stbenjam:
> I kept the part about forwarding port 3808 as that is pretty important information for a reverse proxy setup. I also clarified that this may be helpful in a reverse proxy setup. I know this is being updated in forklift itself, but my goal is to document the process as it is today. We can update the documentation as improvements to forklift are added.

I don't think this is really ok, it's a single line change to fix the port forwarding for 3808.

johnpmitsch:
> I don't think this is really ok, it's a single line change to fix the port forwarding for 3808.

@stbenjam I really meant for this to be a documentation change only, meaning I am documenting the tool as it exists today. Are you suggesting we change the port forwarding in this PR? If so, I'll need some help on how to do so. 

I do have some technical questions around the port forwarding, so I would advocate again that we handle this separately and update the documentation along with the fix.

docs/development.md
```diff
@@ -55,10 +54,40 @@ The box can now be accessed via ssh and the Rails server started directly (this
 ```sh
 vagrant ssh centos7-devel
 cd foreman
-bundle exec foreman start
```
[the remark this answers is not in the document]
johnpmitsch: Updated to keep this in, I originally removed it in lieu of giving the different options for starting the dev server. Maybe this should be added the quickstart section in the README instead?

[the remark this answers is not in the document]
ekohl: Given the above lines it's pretty long because it requires some setup. If we can simply out-of-the-box run this without additional config, it's probably worth it but maybe a follow up?

stbenjam:
It's a one-liner here:
  https://github.com/theforeman/forklift/blob/master/vagrant/lib/forklift/box_distributor.rb#L242

It can be a separate PR, but I don't think we should document something that requires a lot of hoops when it can just work in Forklift.

johnpmitsch:
@stbenjam per our discussion on IRC, I removed the block about port forwarding and will consider a better way of documenting this or integrating it into forklift.

PR #808, opened 2018-08-03
Fix default katello-client and related docs

jturel:
The default katello-client was not valid and would not register - until now!

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -251,32 +251,23 @@ capsule-dev:
 
 ## Client Development
 
-In `99-local.yaml`:
-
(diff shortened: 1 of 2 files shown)


docs/development.md
```diff
@@ -251,33 +251,24 @@ capsule-dev:
...
       katello_client_environment: 'Library'
       katello_client_username: 'admin'
       katello_client_password: 'changeme'
-
-```
-then add
-
-```yaml
-  ansible:
-    group: 'server'
```
jturel: I don't see how adding this group to the katello server box would affect anything - I didn't need it to get this working. Let me know if I've missed something.

PR #813, opened 2018-08-10
Use staging repositories for final upgrade

ehelms:


ehelms:
@zjhuntin Locally this fixed the breaking upgrad enightly pipeline for me, and it makes sense if we are running this as a check of our staging before pushing to release.

PR #814, opened 2018-08-13
Add 3.8 upgrade pipeline

ehelms:


pipelines/pipeline_katello_upgrade_38.yml
```diff
@@ -0,0 +1,82 @@
...
+      - "--disable-system-checks"
+  roles:
+    - foreman_repositories
+    - katello_repositories
+    - foreman_installer
+
+- hosts: pipeline-upgrade-centos7
+  become: yes
+  vars:
+    puppet_repositories_version: 4
```
ekohl: Should we upgrade to 5 here like we do in nightly?

ehelms: Dangit, I swear I put a 5 there. Gonna update.

ehelms: Note this is the 3.6 install which was Puppet 4 by default. Note I have it upgrading to Puppet 5 in 3.7 step

PR #815, opened 2018-08-16
Use foreman meta role + rackspace workaround

ekohl:


roles/rackspace_hostname/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+      package:
+        name: xenstore-utils
+        state: present
+
+    - name: 'Read the hostname'
+      command: xenstore-read vm-data/hostname
+      register: rackspace_hostname_desired_name
+
+    - name: 'Set the hostname based on xenstore'
+      command: 'hostnamectl set-hostname {{ rackspace_hostname_desired_name.stdout }}'
```
komidore64: I think you need double-quotes for interpolation?

PR #820, opened 2018-08-21
Add epel_repositories before haveged role

zjhuntin:


ekohl:
Should we instead move `haveged` into the foreman meta role? This now duplicates the EPEL repo setup.

zjhuntin:
Sure, i just wasn't sure if there was an expectation of it to be before the foreman role.

ehelms:
We originally added `haveged` to deal with our low memory usage requirement on Rackspace. Do we still need it?

I think it feels weird to be part of the Foreman meta. Does Foreman requirement haveged to operate? If no, then it shouldn't be included.

zjhuntin:
@ekohl updated

ekohl:
> Does Foreman requirement haveged to operate? If no, then it shouldn't be included.

Requirement is a big word. It won't break without it, but we use haveged to provide more entropy which is useful during installation (cert generation etc). After that it's not that useful for our use cases.

ekohl:
> We originally added haveged to deal with our low memory usage requirement on Rackspace. Do we still need it?

It's not memory, it's entropy.

ehelms:
Right it's entropy, but due to the low memory usage on Rackspace. Neither here nor there. The real question is do we require or even recommend that users install haveged on their systems?

zjhuntin:
@ehelms @ekohl I'm unsure what the direction y'all want to go in. I'm happy either way but a consensus from the two of you I think is needed before this issue can be totally solved.

If haveged is something we only used to solve issues dealing with rackspace and are solved by ci.centos.org than I think we can get rid of it. I also think we can get rid of it if we don't have users use it. In part because if we don't have enough entropy to generate certs randomly enough than we should know that. Unless we are having users use it as well, in which case lets keep using it.

ekohl:
> Right it's entropy, but due to the low memory usage on Rackspace.

No, you can have a ton of memory but very little entropy. Entropy is derived from sources of randomness in the kernel. On VMs there is typically very little of that. That can make private key generation (which needs a lot of entropy) very slow. Haveged can help with that but I don't know how much it actually helps these days.



> The real question is do we require or even recommend that users install haveged on their systems?

We make no statement whatsoever about haveged but I've seen it in deployed in multiple environments.

IMHO we should just merge this now. If we want to remove it that's a separate discussion.

ehelms:
Perhaps, but I am not a big fan of jamming everything into a meta role just to make things more DRY. DRY is good, but explicit can also be good. I think a meta should, at max, represent only what is exactly required to operate. Everything else should be situational and environmental. We can merge this, but stating I do not like this design change.

PR #823, opened 2018-08-22
Add a script to download boxes from Jenkins jobs

ekohl:
This should make it easy to download the systest box definitions and run them locally:

```
$ ./bin/box-downloader "https://ci.theforeman.org/job/release_test/237/label=el&&ipv6,os=debian9/"
Downloading 80-tmp-systest-foreman-debian9-237.yaml to vagrant/boxes.d/80-tmp-systest-foreman-debian9-237.yaml
Found boxes: systest-foreman-debian9-237
```

Now it's a simple vagrant up systest-foreman-debian9-237 to rerun the tests.

### DIFF
--- a/bin/box-downloader
+++ b/bin/box-downloader
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+
+import os
+
+import click
+import requests
+import yaml
+
+@click.command()
+@click.argument('url')
+def download(url):
+    """Simple program to download Jenkins artifacts"""
+    session = requests.Session()
+
+    response = session.get(url + 'api/json')
+    response.raise_for_status()
+    data = response.json()
+
+    artifacts = {artifact['fileName']: artifact['relativePath'] for artifact in data['artifacts']
+                 if artifact['fileName'].endswith('.yaml')}
+
+    for filename, path in artifacts.items():
+        response = session.get(url + 'artifact/' + path)
+        response.raise_for_status()
+        target = os.path.join('vagrant/boxes.d', filename)
+        with open(target, 'wb') as fp:
+            click.echo('Downloading {} to {}'.format(filename, target))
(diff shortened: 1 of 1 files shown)


ehelms:
Why do we need to download the boxes? As in, what is it that is defining such dynamic boxes in CI?

ekohl:
While releasing I have downloaded the box with the right parameters to reproduce it locally. The systest builder in foreman-infra constructs it with multiple parameters and it's good to get the exact combination. https://ci.theforeman.org/job/foreman-mysql-nightly-systest/ sets the database to mysql for example. In a release you get the correct repository versions and the expected version set so it's also checked that exactly 1.19.0-RC3 is installed. This should make it easy for any developer to reproduce it.

ehelms:
Two thoughts, one related directly and one related to the job that sparked this:

 * Should we err on the side of simply defining the boxes explicitly used in CI here for easier re-creation and testing directly by users? This simplifies the Jenkins usage and makes things more explicit.
 * Should we roll these mysql systests into the nightly release pipeline instead of as a stand-alone job?

ekohl:
> Should we err on the side of simply defining the boxes explicitly used in CI here for easier re-creation and testing directly by users? This simplifies the Jenkins usage and makes things more explicit.

That'll extra work. Remember that in our release pipeline we define the exact expected version (1.19.0-RC3) and test for that.



> Should we roll these mysql systests into the nightly release pipeline instead of as a stand-alone job?

Now that we have more CI resources that is an option. Not sure why exactly they were separate jobs but I suspect it was CI resources. Note they're on a weekly schedule rather than daily.

ehelms:
I don't entirely follow why they are extra work since we explicitly define them for other CI pipelines. I can take your word for it.

Issue #819, opened 2018-08-21
hammer_devel fails with error "gem hammer_cli_foreman (>= 0) more than once"


kgaikwad:
Error log - 
```
Your Gemfile lists the gem hammer_cli_foreman (>= 0) more than once.
    You should probably keep only one of them.
    While it's not a problem now, it could cause errors if you change the version of one of them later.
  
    [!] There was an error parsing `Gemfile`: You cannot specify the same gem twice coming from different sources.
    You specified that hammer_cli_foreman (>= 0) should come from source at `.` and source at `../`
    . Bundler cannot continue.
  
```

In [hammer_install.yml](https://github.com/theforeman/forklift/blob/master/roles/hammer_devel/tasks/hammer_install.yml#L41) task, to add local gems list under `~/hammer-cli-foreman/Gemfile.local`, 
it takes value of `hammer_devel_local_gems` and in `99-local.yaml.example` file that [list](https://github.com/theforeman/forklift/blob/master/vagrant/boxes.d/99-local.yaml.example#L38) contains  `gem 'hammer_cli_foreman'`  itself which is causing above failure.

Removing `gem 'hammer_cli_foreman'`  from hammer_devel_local_gems in `99-local.yaml.example` file solves this issue.

PR #828, opened 2018-08-28
Remove most default vars from example hammer box

akofink:
There's really no good reason for new users to know about and change these removed settings. It works great with the defaults, and those options are really targeted toward more specialized circumstances (i.e. if you're developing a brand new hammer plugin). 

Fixes https://github.com/theforeman/forklift/issues/819

### DIFF
--- a/roles/hammer_devel/defaults/main.yml
+++ b/roles/hammer_devel/defaults/main.yml
@@ -7,6 +7,7 @@ hammer_devel_server_port: 443
 hammer_devel_host: "{{ hammer_devel_server_protocol }}://{{ hammer_devel_server }}:{{ hammer_devel_server_port }}"
 hammer_devel_username: admin
 hammer_devel_password: changeme
+hammer_devel_github_fork_remote_name: "{{ hammer_devel_github_username }}"
 hammer_devel_repositories:
   - theforeman/hammer-cli
   - theforeman/hammer-cli-foreman
--- a/vagrant/boxes.d/99-local.yaml.example
+++ b/vagrant/boxes.d/99-local.yaml.example
@@ -27,16 +27,7 @@ centos7-hammer-devel:
     group: 'hammer-devel'
     variables: # defaults in roles/hammer_devel/defaults/main.yml
       hammer_devel_github_username: <REPLACE ME>
-      hammer_devel_github_fork_remote_name: origin
-      hammer_devel_server: centos7-devel.custom-domain
-      hammer_devel_repositories:
-        - theforeman/hammer-cli
-        - theforeman/hammer-cli-foreman
-        - katello/hammer-cli-katello
-      hammer_devel_local_gems: |-
-        gem 'hammer_cli'
-          gem 'hammer_cli_foreman'
-          gem 'hammer_cli_katello'
+      #hammer_devel_server: centos7-devel.custom-domain
 
 katello-remote-execution:
   box: centos7-katello-nightly


omkarkhatavkar:
vagrant up centos7-hammer-devel is failed with below error. 
```
fatal: [centos7-hammer-devel]: FAILED! => changed=false 
  cmd: /home/vagrant/bin/bundle install
  msg: ''
  rc: 4
  stderr: ''
  stderr_lines: []
  stdout: |-
    Your Gemfile lists the gem hammer_cli_foreman (>= 0) more than once.
    You should probably keep only one of them.
    While it's not a problem now, it could cause errors if you change the version of one of them later.
  
    [!] There was an error parsing `Gemfile`: You cannot specify the same gem twice coming from different sources.
    You specified that hammer_cli_foreman (>= 0) should come from source at `.` and source at `../`
    . Bundler cannot continue.
  
     #  from /home/vagrant/hammer-cli-foreman/Gemfile:20
     #  -------------------------------------------
     #    local_gemfile = File.join(File.dirname(__FILE__), file_name)
     >    self.instance_eval(Bundler.read_file(local_gemfile)) if File.exist?(local_gemfile)
     #  end
     #  -------------------------------------------
  stdout_lines: <omitted>
	to retry, use: --limit @/root/.ansible/retry-files/hammer_devel.retry

centos7-hammer-devel       : ok=14   changed=10   unreachable=0    failed=1   

```

After fix metioned tried [ vagrant provision centos7-hammer-devel ] - resovles the issue. 

ACK

johnpmitsch:
Its a little confusing since you are saying "Don't use these unless you want to change the defaults", but then you have to update `hammer_devel_github_username` iiuc

akofink:
Updated to move the comment below the required `hammer_devel_github_username` variable.

akofink:
Kind of re-tooled this PR completely to remove confusion :P

johnpmitsch:
@akofink nice changes, simple is good :+1: if you could add an explicit note in the docs that this box by default points to `centos7-devel`  server and you can override this with the `hammer_devel_server` variable, I think that would be helpful to many :)

akofink:
I don't want to do that, as I feel that having the variable in the boxes example file *and* an explicit comment as to where the default variables reside is sufficient. This is already more hand holding than most other boxes in the example file. I'd honestly prefer that we remove the server variable from the example file, but I can see the merit of explicitly showing users this variable, as it's probably a commonly overridden one.

For some context, consider the centos7-devel box. It uses [many roles](https://github.com/theforeman/forklift/blob/master/playbooks/devel.yml#L13) with even more default variables across the many roles, only [three of which](https://github.com/theforeman/forklift/blob/master/vagrant/boxes.d/99-local.yaml.example#L8) we document in the example boxes file. Way more people use this box than the hammer-devel box.

ekohl:
@akofink have you considered keeping them in but commented? that shows you have sane defaults that are easily overridden.

akofink:
Now there's a $1M idea! lol why didn't we think of that?

johnpmitsch:
@akofink The commented out default vars are good, nice addition. Throwing in a commented out `hammer_devel_github_fork_remote_name` might be nice too, but not necessary or anything.

I really don't see the harm of mentioning the default dev server and how it can be overriden in the docs, I can only say it wasn't obvious to me as a user (it is now in hindsight). In general, I think explicit documentation that has few assumptions is the least confusing for users.

I'm fine with the changes as is, I just am throwing in my 2 cents as a user who hadn't done much hammer development before :smile:

johnpmitsch:
nvm my comment about `hammer_devel_github_fork_remote_name`, I see its the github username now :+1:

akofink:
I definitely appreciate that perspective, as we don't get new users very often. I thought I agreed with your statement:

> In general, I think explicit documentation that has few assumptions is the least confusing for users.

That is, until today when the extra docs caused more harm than good. I guess that's the "in general" qualifier. :) Thanks for the review!

ekohl:
This fixes https://github.com/theforeman/forklift/issues/819 as well, right?

akofink:
Yes!

PR #830, opened 2018-08-31
Use 'yum clean all' for foreman_repositories

ehelms:


ehelms:
I ran into an issue where `expire-cache` didn't fully handle all upgrade cases and the 3.8 pipeline failed without a full yum clean all.

ekohl:
Weird but I guess it's good enough. Mind adding `args: warn: false` so suppress the warning?

ehelms:
@ekohl Updated

PR #832, opened 2018-09-05
Add test to check for no old SCL packages

ehelms:


ehelms:
This may not be the cleanest bash for the test, but it worked within the confines of bats.

bats/fb-verify-packages.bats
```diff
@@ -0,0 +1,23 @@
...
+set -o pipefail
+
+load os_helper
+load foreman_helper
+
+@test "verify no old SCL packages" {
+  run tSetOSVersion
+  run FOREMAN_VERSION=$(tForemanVersion)
+
+  if tIsRedHatCompatible || tIsCentOSCompatible; then
```
ekohl: CentOS is considerd RH compatible since `/etc/redhat-release` exists there as well.

bats/fb-verify-packages.bats
```diff
@@ -0,0 +1,23 @@
...
+    if [ $FOREMAN_VERSION == '1.20' ];then
+      if [ `yum list installed | grep rh-ruby24 | wc -l` > 0 ]; then
```
ekohl: This can just be written as:
```bash
yum list installed | grep -q -E 'rh-ruby24|tfm-ror51'
```

bats/fb-verify-packages.bats
```diff
@@ -0,0 +1,23 @@
...
+
+load os_helper
+load foreman_helper
+
+@test "verify no old SCL packages" {
+  run tSetOSVersion
+  run FOREMAN_VERSION=$(tForemanVersion)
+
+  if tIsRedHatCompatible || tIsCentOSCompatible; then
+    if [ $FOREMAN_VERSION == '1.20' ];then
```
ekohl: Would it make sense to write this `[[ $(echo "$FOREMAN_VERSION >= 1.20" | bc) == 1 ]]`

ehelms: What is `bc` ? My environment does not recognize it

ekohl: https://www.gnu.org/software/bc/bc.html

bats/fb-verify-packages.bats
```diff
@@ -0,0 +1,20 @@
...
+load os_helper
+load foreman_helper
+
+@test "verify no old SCL packages" {
+  run tSetOSVersion
+  run FOREMAN_VERSION=$(tForemanVersion)
+
+  if tIsRedHatCompatible; then
+    if [[ $(echo "$FOREMAN_VERSION >= 1.20" | bc) == 1 ]]; then
+      if [ `yum list installed | grep -q -E 'rh-ruby24|tfm-ror51' | wc -l` > 0 ]; then
```
ekohl: Ah, I now see why the counting is needed. Perhaps we should use `grep -c` instead?

bats/fb-verify-packages.bats
```diff
@@ -0,0 +1,20 @@
...
+load os_helper
+load foreman_helper
+
+@test "verify no old SCL packages" {
+  run tSetOSVersion
+  run FOREMAN_VERSION=$(tForemanVersion)
+
+  if tIsRedHatCompatible; then
+    if [ $FOREMAN_VERSION == '1.20' ]; then
+      if [ `yum list installed | grep -q -E 'rh-ruby24|tfm-ror51'` ]; then
```
ehelms: @ekohl This actually worked, I just needed to keep it still wrapped in the if

bats/fb-verify-packages.bats
ekohl: Sounds OK for now.

bats/fb-verify-packages.bats
ekohl: I guess the subshell is needed because of the pipe

ehelms:
Good call, added to to default test list

PR #837, opened 2018-09-26
certs-tar rename

ekohl:


pipelines/pipeline_katello_nightly.yml
```diff
@@ -26,7 +26,10 @@
     foreman_installer_skip_installer: true
     foreman_installer_additional_packages:
       - foreman-installer-katello
+    koji_task_ids:
+      - 135610
```
ehelms: Needs dropped

pipelines/pipeline_katello_nightly.yml
```diff
@@ -26,7 +26,10 @@
...
   roles:
+    - koji
```
ehelms: Needs dropped

playbooks/foreman_proxy_content.yml
```diff
@@ -17,14 +17,14 @@
...
           - '--disable-system-checks
+            --certs-tar-file "{{ foreman_proxy_content_certs_tar }}"
             --foreman-proxy-content-enable-ostree true
             --foreman-proxy-trusted-hosts "{{ server_fqdn.stdout }}"
             --foreman-proxy-trusted-hosts "{{ ansible_nodename }}"
             --foreman-proxy-foreman-base-url "https://{{ server_fqdn.stdout }}"
             --foreman-proxy-register-in-foreman true
             --foreman-proxy-oauth-consumer-key "{{ oauth_consumer_key.stdout }}"
             --foreman-proxy-oauth-consumer-secret "{{ oauth_consumer_secret.stdout }}"
-            --foreman-proxy-content-certs-tar "{{ foreman_proxy_content_certs_tar }}"
```
ehelms: This will now fail for all versions except for nightly, I think we need a version check here to use this parameter in some cases.

ekohl:
Updated. I thought I already removed that from it. I'm leaning to merging it now to unblock the pipeline and look at older versions after that.

ehelms:
The problem lies in that it will break those older version release pipelines. Double edged sword. Either fix nightly but break ability to release 3.9/3.10 or keep them able to release and keep nightly broken. The compromise is to wait until I get 3.9.1 and 3.10 RC1 released and then merge this and then fix it properly.

ekohl:
The current approach is untested, but I think this provides a framework for more future renames

ehelms:
I hit the following before the most recent code push update:

```
TASK [foreman_installer : include_tasks] ***************************************
fatal: [centos7-katello-3.8]: FAILED! => 
  reason: |-
    Syntax Error while loading YAML.
      did not find expected key
  
    The error appears to have been in '/home/ehelms/workspace/forklift/roles/foreman_installer/tasks/install.yml': line 18, column 71, but may
  
      set_fact:
        foreman_installer_options: "{{ foreman_installer_options | join(" ") }}"
                                                                          ^ here
    We could be wrong, but this one looks like it might be an issue with
    missing quotes.  Always quote template expression brackets when they
    start a value. For instance:
  
        with_items:
          - {{ foo }}
  
    Should be written as:
  
        with_items:
          - "{{ foo }}"
```

ehelms:
With the update, I am getting passing builds.

roles/foreman_installer/tasks/install.yml
```diff
@@ -4,12 +4,29 @@
     msg: 'foreman_installer_options must be an array'
   when: foreman_installer_options is string
 
+
```
ehelms: Extra empty line here.

ehelms:
On upgrade:

```
    An unhandled exception occurred while templating ''[ ''--disable-system-checks'', ''--upgrade'', ''--certs-update-all'' ] + {{ foreman_installer_options_internal_use_only }}''. Error was a <class ''ansible.errors.AnsibleError''>, original message: recursive loop detected in template string: [ ''--disable-system-checks'', ''--upgrade'', ''--certs-update-all'' ] + {{ foreman_installer_options_internal_use_only }}'
```

ekohl:
Updated

PR #839, opened 2018-09-28
Add foreman_client_repositories role and switch to using them

ehelms:


ekohl:
Won't this break all the old (released) pipelines? Some of them use the staging repos of their versions

roles/foreman_client_repositories/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: 'Foreman Client Staging Repository'
+  yum_repository:
+    name: foreman-client-koji
+    description: "Foreman {{ foreman_client_repositories_version }} Client Koji Repository"
+    baseurl: "http://koji.katello.org/releases/yum/foreman-client-{{ foreman_client_repositories_version }}/el{{ ansible_distribution_major_version }}/x86_64/"
+    priority: 1
```
zjhuntin: For my own knowledge what does "Enforce ordered protection of repositories" mean? Is this the order in which yum would go through repositories to find a particular package?

ehelms:
I added a conditional and kept the old code to try to limit the scope and use of it to Katello 3.8 or less. @ekohl if you dont mind giving that a once over

roles/foreman_client_repositories/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+    gpgcheck: 0
+  when: foreman_client_repositories_environment == "staging"
+
+- name: 'Setup Foreman Client Release Repository'
+  yum:
+    name: https://fedorapeople.org/groups/foreman/releases/yum/{{ foreman_client_repositories_version }}/client/el{{ ansible_distribution_major_version }}/x86_64/foreman-client-repos-latest.rpm
```
ekohl: Is it allowed to have multiple keys with the same name? Doesn't look intended though.

roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -43,3 +43,4 @@
     baseurl: "http://koji.katello.org/releases/yum/katello-{{ katello_repositories_version }}/client/el{{ ansible_distribution_major_version }}/x86_64/"
     priority: 1
     gpgcheck: 0
+  when: katello_repositories_version != 'nightly' or katello_repositories_version < 3.9
```
ekohl: The only thing I wonder about is if the type is correct. What if it's a string instead of a float?

roles/foreman_client_repositories/tasks/main.yml
ehelms: Yep

ehelms:
@ekohl I droppe dthe duplicate name and added a convert to float to make the version comparison on equal footing

PR #841, opened 2018-10-04
Use Ruby SCL instead of RVM in devel

ekohl:
Also updates the default ruby version to 2.5 to match RPM packaging.

### DIFF
--- a/roles/katello_devel/meta/main.yml
+++ b/roles/katello_devel/meta/main.yml
@@ -15,5 +15,4 @@ dependencies:
     foreman_installer_options_internal_use_only:
       - "--disable-system-checks"
       - "--katello-devel-scl-ruby={{ ruby_scl_version }}"
-      - "--katello-devel-enable-ostree=true"
       - "--katello-devel-admin-password {{ foreman_installer_admin_password }}"
--- a/roles/katello_devel/meta/main.yml
+++ b/roles/katello_devel/meta/main.yml
@@ -5,6 +5,7 @@ dependencies:
     foreman_server_repositories_ostree: true
     foreman_server_repositories_katello: true
     foreman_server_repositories_foreman_client: true
+  - role: ruby_scl
   - role: foreman_installer_devel_scenario
   - role: foreman_installer
     katello_repositories_version: nightly
@@ -13,5 +14,6 @@ dependencies:
       - foreman-installer-katello
     foreman_installer_options_internal_use_only:
       - "--disable-system-checks"
+      - "--katello-devel-scl-ruby={{ ruby_scl_version }}"
       - "--katello-devel-enable-ostree=true"
       - "--katello-devel-admin-password {{ foreman_installer_admin_password }}"
--- a/roles/foreman_installer_devel_scenario/files/katello-devel.yaml
+++ b/roles/foreman_installer_devel_scenario/files/katello-devel.yaml
@@ -5,6 +5,8 @@
 :log_name: katello-devel.log
 :log_level: DEBUG
 :no_prefix: false
+# Forklift always overwrites the answers so let's not pretend we save them
+:dont_save_answers: true
 :answer_file: /etc/foreman-installer/scenarios.d/katello-devel-answers.yaml
 :installer_dir: /usr/share/foreman-installer/katello
 :module_dirs: /usr/share/foreman-installer/modules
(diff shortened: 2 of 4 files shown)


akofink:
Actually, I mistyped, `bundle install` via ssh doesn't work properly. It's still complaining about the ruby version needing to be > 2.2.2, even though `ruby -v` returns 2.5.1 (without running scl enable first). Seems like something might be wrong with the devel environment perhaps. I'll try to retest again in a couple days.

ekohl:
Perhaps `bundle` doesn't come from the SCL? If you still have the box a `which bundle` and `bundle --version` could be interesting.

akofink:
Sorry, I already destroyed it. :/

ekohl:
This guy, so destructive

akofink:
Seems right. It seems more like the environment is different for puppet than with a standard login shell.

```
[vagrant@centos7-devel-test foreman]$ which bundle
/opt/rh/rh-ruby25/root/usr/bin/bundle
[vagrant@centos7-devel-test foreman]$ bundle --version
Bundler version 1.16.1
```

ekohl:
I think that puppet doesn't evaluate `/etc/profile.d` while running commands, which makes sense and IMHO the biggest pain point in SCLs. It could also be related to the explicit setting of the `path`. We may need to introduce a `scl_ruby` parameter.

akofink:
Yeah, I was thinking the same. We have `$use_rvm` and `$rvm_ruby`, so I think it'd be good to also have the same for SCL.

ehelms:
I think I'd like someone to use this in a fresh devel environment for a week or so to ensure there are no gotchas that will degrade developer experience.

akofink:
I'll be glad to be the guinea pig, once we get the puppet-katello_devel changes to make it work right.

akofink:
puppet-katello_devel PR https://github.com/theforeman/puppet-katello_devel/pull/178

akofink:
Hmm... not working for some reason: `Unable to open /etc/scl/conf/bundle`

ehelms:
@akofink How goes the piloting use of this?

akofink:
As I've previously stated, we need to figure out how to use SCLs in general with puppet. [This PR](https://github.com/theforeman/puppet-katello_devel/pull/178) does not do it right.

iNecas:
rebase needed

akofink:
@ekohl would you mind trying this out one last time with https://github.com/theforeman/puppet-katello_devel/pull/178? It's working fine for me, but @johnpmitsch was seeing issues.

johnpmitsch:
@akofink @ekohl I missed there were two PRs.... :see_no_evil: retesting again now!

johnpmitsch:
though its hard to test with the conflicts, @ekohl could you rebase this?

playbooks/devel.yml
```diff
@@ -4,16 +4,21 @@
...
     foreman_server_repositories_ostree: true
-    katello_repositories_version: nightly
-    katello_repositories_environment: staging
+    foreman_repositories_version: nightly
+    foreman_repositories_environment: staging
     foreman_installer_scenario: katello-devel
     foreman_installer_additional_packages:
       - foreman-installer-katello
+    foreman_installer_module_prs:
+      - theforeman/katello_devel/178
```
akofink: I guess this should be removed.

roles/katello_devel/meta/main.yml
```diff
@@ -13,5 +14,8 @@ dependencies:
       - foreman-installer-katello
     foreman_installer_options_internal_use_only:
       - "--disable-system-checks"
+      - "--katello-devel-use-rvm=false"
+      - "--katello-devel-use-scl-ruby=true"
```
johnpmitsch: Do we need these options? Aren't they the defaults?

akofink:
https://community.theforeman.org/t/incoming-change-scl-ruby-by-default-in-forklift-centos7-devel-box/13181

roles/katello_devel/meta/main.yml
ekohl: I thought about this as well, but I assumed we stored answers for the devel scenario. Now it looks like forklift always overwrites answers so that's pointless. We might as well add the option `dont_save_answers: true` to remove any illusions they persist.

ekohl:
Updated to better match the defaults

johnpmitsch:
giving it a test!

PR #842, opened 2018-10-08
Only run koji tasks when tasks are given & always include koji role

ekohl:
This makes it very easy to include koji tasks into a pipeline.


evgeni:
Two thoughts here:

* How about making `koji` a dependency of the `foreman_repositories` role? I'd assume nobody would try to use forklift to test smth w/o using that role. :)
* I am a huge non-fan of `block`, mostly because it makes reading the YAML harder (with the condition being at the end etc). You can [use conditionals in the `meta.yaml`](https://docs.ansible.com/ansible/devel/user_guide/playbooks_conditionals.html#applying-when-to-roles-imports-and-includes) or [when using `include_role`](https://docs.ansible.com/ansible/devel/user_guide/playbooks_reuse_roles.html#using-roles). You could also move all the real work to `real_tasks.yaml` and `include` that with a proper condition, but I am not sure this is nice code.

ekohl:
I didn't get a chance to test this yet but I think this is cleaner.

evgeni:
nah, I'm good

PR #843, opened 2018-10-14
Allow testing a puppet module branch

ehelms:
This also re-factors the module PR handling to use more native Ansible
and moves the parameter to a list.

### DIFF
--- a/roles/foreman_installer/tasks/module_prs.yml
+++ b/roles/foreman_installer/tasks/module_prs.yml
@@ -1,0 +1,40 @@
+---
+- name: "Determine if Puppet 4 is used"
+  stat:
+    path: /opt/puppetlabs/puppet/bin/gem
+  register: puppetlabs_gem
+
+- name: "Install gems to rebuild the kafo module cache"
+  gem:
+    name: "{{ item }}"
+    executable: /opt/puppetlabs/puppet/bin/gem
+  with_items:
+    - puppet-strings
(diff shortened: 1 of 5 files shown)


roles/foreman_installer/tasks/module_prs.yml
```diff
@@ -0,0 +1,42 @@
...
+  register: puppetlabs_gem
+
+- name: "Install gems to rebuild the kafo module cache"
+  gem:
+    name: "{{ item }}"
+    executable: /opt/puppetlabs/puppet/bin/gem
+  with_items:
+    - yard
+    - puppet-strings
+    - rgen # https://tickets.puppetlabs.com/browse/PDOC-168
```
ekohl: This is no longer needed since puppet-strings depends on it.

roles/foreman_installer/tasks/module_prs.yml
```diff
@@ -0,0 +1,42 @@
...
+  stat:
+    path: /opt/puppetlabs/puppet/bin/gem
+  register: puppetlabs_gem
+
+- name: "Install gems to rebuild the kafo module cache"
+  gem:
+    name: "{{ item }}"
+    executable: /opt/puppetlabs/puppet/bin/gem
+  with_items:
+    - yard
```
ekohl: This is no longer needed since puppet-strings depends on it.

ehelms:
@ekohl Updated

PR #846, opened 2018-10-17
Add 1.20 box configs

ehelms:


evgeni:
@ehelms you want to pin the cop, or fix the offences (in a second PR)?

PR #851, opened 2018-10-18
Switch to pulp nightly

johnpmitsch:
For Katello 2.10 development, we are using pulp's nightly builds
until the beta is released.

### DIFF
--- a/roles/katello_repositories/tasks/staging_repos.yml
+++ b/roles/katello_repositories/tasks/staging_repos.yml
@@ -20,13 +20,25 @@
     priority: 1
     gpgcheck: 0
 
-- name: Add Pulp repository
+- name: Add Pulp Stable repository
   yum_repository:
     name: pulp-repository
     description: Pulp release repository
-    baseurl: "https://repos.fedorapeople.org/repos/pulp/pulp/{{ katello_repositories_pulp_release }}/{{ katello_repositories_pulp_version }}/7Server/x86_64/"
+    baseurl: "https://repos.fedorapeople.org/repos/pulp/pulp/{{ katello_repositories_pulp_release }}/{{ katello_repositories_pulp_version }}/$releasever/$basearch"
     gpgcheck: no
     enabled: yes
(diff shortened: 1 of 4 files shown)


roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -20,13 +20,25 @@
...
     gpgcheck: no
     enabled: yes
+  when:
+    - katello_repositories_pulp_version != "nightly"
+
+- name: Add Pulp Nightly repository
+  yum_repository:
+    name: pulp-repository
+    description: Pulp nightly repository
+    baseurl: "https://repos.fedorapeople.org/repos/pulp/pulp/testing/automation/2-master/stage/7Server/x86_64/"
```
ekohl: I'd prefer using the `$releasever` and `$arch` here instead of hardcoding but it looks like that wasn't done above either.

johnpmitsch: do you mean create new ansible variables for them? I'm not sure where I would be getting them from

ekohl: No, yum supports this natively. I do see it's `$basearch` and not `$arch`. From https://repos.fedorapeople.org/repos/pulp/pulp/rhel-pulp.repo
```
baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/stable/2/$releasever/$basearch/
```
It should be noted that on RHEL expands it to `7Server` where CentOS has `7`, but on pulp they handle this on the server side.

ehelms:
@johnpmitsch The packaging change has been merged, does this need updates before I merge?

roles/katello_repositories/tasks/staging_repos.yml
johnpmitsch: @ekohl updated

johnpmitsch:
@ehelms I just updated with @ekohl's suggestion

PR #853, opened 2018-10-23
allow empty (nil) box names

evgeni:
this is useful if you want to create a box from scratch, w/o any base
box, like you would do if you PXE boot the box


rplevka:
ACK!

PR #859, opened 2018-10-29
don't install/update foreman-release-scl in the installer role

evgeni:
it is already handled in the main tasks of foreman_repositories


evgeni:
https://github.com/theforeman/forklift/blob/f53c704c378bcc86dd1c7c72f1b918ff7808040d/roles/foreman_repositories/tasks/main.yml#L4-L11

evgeni:
this change allows to use the `foreman_installer` role with `foreman_installer_upgrade=True` in scenarios where `foreman_repositories` is not used at all (like when you have a local repo and similar)

PR #860, opened 2018-10-30
Fix Katello upgrade nightly pipeline

ehelms:


jlsherrill:
can you explain why this fixes it?

ehelms:
Otherwise it interprets this as a float, which gets translated to '1.2'

PR #861, opened 2018-11-01
Ensure proxy downloaded package is an RPM

ehelms:


ehelms:
@evgeni @jlsherrill Note that I had to make another change from curl to wget. For the life of me, I could not get curl to work in this scenario whereas wget worked like a charm.

evgeni:
What was the issue with curl? But as long it works I don't mind either way.

ehelms:
The issue with curl is it never downloaded the package. I even tried `-o walrus-0.71-1.noarch.rpm` and it would just download nothing. Happy to have someone else figure it out but this worked for me.

evgeni:
Ah! curl doesn't follow redirects by default. You need to pass `-L` for that :)

ehelms:
@evgeni Take a look, I moved back to curl

PR #863, opened 2018-11-02
allow to set ${provider}_options in settings.yaml

evgeni:
this allows to provide default configuration for *all* boxes of a
specific provider


evgeni:
cop fixes in https://github.com/theforeman/forklift/pull/864

PR #866, opened 2018-11-05
 explicitly set the pulp release in the pipeline

evgeni:
otherwise we allways test against pulp nightlies, and that seems wrong


evgeni:
@ekohl definitely

evgeni:
@ekohl done, mind re-reviewing?

PR #869, opened 2018-11-06
Always set openstack settings

ekohl:


tstrych:
ACK

vagrant/lib/forklift/box_distributor.rb
```diff
@@ -272,8 +272,8 @@ def configure_rackspace(machine, box)
         override.vm.box  = 'dummy'
         p.server_name    = machine.vm.hostname
         p.flavor         = /4GB/
-        p.image          = box.fetch('image_name')
-        override.ssh.pty = true if box.fetch('pty')
+        p.image          = box['image_name']
+        override.ssh.pty = true if box['pty']
```
ehelms: We want to allow these to be nil? If so, I'd prefer that being more explicit `box.fetch('pty', nil)`

ekohl: I'm not that familiar with Ruby so I'll trust you on it. For the openstack provider it's important that the box is always `nil` because otherwise `vagrant status` will attempt to fetch it and things fail. After that it might be broken when you launch the particular box, but they shouldn't be in blocking the other boxes.

ekohl:
@ehelms updated

PR #875, opened 2018-11-19
Drop 3.6 from 3.8 upgrade testing

ehelms:
No more releases will be made to 3.6, and 3.6 fails due to Candlepin
selinux issues.


zjhuntin:
ACK

PR #880, opened 2018-11-22
Simplifying our forklift testing pipeline definitions

evgeni:
this is based on the [Simplifying our forklift pipeline definitions RFC](https://community.theforeman.org/t/simplifying-our-forklift-pipeline-definitions/11692)

PR #881, opened 2018-11-23
split bats tests into four stages: setup, tests, additional and teardown

evgeni:
this allows most pipelines to just set "additional" instead of overriding the whole test list

### DIFF
--- a/pipelines/pipeline_foreman_nightly.yml
+++ b/pipelines/pipeline_foreman_nightly.yml
@@ -23,6 +23,7 @@
       - "fb-verify-packages.bats"
       - "fb-test-foreman.bats"
       - "fb-test-puppet.bats"
+    bats_teardown:
       - "fb-finish.bats"
   roles:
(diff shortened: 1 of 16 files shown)


evgeni:
this is related to #880, but can work on it's own

ehelms:
This has already been merged but find I this design a bit hard to grok. For one, we don't label our bats tests, there is some implied ordering but theres not really a concept of setup and teardown bats tests so I think the bats tests need updating to reflect this implication. Second, what is the point of bats_additional?  I find this terminology confusing when I consider there is setup, tests, teardown.

evgeni:
The idea behind `additional` was to allow to run *additional* tests (hah!) without overriding the main set of the tests. Like in the diff, adding the Proxy tests, that we only execute when we actually have deployed a external proxy.

And for not labeling, true. The `teardown` section was really only to allow to add `additional` between the main tests and destroying everything.

PR #882, opened 2018-11-25
make some task/role optional

bmx0r:
Here is a small workaround that i add to ease the deploy of a production box
I figure out that i didn't need the following:
- Atomic repo
- puppet-strings gem
- the hardcoded vagrant user in Ansible config (i believe vagrant handle that himself, and it is not necessary)

This is probably not the right way to do it, feel free to propose something so I can refactor it the right way
Regards
Mike

### DIFF
--- a/roles/foreman_installer/tasks/module_prs.yml
+++ b/roles/foreman_installer/tasks/module_prs.yml
@@ -10,7 +10,7 @@
     executable: /opt/puppetlabs/puppet/bin/gem
   with_items:
     - puppet-strings
-  when: puppetlabs_gem.stat.exists
+  when: ( puppetlabs_gem.stat.exists and determine_puppetlabs_gem_check is not defined )
   tags:
     - packages
 
--- a/roles/foreman_installer/tasks/module_prs.yml
+++ b/roles/foreman_installer/tasks/module_prs.yml
@@ -10,7 +10,7 @@
     executable: /opt/puppetlabs/puppet/bin/gem
   with_items:
     - puppet-strings
-  when: ( puppetlabs_gem.stat.exists and determine_puppetlabs_gem_check is not defined )
+  when: puppetlabs_gem.stat.exists
   tags:
     - packages
 
--- a/roles/ostree_repositories/tasks/main.yml
+++ b/roles/ostree_repositories/tasks/main.yml
@@ -5,4 +5,4 @@
     description: Centos Atomic Build Repository
     baseurl: http://buildlogs.centos.org/centos/7/atomic/x86_64/Packages/
     gpgcheck: no
-  when: no_ostree_repo is not defined
+    enabled : "{{ ostree_repositories_enabled }}"
--- a/roles/ostree_repositories/tasks/main.yml
+++ b/roles/ostree_repositories/tasks/main.yml
@@ -5,3 +5,4 @@
     description: Centos Atomic Build Repository
     baseurl: http://buildlogs.centos.org/centos/7/atomic/x86_64/Packages/
     gpgcheck: no
+  when: no_ostree_repo is not defined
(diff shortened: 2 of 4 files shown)


roles/foreman_installer/tasks/module_prs.yml
```diff
@@ -10,7 +10,7 @@
     executable: /opt/puppetlabs/puppet/bin/gem
   with_items:
     - puppet-strings
-  when: puppetlabs_gem.stat.exists
+  when: ( puppetlabs_gem.stat.exists and determine_puppetlabs_gem_check is not defined )
```
ekohl: I'd prefer to exclude the whole file if it isn't needed. I don't see where `determine_puppetlabs_gem_check` would come from.

roles/ostree_repositories/tasks/main.yml
```diff
@@ -5,3 +5,4 @@
     description: Centos Atomic Build Repository
     baseurl: http://buildlogs.centos.org/centos/7/atomic/x86_64/Packages/
     gpgcheck: no
+  when: no_ostree_repo is not defined
```
ekohl: It'd be better to solve this at a higher level and not include the role when not it's not needed

roles/foreman_installer/tasks/module_prs.yml
bmx0r: for now it is just two variables that i set when calling the playbook
````
  import_playbook: 'forklift/playbooks/katello.yml'
  vars:
          determine_puppetlabs_gem_check: true
          no_ostree_repo: true
````

File: roles/ostree_repositories/tasks/main.yml (hunk 2 above)
Comment (bmx0r) [reply roles/ostree_repositories/tasks/main.yml]: I get that, but how :)
in katello/meta/main.yml
```
dependencies:
...
  - role: ostree_repositories
```
is there some conditional dependencies in ansible?

ekohl: An alternative would be to set the repo to enabled/disabled. That means you still include the role, but also disable any potentially active repo. It'd probably be named something like `ostree_repositories_enabled` and default to `yes`, allowing you to override it with `no` to disable.

roles/foreman_installer/tasks/module_prs.yml
ehelms: Why do you find you encounter this? This task file should only be included if oyu are specifying PRs or special branches for testing per:

https://github.com/theforeman/forklift/blob/4612ea43361e61f94ca75b3dba8f843997944578/roles/foreman_installer/tasks/main.yml#L7-L8

roles/ostree_repositories/tasks/main.yml
ehelms: Out of curiosity, where are you getting ostree from if not this repository? That may help us craft the solution.

roles/foreman_installer/tasks/module_prs.yml
ekohl: Probably because they were changed to arrays so the check no longer works as intended.

ekohl: https://github.com/theforeman/forklift/pull/885

roles/ostree_repositories/tasks/main.yml
bmx0r: well, I might have miss something, but i disable it and i manage to install a working katello server. I deploy using kickstart, and i do not see where the centos atomic repo could add value to my setup.
Regards
Mike

evgeni: since you initially brought that up, we've merged support for a meta role that sets up the repositories: `foreman_server_repositories` and you can turn off OSTree by passing `foreman_server_repositories_ostree: false` in your playbook now. Does that fix it for you?

ehelms:
I believe the goal of this PR has been satisfied. Please re-open if you feel otherwise.

PR #884, opened 2018-11-26
Make version compare handle nightly

parthaa:
The prior code for 'Katello Client Koji repository' did not check for
the nightly causing the following error
""""
The conditional check
'katello_repositories_version is version('3.8', '<=')'
failed. The error was:
Version comparison: '<' not supported between instances of
'str' and 'int'
""""

### DIFF
--- a/roles/katello_repositories/tasks/staging_repos.yml
+++ b/roles/katello_repositories/tasks/staging_repos.yml
@@ -47,4 +47,6 @@
     baseurl: "http://koji.katello.org/releases/yum/katello-{{ katello_repositories_version }}/client/el{{ ansible_distribution_major_version }}/x86_64/"
     priority: 1
     gpgcheck: 0
-  when: katello_repositories_version is version('3.8', '<=')
+  when:
+    - katello_repositories_version != "nightly"
+    - katello_repositories_version is version('3.8', '<=')


evgeni:
Which version of ansible is that? Mine (2.7) parsed nightly fine.

parthaa:
```
$ rpm -q ansible
ansible-2.7.2-1.fc29.noarch
```

parthaa:
> Mine (2.7) parsed nightly fine.

I don't see whats special in my env to cause this issue. I just know the vagrant up step fails without this change..
```
Fatal: [my-dev-machine]: FAILED! =>
  msg: |-
    The conditional check 'katello_repositories_version is version('3.8', '<=')' failed. The error was: Version comparison: '<' not supported between instances of 'str' and 'int'
 
    The error appears to have been in '/home/paji/vagrant/katello-deploy/roles/katello_repositories/tasks/staging_repos.yml': line 43, column 3, but may
 
    - name: 'Katello Client Koji repository'
      ^ here
        to retry, use: --limit @/home/paji/.ansible/retry-files/devel.retry
 
my-dev-machine                     : ok=16   changed=11   unreachable=0    failed=1
```

here is my machine configuration ->
```
my-dev-machine:
  primary: true
  box: centos7
  ansible:
    group: 'devel'
    variables:
      ssh_forward_agent: true
      katello_devel_github_username: parthaa
```

evgeni:
Hah! I can reproduce this when using Ansible with Python3, but not with Python2.

ACK on the workaround for now, but IMHO this is an Ansible bug.

evgeni:
Actually, a Python bug: https://bugs.python.org/issue14894

PR #890, opened 2018-12-12
Allow rebuilding forklift boxes

ekohl:


roles/forklift/tasks/rebuild.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: 'vagrant {{ item }} boxes'
```
evgeni: you can't use `{{ item }}` in the `name` as that is generated outside of the loop. so propbably just `rebuild vagrant boxes` or so

ekohl: Updated

PR #892, opened 2018-12-17
Update foreman-admin password param to new initial change

chris1984:


chris1984:
@ekohl same thing happened when i did a reset and pushed up so, here is the new pr.

pipelines/pipeline_katello_310.yml
```diff
@@ -45,7 +45,7 @@
     foreman_installer_scenario: katello
     foreman_installer_options_internal_use_only:
       - "--disable-system-checks"
-      - "--foreman-admin-password {{ foreman_installer_admin_password }}"
+      - "--foreman-initial-admin-password {{ foreman_installer_admin_password }}"
```
ekohl: You don't have to modify this since we know 3.10 won't use the new parameter. This is true for most versions below. Only where we know it's nightly or could be (e.g., when the user can control the repos via variables) needs to be modified.

chris1984:
@ekohl updated the install task with a replace one, not sure if it looks right when you get a chance to look.

roles/foreman_installer/tasks/install.yml
```diff
@@ -21,6 +21,11 @@
     foreman_installer_options_joined: "{{ foreman_installer_options_joined | replace('--certs-tar-file', '--foreman-proxy-content-certs-tar') }}"
   when: foreman_installer_version_file['content'] | b64decode is version('1.21', '<')
 
+- name: 'Replace --inital-admin parms with --admin legacy params'
+  set_fact:
+    foreman_installer_options_joined: "{{ foreman_installer_options_joined | replace('--foreman-admin-password', '--foreman-admin-username', '--foreman-admin-email', '--foreman-admin-first-name', '--foreman-admin-last-name') }}"
```
ekohl: This does not what you think it does. In the previous example it replaces `--certs-tar-file` with `--foreman-proxy-content-certs-tar`. You only specify the old options.

chris1984:
@ekohl updated based on what we talked about.

PR #894, opened 2018-12-19
allow disabling vagrant plugins, even if they are installed

evgeni:


tstrych:
@evgeni  Not sure, but when you didn't define `hostmanager_enable`at all, you will always end  
[here ](https://github.com/theforeman/forklift/pull/894/files#diff-f123d7ad7c81373b2bd1b14344bfbef7R125) without using hostmanager

tstrych:
and possibly can be added to settings.yaml.example as option :)

PR #896, opened 2019-01-04
Add a devel pipeline

akofink:
It seems to [work up to the same spot at which the devel box currently fails](https://paste.fedoraproject.org/paste/a7b8qYAM6z777WcH1QHnxg). To test:

```sh
ansible-playbook pipelines/katello_pipeline.yml -e forklift_state=up -e katello_version=devel
```


pipelines/katello_pipeline.yml
```diff
@@ -25,10 +25,12 @@
     - foreman_repositories
     - katello_repositories
     - role: foreman_client_repositories
-      when: katello_version == "nightly" or katello_version is version('3.9', '>=')
+      when: katello_version == "nightly" or katello_version is number and katello_version is version('3.9', '>=')
```
akofink: Maybe this should be:

```yml
when: katello_version is not number or katello_version is version('3.9', '<')
```

pipelines/katello_pipeline.yml
```diff
@@ -25,10 +25,12 @@
...
     - disable_firewall
     - haveged
     - update_os_packages
+    - role: foreman_installer_devel_scenario
+      when: katello_version == "devel"
     - foreman_installer
```
johnpmitsch: does this run both the devel installer and the normal installer?

akofink: It does both, just like the [devel playbook](https://github.com/theforeman/forklift/blob/master/playbooks/devel.yml#L21)

akofink: The [foreman_installer_devel_scenario](https://github.com/theforeman/forklift/blob/master/roles/foreman_installer_devel_scenario/tasks/main.yml) role doesn't actually run the installer.

johnpmitsch: ah ok, didn't realize we do both

johnpmitsch:
Since this pipeline is here to test the [devel playbook](https://github.com/theforeman/forklift/blob/master/playbooks/devel.yml#L21) How does this pipeline stay in sync with the devel playbook?

akofink:
There may be some drift between the devel playbook and this pipeline in the same way that the production playbook and production pipelines need to manage roles separately; however, the roles themselves are shared, so any changes to the roles will be seen in these pipelines.

ehelms:
I suppose the devel playbook could be moved into a meta devel role and then both the playbook and pipeline both use the role thereby centralizing.

akofink:
I agree that would be a good solution, but I think we'd also need to do that for production in order to use the same pipeline code for both production and devel. It'd be a good future PR :shipit:

johnpmitsch:
>  I think we'd also need to do that for production in order to use the same pipeline code for both production and devel.

Would there be an issue with just doing this for devel to start and we can see how it works?

ehelms:
Production already follows this idea if you look at the katello playbook it
calls a metal role.

On Jan 4, 2019 2:53 PM, "John Mitsch" <notifications@github.com> wrote:

I think we'd also need to do that for production in order to use the same
pipeline code for both production and devel.

Would there be an issue with just doing this for devel to start and we can
see how it works?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/theforeman/forklift/pull/896#issuecomment-451550799>,
or mute the thread

akofink:
I'm not entirely sure how it'd look (correct me if I'm wrong @ehelms), but we'd be changing [this](https://github.com/theforeman/forklift/blob/f2d795b994c0bad83d665269de76701dcbfb0132/pipelines/katello_pipeline.yml#L17) to look something like this:


```yml
  roles:
    - katello
      when: katello_version != "devel"
    - katello_devel
      when: katello_version == "devel"
```

akofink:
I don't mind giving it shot in this PR

ehelms:
I was thinking a devel role to start

On Jan 4, 2019 3:08 PM, "Andrew Kofink" <notifications@github.com> wrote:

I don't mind giving it shot in this PR

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/theforeman/forklift/pull/896#issuecomment-451554636>,
or mute the thread

roles/katello/meta/main.yml
```diff
@@ -1,13 +1,17 @@
 ---
 dependencies:
+  - role: umask
   - role: selinux
   - role: etc_hosts
+  - role: disable_ipv6
```
ehelms: I think this was missing on purpose. @ekohl can you confirm?

johnpmitsch:
I got this message at first
```
TASK [forklift : Write box file] ***************************************************************************************************************************************************************
Wednesday 23 January 2019  11:44:15 -0500 (0:00:00.096)       0:00:01.779 ***** 
Wednesday 23 January 2019  11:44:15 -0500 (0:00:00.096)       0:00:01.778 ***** 
fatal: [localhost]: FAILED! => changed=false 
  checksum: 9b487665f9994b404035eacd1b74b16468232b05
  msg: Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!
	to retry, use: --limit @/home/jomitsch/.ansible/retry-files/katello_pipeline.retry

PLAY RECAP *************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=1   
```

But installing `libselinux-python` did fix things, so it seems to be an environment thing

(no action needed, just sharing this in case its relevant to something else)

johnpmitsch:
The pipeline ran fine, but when I ran the actual centos7-devel box, I got this
```
TASK [foreman_installer : Replace certs-tar and --inital-admin parms with legacy options] ***
Wednesday 23 January 2019  14:28:07 -0500 (0:00:00.058)       0:02:20.367 ***** 
Wednesday 23 January 2019  14:28:07 -0500 (0:00:00.058)       0:02:20.367 ***** 

TASK [foreman_installer : Run installer] ***************************************
Wednesday 23 January 2019  14:28:07 -0500 (0:00:00.075)       0:02:20.443 ***** 
Wednesday 23 January 2019  14:28:07 -0500 (0:00:00.075)       0:02:20.442 ***** 
fatal: [centos7-devel]: FAILED! => changed=true 
  cmd: foreman-installer -v --scenario katello-devel  --disable-system-checks --katello-devel-enable-ostree=true --foreman-admin-password changeme
  delta: '0:00:04.464815'
  end: '2019-01-23 19:28:12.556122'
  msg: non-zero return code
  rc: 1
  start: '2019-01-23 19:28:08.091307'
  stderr: |-
    ERROR: Unrecognised option '--foreman-admin-password'
  
    See: 'foreman-installer --help'
  stderr_lines:
  - 'ERROR: Unrecognised option ''--foreman-admin-password'''
  - ''
  - 'See: ''foreman-installer --help'''
  stdout: ''
  stdout_lines: <omitted>
	to retry, use: --limit @/home/jomitsch/.ansible/retry-files/devel.retry

centos7-devel              : ok=32   changed=21   unreachable=0    failed=1   
```

johnpmitsch:
Are you going to add a box defintition for this too?

akofink:
Ideally, we'd match this pipeline to the existing devel box definition as much as possible.

johnpmitsch:
> Ideally, we'd match this pipeline to the existing devel box definition as much as possible.

That is what the meta role is for, right?

I'm just wondering what I'm supposed to call here in `boxes` https://github.com/theforeman/foreman-infra/pull/920/files#diff-fc95dee6eb59f4021d1c6b59a15a54c9R2

akofink:
Correct - the meta role allows us to define a single role in the playbook, and use the same role for the pipeline. I think what you have there looks mostly correct (commented on the PR).

johnpmitsch:
@akofink does a `pipeline-katello-devel` box have to be defined in forklift for what I have to work? I don't really understand it enough to say, maybe @ehelms can weigh in

akofink:
Not to my knowledge - the pipeline-katello-devel box is created by the pipeline when it's run with ansible.

ehelms:
Looking good. @evgeni can you take a look at the modifications to pipelines?

pipelines/katello_pipeline.yml
```diff
@@ -15,18 +15,10 @@
...
-    - disable_ipv6
-    - koji
-    - foreman_server_repositories
-    - role: foreman_client_repositories
-      when: katello_version == "nightly" or katello_version is version('3.9', '>=')
-    - disable_firewall
-    - haveged
-    - update_os_packages
-    - foreman_installer
+    - role: katello
```
evgeni: the `katello` (or rather `foreman`) role does not include `disable_firewall`, `disable_ipv6` and `update_os_packages` (and they shouldn't). but we should keep them here, imho.

they also do not include `foreman_client_repostitories`. not sure about that.

roles/katello_devel/meta/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+dependencies:
+  - role: etc_hosts
+  - role: epel_repositories
```
evgeni: should these be `foreman_server_repositories` these days?

roles/katello_devel/meta/main.yml
```diff
@@ -0,0 +1,19 @@
...
+  - role: ostree_repositories
+  - role: puppet_repositories
+  - role: foreman_repositories
+  - role: katello_repositories
+  - role: katello_client_repositories
```
evgeni: or `foreman_client_repositories`?

evgeni:
> Looking good. @evgeni can you take a look at the modifications to pipelines?

Commented, nothing serious :)

akofink:
Updated! I tested both the devel pipeline and the devel box, and both came up without errors. Let me know what you find!

pipelines/vars/katello_devel.yml
```diff
@@ -0,0 +1,18 @@
+forklift_name: pipeline-katello-devel
+forklift_boxes:
+  pipeline-katello-devel-centos7:
+    box: centos7
+    memory: 8192
+katello_repositories_version: nightly
+# katello_repositories_environment: staging
```
ehelms: This can be dropped

pipelines/vars/katello_devel.yml
```diff
@@ -0,0 +1,18 @@
...
+katello_repositories_pulp_version: nightly
+foreman_repositories_version: nightly
+foreman_client_repositories_version: "{{ foreman_repositories_version }}"
+# foreman_repositories_environment: staging
```
ehelms: This can be dropped

pipelines/vars/katello_devel.yml
```diff
@@ -0,0 +1,18 @@
...
+foreman_installer_scenario: katello-devel
+foreman_installer_additional_packages:
+  - foreman-installer-katello
+foreman_installer_options_internal_use_only:
+  - "--disable-system-checks"
+  - "--katello-devel-enable-ostree=true"
+pipeline_run_tests: False
```
ehelms: Not explicitly needed

roles/katello_devel/meta/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+dependencies:
+  - role: etc_hosts
+  - role: foreman_server_repositories
+    foreman_server_repositories_ostree: true
+    foreman_server_repositories_katello: true
+    foreman_server_repositories_foreman_client: true
+    # foreman_server_repositories_katello_client: true
```
ehelms: This can be dropped

playbooks/devel.yml
```diff
@@ -13,7 +13,4 @@
       - "--disable-system-checks"
       - "--katello-devel-enable-ostree=true"
   roles:
-    - etc_hosts
-    - foreman_server_repositories
-    - foreman_installer_devel_scenario
-    - foreman_installer
+    - katello_devel
```
ehelms: I'd vote for changing this playbooks name to be more explicit to `katello_devel.yml` since it uses the katello_devel role and is focused on Katello and not general development environments.

ehelms:
Last comment then good to go. I just want us to be clear that this is Katello specific and not general purpose. I think there are changes we could make to enable a more general purpose starting point but they don't need to be included here given status quo.

akofink:
Updated devel.yml to katello_devel.yml in an additional commit. This is going to break people's local boxes config.

ehelms:
@akofink I think as long as we send out a note to discourse, that breakage is acceptable.

playbooks/katello_devel.yml
```diff
@@ -13,7 +13,4 @@
       - "--disable-system-checks"
       - "--katello-devel-enable-ostree=true"
   roles:
-    - etc_hosts
-    - foreman_server_repositories
-    - foreman_installer_devel_scenario
-    - foreman_installer
+    - katello_devel
```
ehelms: Sorry just occurred to me, the `vars` section in this playbook. Is that needed anymore with the meta role?

PR #897, opened 2019-01-07
make example bullet points in devel.md render properly

evgeni:


ehelms:
Visually, I feel like the bullets should be indented when rendered. Maybe we need to bump them to the right a space or two?

evgeni:
@ehelms like this?

PR #898, opened 2019-01-08
Add Dominic Cleal's advanced dns configuration

akofink:
This is by far the best option for hands off name resolution IMO, though it requires modifying your host's network configuration to use dnsmasq (natively supported in NetworkManager). I've had some requests to include it in the forklift docs, so here it is! I think a link to the instructions is sufficient, as it's an advanced libvirt configuration not specific to forklift.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,9 @@ vagrant plugin install vagrant-hostmanager
 
 By default, the boxes are set with `example.com` domain.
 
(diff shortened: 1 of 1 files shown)


akofink:
Nice! I'll have to try that out. :)

lzap:
Yeah this works for me: https://lukas.zapletalovi.com/2017/10/definitive-solution-to-libvirt-guest-naming.html

PR #903, opened 2019-01-15
Skip intermediate step for Katello 3.8 upgrade pipeline

jturel:
Installing Katello 3.6 on CentOS 7.6 is not successful due to an AVC denial when candlepin tries to connect to postgres. I think we can get away with testing only  3.7 -> 3.8 upgrade as part of that pipeline rather than 3.6 -> 3.7 -> 3.8

This change adds the 'intermediate' tag to the intermediate upgrade play so that external sources (jenkins) can skip this step.

There may be a better way to go about this but this seemed a concise enough way. I also did some cleanup of unneeded vars.

Foreman-infra counterpart: https://github.com/theforeman/foreman-infra/pull/910


jturel:
@evgeni I think you did this upgrade pipeline refactor so I'm curious about your take on my approach!

jturel:
Works to the extent that we can reach the point of running bats.

jturel:
@evgeni how do you like this one? :)

evgeni:
@jturel so you're basically installing 3.7, upgrading to 3.7 (noop) and upgrading to 3.8? that's a nice hack! much nicer than before.

evgeni:
Yeah, I think I like this more than my idea with `meta: end_play`

:+1:

ekohl:
Sort of related: does this mean we can drop the `pipelines/pipeline_katello_*` files as well?

evgeni:
@ekohl you can't do `when` on a whole play, just separate tasks/roles :(

evgeni:
And yes, I think the other pipeline files can be dropped now.

jturel:
I'll put up another PR to remove the old pipelines

ekohl:
I think you could set `foreman_installer_skip_installer: katello_version_start == katello_version_intermediate` to optimize a bit in the intermediate step (not sure about the exact syntax) but I guess this is already an improvement.

jturel:
@ekohl I like that idea. I added it and removed the old upgrade pipelines too. Should work, and I'm running the pipeline now to be sure.

jturel:
Works!

jturel:
Feel free to hit this with a merge so I can kick the 3.8 pipeline :)

ekohl:
Thanks!

PR #911, opened 2019-01-30
[WIP] obalize the pipelines

ekohl:
Requires https://github.com/theforeman/obal/pull/169

Looks like we do need to override the prog (which we now hardcode), but the idea is:

```
$ ./bin/forklift --help
usage: obal [-h] action ...

positional arguments:
  action                which action to execute
    katello_pipeline    Perform a Katello installation with a separate proxy
                        and run tests on it
    katello_upgrade_pipeline
                        Perform a Katello upgrade pipeline

optional arguments:
  -h, --help            show this help message and exit
```

```
$ ./bin/forklift katello_pipeline --help
usage: obal katello_pipeline [-h] [-v] [-e EXTRA_VARS]
                             [--forklift-state FORKLIFT_STATE]
                             [--katello-version KATELLO_VERSION]

Perform a Katello installation with a separate proxy and run tests on it

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         verbose output
  --forklift-state FORKLIFT_STATE
                        The state of the machines to ensure
  --katello-version KATELLO_VERSION
                        The Katello version

advanced arguments:
  -e EXTRA_VARS, --extra-vars EXTRA_VARS
                        set additional variables as key=value or YAML/JSON, if
                        filename prepend with @
```

### DIFF
--- a/bin/forklift
+++ b/bin/forklift
@@ -1,0 +1,5 @@
+#!/bin/sh
+export OBAL_DATA="$(dirname $(dirname $(realpath $0)))"
+export OBAL_PLAYBOOKS="${OBAL_DATA}/pipelines"
+export OBAL_INVENTORY="${OBAL_DATA}/inventories"
+exec obal "$@"
--- a/pipelines/katello_pipeline/katello_pipeline.yaml
+++ b/pipelines/katello_pipeline/katello_pipeline.yaml
@@ -1,7 +1,7 @@
 - hosts: localhost
   vars_files:
-    - vars/katello_base.yml
-    - vars/katello_{{ katello_version }}.yml
+    - ../vars/katello_base.yml
+    - ../vars/katello_{{ katello_version }}.yml
   roles:
     - forklift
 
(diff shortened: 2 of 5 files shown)


ekohl:
Contains https://github.com/theforeman/forklift/pull/912 because that's needed, but can be merged even without this. I ran this locally and confirmed this works. It is missing tab completion but that'd probably be very slow because our ansible inventory is very slow.

ehelms:
I like the general idea of what you are going for and the re-use of Obal. However, I think between you, @evgeni and myself we need to chat about whether we want to do this in it's current form. Obal has a specific mission. Builtin to Obal is a method to turn a series of playbooks, roles, and modules into a CLI. I think we need to extract that core logic rather than Obalize. Obal is specifically dedicated to packaging workflows. And while this will be more work, I think it is cleaner to extract the core CLI logic and re-use that part. This will allow it to be smaller, more focused and not over-confuse the intent of Obal itself. Obal then becomes just the packaging dedicated roles/playbooks/modules using some core.

Let the bikeshedding on a name for the core part begin!

ehelms:
@ekohl based on other discussions we are having, can this be closed?

ekohl:
I still like the concept because running pipelines is now very cumbersome. I have to lookup documentation every single time. Going to close it though until we have figured out the obalization step.

PR #912, opened 2019-01-30
Use a forklift_directory variable

ekohl:
This also changes the value to be based on the role path (which is constant) while the CWD might change.

### DIFF
--- a/roles/forklift/defaults/main.yml
+++ b/roles/forklift/defaults/main.yml
@@ -1,2 +1,3 @@
 ---
+forklift_directory: "{{ role_path | dirname | dirname }}"
 forklift_state: ''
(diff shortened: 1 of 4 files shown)


roles/forklift/defaults/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
+forklift_directory: "{{ role_path | dirname | dirname }}"
```
ekohl: I now realized this might break use cases where forklift is used as a library (sat-deploy)

ehelms: You can test via https://github.com/pulp/pulplift

ehelms: Why the double dirname?

ehelms:
Can you expand on what problem this is solving?

roles/forklift/defaults/main.yml
ekohl: Because it's in `roles/forklift`

ekohl:
> Can you expand on what problem this is solving?

In https://github.com/theforeman/forklift/pull/911 I'm moving the pipelines around which changes the working directory. I thought this was more stable.

PR #914, opened 2019-02-06
Fix vagrant-hostmanager

jturel:
/etc/hosts wasn't getting updated on my local machine ... but now it is!

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -30,13 +30,13 @@ module Forklift
         'cachier' => {
           'mount_options' => ['rw', 'vers=3', 'tcp', 'nolock']
         },
-        'cachier_enabled': true,
-        'hostmanager_enabled': true
+        'cachier_enabled' => true,
+        'hostmanager_enabled' => true
       }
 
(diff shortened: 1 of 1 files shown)


vagrant/lib/forklift/box_distributor.rb
```diff
@@ -30,13 +30,13 @@ def settings
...
-        'cachier_enabled': true,
-        'hostmanager_enabled': true
+        'cachier_enabled' => true,
+        'hostmanager_enabled' => true
       }
 
       overrides = YAML.load_file(settings_file) if File.exist?(settings_file)
 
-      @settings ||= default_settings.merge(overrides)
+      default_settings.merge(overrides)
```
jturel: this assignment seems unnecessary

vagrant/lib/forklift/box_distributor.rb
```diff
@@ -30,13 +30,13 @@ def settings
         'cachier' => {
           'mount_options' => ['rw', 'vers=3', 'tcp', 'nolock']
         },
-        'cachier_enabled': true,
-        'hostmanager_enabled': true
+        'cachier_enabled' => true,
+        'hostmanager_enabled' => true
```
jturel: these were symbols in the hash, but being referenced as strings

PR #915, opened 2019-02-06
Make devel box explicitly katello focused

ehelms:


jturel:
Is this in preparation for some other change or just a general clarification?

ehelms:
For now a general clarification given the tight coupling that exists. I think it would be good to generalize back to a devel (for all environments and plugins) environment and setup at some point.

jturel:
There are two centos7-devels you problem want to touch up:

```
README.md:vagrant up centos7-devel
README.md:Sometimes you want to spin up the same box type (e.g. centos7-devel) from within the forklift directory. While this can be added to the Vagrantfile directly, updates to the forklift repository could wipe out your local changes. To help with this, you can define a custom box re-using the configuration within the Vagrantfile. To do so, create a `99-local.yaml` file. For example, to create a custom box on CentOS 7 with nightly and run the installers reset command:
```

ehelms:
Good find @jturel  thanks!

PR #917, opened 2019-02-11
Don't run bats in devel pipeline

akofink:


johnpmitsch:
Looks good, any thoughts about removing the proxy testing for devel here too? We can still add the proxy test later, but I would rather keep the jenkins job simple to start, just testing the katello_devel role, since this matches the centos7-devel VM.

I think you could do something with Ansible's [add_hosts](https://docs.ansible.com/ansible/latest/modules/add_host_module.html) to dynamically add the proxy host (but I haven't tried)

akofink:
This is a tiny change, but turning off the proxy will require exposing variables to control which boxes are brought up. Because of this, I think we should keep the changes separate and turn off the proxy in a future PR.

ehelms:
Since you have to set this explicitly, it makes me think this line `https://github.com/theforeman/forklift/blob/master/pipelines/katello_pipeline.yml#L70` is incorrectly specified. Does this need an update to `pipeline_run_tests == True` ?

akofink:
No, I'm quite sure you can use boolean variables like this with ansible, but I think we might need to specify a sane default if it's not defined (there is no defaults for this variable).

ehelms:
Hrmm OK. As we discussed on IRC, we may be pushing the limits of shoe-horning in development to this playbook.

PR #920, opened 2019-02-15
Separate devel pipeline from release pipeline

johnpmitsch:


johnpmitsch:
foreman-infra PR: [theforeman/foreman-infra#950](https://github.com/theforeman/foreman-infra/pull/950)

PR #921, opened 2019-02-19
Always use taxonomies in forman_provisioning playbook

xprazak2:
Foreman now installs with taxonomies enabled, so we need to make sure resources are assigned to org/loc.

### DIFF
--- a/roles/foreman_provisioning/tasks/main.yml
+++ b/roles/foreman_provisioning/tasks/main.yml
@@ -1,21 +1,4 @@
 ---
-# make installation media available
-# until it's done by seed - http://projects.theforeman.org/issues/22661
-- name: 'assign organization and location to all seeded media'
-  with_items:
-    - CentOS mirror
-    - CoreOS mirror
-    - Debian mirror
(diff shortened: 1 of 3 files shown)


roles/foreman_provisioning/tasks/main.yml
```diff
@@ -13,7 +13,7 @@
     - Ubuntu mirror
   shell: >
     {{ foreman_provisioning_hammer }} medium update --name "{{ item }}" --organization-titles "{{ foreman_provisioning_organization }}" --location-titles "{{ foreman_provisioning_location }}"
-  when: foreman_provisioning_use_taxonomies and foreman_provisioning_foreman_version == 'nightly'
+  when: foreman_provisioning_foreman_version == 'nightly'
```
ekohl: Why do you only do this on nightly? Is this even needed now that we have seeding scripts that should always assign it to a taxonomy?

xprazak2: Not needed now, seeds take care of it. Good catch.

PR #925, opened 2019-03-01
Add Foreman pipeline

ehelms:
Adds Foreman pipeline's similar to the Katello ones with ability to specify OS:

```
 ansible-playbook pipelines/foreman_pipeline.yml -e forklift_state=up -e foreman_version=nightly -e foreman_os=centos7

 ansible-playbook pipelines/foreman_pipeline.yml -e forklift_state=up -e foreman_version=nightly -e foreman_os=debian9
```


pipelines/pipeline_foreman_nightly.yml
```diff
@@ -1,13 +1,39 @@
 - hosts: localhost
```
evgeni: I guess this file was used as a playground and should not be here in the PR?

pipelines/vars/foreman_1.20.yml
```diff
@@ -0,0 +1,4 @@
+forklift_name: pipeline-foreman-1.20
```
evgeni: missing the OS?

pipelines/vars/foreman_1.21.yml
```diff
@@ -0,0 +1,4 @@
+forklift_name: pipeline-foreman-1.21
```
evgeni: missing the OS?

evgeni:
the `pipelines/vars/foreman_centos7.yml` is empty?

ehelms:
Updated

pipelines/vars/foreman_nightly.yml
```diff
@@ -0,0 +1,3 @@
+forklift_name: "pipeline-foreman-nightly-{{ foreman_os }}"
+forklift_boxes: "{{ {'pipeline-foreman-nightly-' + foreman_os: server_box} }}"
```
ekohl: Should this be `{forklift_name: server_box}`

pipelines/vars/foreman_nightly.yml
```diff
@@ -0,0 +1,3 @@
+forklift_name: "pipeline-foreman-nightly-{{ foreman_os }}"
```
ekohl: Should this be `pipeline-foreman-{{ foreman_repositories_version }}-{{ foreman_os }}`

pipelines/vars/foreman_1.20.yml
```diff
@@ -0,0 +1,4 @@
+forklift_name: "pipeline-foreman-1.20-{{ foreman_os }}"
+forklift_boxes:
+  pipeline-foreman-1.20-centos7: "{{ server_box }}"
```
ekohl: This probably needs to copy the pattern from nightly to include the `foreman_os`. Same in the 1.21 file.

ehelms:
Good thoughts... I think I can simplify this even further and apply a similar approach to Katello.

pipelines/vars/foreman_base.yml
```diff
@@ -0,0 +1,12 @@
+foreman_installer_scenario: foreman
+foreman_installer_options_internal_use_only:
+  - "--foreman-initial-admin-password {{ foreman_installer_admin_password }}"
+  - "--puppet-server-max-active-instances 1"
+  - "--puppet-server-jvm-min-heap-size 1G"
+  - "--puppet-server-jvm-max-heap-size 1G"
+server_box:
+  box: "{{ foreman_os }}"
+  memory: 4192
```
ekohl: Is the switch from 4680 intentional? Either probably works.

pipelines/vars/foreman_base.yml
```diff
@@ -0,0 +1,12 @@
+foreman_installer_scenario: foreman
+foreman_installer_options_internal_use_only:
+  - "--foreman-initial-admin-password {{ foreman_installer_admin_password }}"
+  - "--puppet-server-max-active-instances 1"
+  - "--puppet-server-jvm-min-heap-size 1G"
+  - "--puppet-server-jvm-max-heap-size 1G"
+server_box:
+  box: "{{ foreman_os }}"
```
ekohl: Another thought: this could also use `{{ foreman_os }}-foreman-{{ foreman_version }}`. Then it would inherit all the defaults from [versions.yaml](https://github.com/theforeman/forklift/blob/master/vagrant/config/versions.yaml). This is important when we're going to switch the Puppet version (which I'm about to do). It also means you don't have to set the `foreman_repositories_version` and related katello/pulp versions.

pipelines/foreman_pipeline.yml
```diff
@@ -0,0 +1,29 @@
...
+  - pipeline-foreman-{{ foreman_version }}-{{ foreman_os }}
+  become: yes
+  vars_files:
+    - vars/foreman_base.yml
+    - vars/foreman_base_staging.yml
+  roles:
+    - disable_firewall
+    - foreman
+
+- hosts: pipeline-foreman-{{ foreman_version }}-{{ foreman_os }}
```
ekohl: Do you want to merge this into the previous play since it's on the same host or do you intend to add a proxy at some point?

ehelms:
I am going to leave this open to help inform, but I believe that https://github.com/theforeman/forklift/pull/928 should include the learnings from this and addition of Foreman pipelines.

pipelines/vars/foreman_base.yml
evgeni: I like that!

evgeni: So this does not work as intended right now, as we only set the right `*_version` vars inside the ansible provisioner, which we don't use during pipelines. I'll see what we can do here.

evgeni:
this has been now absorbed into #928 :)

ekohl:
We'll close it in favor of the borg then.

PR #926, opened 2019-03-01
don't check if katello_version is a number, it isn't

evgeni:
when you pass variables to ansible on the shell, they are always
interpreted as strings:

    % ansible -m debug -a 'msg={{ katello_version is number }}' -e katello_version=3.10 localhost
    localhost | SUCCESS => {
        "msg": false
    }

therefor always include the katello role when katello_version is not
"devel"

### DIFF
--- a/pipelines/katello_pipeline.yml
+++ b/pipelines/katello_pipeline.yml
@@ -22,7 +22,7 @@
       foreman_server_repositories_katello_client: katello_version == 'nightly' or katello_version is number and katello_version is version('3.9', '>=') | bool
     - role: update_os_packages
     - role: katello
-      when: katello_version is number or katello_version == "nightly"
+      when: katello_version != "devel"
     - role: katello_devel
       when: katello_version == "devel"
 


pipelines/katello_pipeline.yml
```diff
@@ -22,7 +22,7 @@
       foreman_server_repositories_katello_client: katello_version == 'nightly' or katello_version is number and katello_version is version('3.9', '>=') | bool
     - role: update_os_packages
     - role: katello
-      when: katello_version is number or katello_version == "nightly"
+      when: katello_version != "devel"
     - role: katello_devel
```
evgeni: @akofink @johnpmitsch do we even still need this separation, given that you've split off the devel pipe anyways?

evgeni: if we don't, please just ack https://github.com/theforeman/forklift/pull/927

evgeni:
closing this as #927 won the race

PR #927, opened 2019-03-01
don't try to load katello_devel in the katello pipeline

evgeni:


evgeni:
this is an alternative to #926, depending on the discussion there.

evgeni:
great, thanks!

PR #928, opened 2019-03-01
pipelines to test more plugins together

evgeni:
This implements the first level of my [Testing more plugins together regularly RFC](https://community.theforeman.org/t/testing-more-plugins-together-regularly/12897)


evgeni:
this will require changes to infra, so don't merge just yet please :)

pipelines/install_pipeline.yml
```diff
@@ -0,0 +1,68 @@
+- hosts: localhost
+  vars_files:
+    - vars/{{ pipeline_type | default('katello') }}_base.yml
```
evgeni: I'm still debating with myself if I want to keep this `default` here or not…

pipelines/install_pipeline.yml
```diff
@@ -0,0 +1,68 @@
...
+    - vars/{{ pipeline_type | default('katello') }}_base_staging.yml
+    - vars/{{ pipeline_type | default('katello') }}_{{ pipeline_version }}.yml
+  roles:
+    - role: disable_firewall
+    - role: disable_ipv6
+    - role: foreman_server_repositories
+      foreman_server_repositories_katello_client: pipeline_version == 'nightly' or pipeline_version is number and pipeline_version is version('3.9', '>=') | bool
+    - role: update_os_packages
+    - role: katello
+      when: pipeline_version is number or pipeline_version == "nightly"
```
evgeni: this will be updated based on #926 and #927

pipelines/install_pipeline.yml
ehelms: Drop it.

ehelms:
Let's go ahead and combine this with my PR to add Foreman pipelines. The key items to pull form that are:

 * OS specification for boxes
   - https://github.com/theforeman/forklift/pull/925/files#diff-af87b9abe11a42c483a7ea54dd6030e0R8
 * Move forklift name to base using available variables
   - https://github.com/theforeman/forklift/pull/925/files#diff-5b0d434fc320cbba41ed9f10c4ccaee0R10
 * Combine OS and use forklift name to create the boxes
   - https://github.com/theforeman/forklift/pull/925/files#diff-5b0d434fc320cbba41ed9f10c4ccaee0R11

Note for the Foreman pipeline, I did not need any version specific vars files. The supplied variables do all the work. I feel like we could do the same through a version mapping for Katello and Luna.

pipelines/upgrade_pipeline.yml
```diff
@@ -0,0 +1,65 @@
...
+    - vars/{{ pipeline_type }}_base_release.yml
+    - vars/{{ pipeline_type }}_upgrade_{{ pipeline_version }}.yml
+    - vars/{{ pipeline_type }}_{{ pipeline_version_start }}.yml
+  roles:
+    - umask
+    - selinux
+    - etc_hosts
+    - foreman_server_repositories
+    - role: foreman_client_repositories
+      when: pipeline_version_start == "nightly" or pipeline_version_start is version('3.9', '>=')
```
ekohl: 3.9 sounds like a Katello version. I think we started with 1.20. We should always have the foreman version at this point so let's normalize on that. This is repeated a few times so I won't duplicate the comment :)

pipelines/upgrade_pipeline.yml
```diff
@@ -0,0 +1,65 @@
...
+- hosts: pipeline-upgrade-{{ pipeline_type }}-{{ pipeline_version}}-{{ pipeline_os }}
+  become: true
+  vars_files:
+    - vars/{{ pipeline_type }}_base.yml
+    - vars/{{ pipeline_type }}_base_release.yml
+    - vars/{{ pipeline_type }}_upgrade_{{ pipeline_version }}.yml
+    - vars/{{ pipeline_type }}_{{ pipeline_version_intermediate }}.yml
+  vars:
+    foreman_installer_upgrade: True
+    foreman_installer_skip_installer: "{{ true if pipeline_version_start == pipeline_version_intermediate else false }}"
```
ekohl: Not sure if Ansible can do this, but can it be `{{ pipeline_version_start == pipeline_version_intermediate }}`

evgeni: no idea, that's not my code here :eyes:

pipelines/upgrade_pipeline.yml
evgeni: that's fair, yeah.

ehelms:
Can we drop the version specific variable files? All of that information should be storable in some dictionary to do look ups in or somehow in the `vagrant/config/versions.yaml` file.

evgeni:
> Can we drop the version specific variable files? All of that information should be storable in some dictionary to do look ups in or somehow in the `vagrant/config/versions.yaml` file.

I'd like to yeah. Currently, `vagrant/config/versions.yaml` is only passed to Vagrant's internal Ansible provisioner, which we use for `centos7-foreman-1.21` but not for the `pipeline-*` boxes that we run directly via `ansible-playbook`. We'd need to enhance the vagrant inventory to read from that file.

pipelines/upgrade_pipeline.yml
evgeni: I'll fix that as soon as https://github.com/theforeman/forklift/pull/929 is in.

evgeni: done via #929

evgeni:
> > Can we drop the version specific variable files? All of that information should be storable in some dictionary to do look ups in or somehow in the `vagrant/config/versions.yaml` file.
> 
> I'd like to yeah. Currently, `vagrant/config/versions.yaml` is only passed to Vagrant's internal Ansible provisioner, which we use for `centos7-foreman-1.21` but not for the `pipeline-*` boxes that we run directly via `ansible-playbook`. We'd need to enhance the vagrant inventory to read from that file.

Hah, so I've implemented that in the inventory just to realize that does not help us, as during an upgrade pipeline we need to actually switch versions.

Back to the drawing board.

ehelms:
Two questions (well, one is a request):

 1) Can you go ahead and open foreman-infra PRs that take in this change?
 2) Could the upgrade variables just be merged into the version yaml files for each type and just have one?

evgeni:
> Two questions (well, one is a request):
> 
> 1. Can you go ahead and open foreman-infra PRs that take in this change?

I will, yes.

> 2. Could the upgrade variables just be merged into the version yaml files for each type and just have one?

To have the `_start`, `_intermediate` and `_final` vars? *probably* yes. Need to test that as such a merge would also mean that when we we load vars file for each step to know the current (to be installed/upgraded-to) version, we also will override the step (start, intermediate, final) vars.

evgeni:
so I got rid of the version vars files and left the upgrade version vars there as we need to know the upgrade path somewhere.

might want to move that to `versions.yaml` at some point, but it felt too complicated now.

ekohl:
If I can nit a bit about the filenames: `pipelines/install_pipeline.yml` feels a bit redundant to me.

roles/forklift_versions/lookup_plugins/forklift_versions.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class LookupModule(LookupBase):
+
+    def run(self, terms, variables=None, **kwargs):
+        ret = []
+
+        for term in terms:
+            display.debug("Forklift lookup term: %s" % term)
+            lookup_params = {key: value for (key, value) in [x.split('=') for x in term.split()]}
```
ekohl:
```suggestion
            lookup_params = dict(x.split('=', 1) for x in term.split())
```

roles/forklift_versions/lookup_plugins/forklift_versions.py
```diff
@@ -0,0 +1,69 @@
...
+            scenario = lookup_params['scenario']
```
ekohl: Do we want to raise a nice error on KeyError?

evgeni:
@ehelms WIP jenkins change at https://github.com/theforeman/foreman-infra/pull/956

roles/forklift_versions/lookup_plugins/forklift_versions.py
evgeni: yes, I need to adapt ansible style error handling to this, the above is more of a draft :)

evgeni:
> If I can nit a bit about the filenames: `pipelines/install_pipeline.yml` feels a bit redundant to me.

hah, so it's not only me. good. `install.yml` and `upgrade.yml`?

ehelms:
I like the filenames as is right now. They are explicit. Which I think is good to prevent users from thinking they can use install to give them an installation. Or upgrade to perform an upgrade.

evgeni:
Doesn't the `pipelines` folder give that hint already?

ehelms:
I just tend to think that pipeline is not a common idiom for any ecosystem we are using here and that it doesn't hurt to be more explicit here compared to the playbooks directory. Aint gonna block it, just weighin' in.

pipelines/vars/luna_upgrade_3.8.yml
```diff
@@ -0,0 +1 @@
+katello_upgrade_3.8.yml
```
ehelms: Puttin gthe name of a file in the file causes it to include the contents of the file?

pipelines/install_pipeline.yml
```diff
@@ -0,0 +1,71 @@
+- hosts: localhost
+  vars_files:
+    - vars/{{ pipeline_type }}_base.yml
+    - vars/{{ pipeline_type }}_{{ pipeline_version }}.yml
```
ehelms: Should this line be gone now?

pipelines/vars/luna_upgrade_3.8.yml
evgeni: Nah, this is the GH ui being silly, it's a symlink as luna and katello upgrade paths are identical.

evgeni:
Oups. Yes

evgeni:
Given it's "just" 2:1 for the rename without `pipeline`, I'll leave it for now, let's see if anyone else is in favor of renaming.

Other than that, I'd say this is in good shape for a fresh review round.

Questions/thoughts that I have personally:
* are the variables I introduced/renamed good? `pipeline_os`, `pipeline_type`, `pipeline_version`. I'm thinking of using `pipeline_scenario` instead of `_type`?
* can we somehow nicely add the upgrade vars to `versions.yaml`? I did not so far as the path is different between Katello and Foreman (not that I've added a Foreman upgrade pipe yet). And we can also not "compute" it as some versions have only one step in their upgrade pipe (3.7 → 3.8, as there is no 3.6 support anymore).

roles/forklift_versions/lookup_plugins/forklift_versions.py
```diff
@@ -0,0 +1,94 @@
...
+            else:
+                upgrade_versions = set()
+                for version in reversed(versions['installers']):
+                    if version[scenario] == scenario_version:
+                        upgrade_versions.add(scenario_version)
+                    elif 1 <= len(upgrade_versions) < 3:
+                        upgrade_versions.add(version[scenario])
+                upgrade_versions = sorted(upgrade_versions, key=version_sort_key)
+                while len(upgrade_versions) < 3:
+                    upgrade_versions.insert(0, upgrade_versions[0])
```
ekohl: Catch an `IndexError` in case the specified version wasn't in the versions file at all?

I was also thinking about how clear this is, but it's probably less readable:

```python
upgrade_versions = [upgrade_versions[0]] * (3 - len(upgrade_versions)) + upgrade_versions
```

Should 3 be a constant somewhere?

roles/forklift_versions/lookup_plugins/forklift_versions.py
```diff
@@ -0,0 +1,94 @@
...
+                        ret.append(forklift_vars)
+                        break
+            else:
+                upgrade_versions = set()
+                for version in reversed(versions['installers']):
+                    if version[scenario] == scenario_version:
+                        upgrade_versions.add(scenario_version)
+                    elif 1 <= len(upgrade_versions) < 3:
+                        upgrade_versions.add(version[scenario])
+                upgrade_versions = sorted(upgrade_versions, key=version_sort_key)
```
ekohl: Why do we sort the versions in the end? Shouldn't we sort installers since the loop already relies on order.

evgeni: This is Evgeni being SMRT ;-)

`upgrade_versions` starts its life as a `set`, so that when I loop over the installers, I can add them as they come and ignore the fact that we have Foreman 1.20 twice. However, when you convert a `set` to a `list`, it seems to lose insertion order (or rather, a `set` is alphabetically ordered):

```
>>> list(set(['1.20', '1.20', '1.21', 'nightly']))
['nightly', '1.21', '1.20']
>>> list(set(['3.9', '3.10', '3.11']))
['3.10', '3.11', '3.9']
```

ekohl: Oh yes, there is no SortedSet. Carry on.

roles/forklift_versions/lookup_plugins/forklift_versions.py
evgeni: I don't think that's readable ;-)

Implemented the two other suggestions

pipelines/install_pipeline.yml
```diff
@@ -1,41 +1,47 @@
...
-    - vars/katello_base.yml
-    - vars/katello_{{ katello_version }}.yml
+    - vars/{{ pipeline_type }}_base.yml
   roles:
     - forklift
 
 - hosts:
-  - pipeline-katello-{{ katello_version }}-centos7
-  - pipeline-proxy-{{ katello_version }}-centos7
+  - pipeline-{{ pipeline_type }}-{{ pipeline_version }}-{{ pipeline_os }}
```
ekohl: Is this now:
```suggestion
  - {{ forklift_name }}
```

pipelines/install_pipeline.yml
```diff
@@ -1,41 +1,47 @@
...
+  - pipeline-{{ pipeline_type }}proxy-{{ pipeline_version }}-{{ pipeline_os }}
```
ekohl: And this:
```suggestion
  - {{ forklift_proxyname }}
```

pipelines/install_pipeline.yml
evgeni: potentially, it is, but then, who was there first, the chicken or the egg?
(`vars_files` is not loaded at the time ansible evaluates `hosts`)

evgeni:
:crossed_fingers:

PR #929, opened 2019-03-05
unify the way we load the foreman client repo in the pipelines

evgeni:


pipelines/katello_pipeline.yml
```diff
@@ -19,7 +19,7 @@
     - role: disable_firewall
     - role: disable_ipv6
     - role: foreman_server_repositories
-      foreman_server_repositories_katello_client: katello_version == 'nightly' or katello_version is number and katello_version is version('3.9', '>=') | bool
+      foreman_server_repositories_foreman_client: katello_version == "nightly" or katello_version is version('3.9', '>=') | bool
```
ekohl: Should we have a tri-state value for this parameter? true (always on), auto (on if the version is new enough) or false (always off). Then we don't need to duplicate all when code.

evgeni: I'll think about this

evgeni: @ekohl how about this? instead of making weird states, I just made `foreman_client_repositories` (which is what gets loaded when you pass `foreman_server_repositories_foreman_client: true`) to be a NOOP on pre-1.20.

roles/foreman_client_repositories/tasks/repo.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: 'Foreman Client Staging Repository'
+  yum_repository:
+    name: foreman-client-koji
+    description: "Foreman {{ foreman_client_repositories_version }} Client Koji Repository"
+    baseurl: "http://koji.katello.org/releases/yum/foreman-client-{{ foreman_client_repositories_version }}/el{{ ansible_distribution_major_version }}/x86_64/"
```
ekohl: `foreman_client_repositories_dists[ansible_os_family]` should also work here, right?

evgeni: probably, didn't look further into the actual repo code

PR #933, opened 2019-03-09
Remove containers folder that was moved repositories

ehelms:
Moved to https://github.com/ehelms/grapple


ehelms:
Here https://github.com/theforeman/foreman-infra/pull/957

PR #934, opened 2019-03-12
Add nodejs-scl defaulting to version 8

ehelms:


ehelms:
Requires -- https://github.com/theforeman/puppet-katello_devel/pull/195

roles/nodejs_scl/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: 'Install Nodejs SCL'
+  yum:
+    name:
+      - "{{ nodejs_scl_version }}"
+      - "{{ nodejs_scl_version }}"
+    state: present
+
+- name: 'Enable SCL at login'
+  blockinfile:
+    dest: /etc/profile.d/enable-{{ nodejs_scl_version }}.sh
```
ekohl: This makes me wonder about the Ruby SCL. If we change the value, it continues to load the old SCL. Perhaps this needs to be static and just `enable-nodejs-scl.sh`. Same with Ruby.

PR #935, opened 2019-03-13
Set forklift state to up by default

ehelms:


ehelms:
Simpifying the command structure since if you are running a pipeline using forklift role you want it to actually run by default

zjhuntin:
:+1:

PR #938, opened 2019-03-14
Set GEM_HOME and PATH to use ~/.gem explicitly

akofink:
This used to work fine, but something changed for some reason, and we
need to explicitly tell bundler to install locally

### DIFF
--- a/roles/ruby_scl/tasks/main.yml
+++ b/roles/ruby_scl/tasks/main.yml
@@ -10,10 +10,12 @@
       - "{{ ruby_scl_version }}-ruby-devel"
     state: present
 
-- name: 'Enable SCL at login'
+- name: Enable and configure SCL at login
   blockinfile:
     dest: /etc/profile.d/enable-{{ ruby_scl_version }}.sh
     create: yes
     block: |
       #!/bin/bash
       source scl_source enable {{ ruby_scl_version }}
+      export GEM_HOME=$(ruby -e 'print Gem.user_dir')
+      export PATH=$PATH:$(ruby -e 'print Gem.user_dir')/bin


roles/hammer_devel/tasks/hammer_install.yml
```diff
@@ -53,5 +53,11 @@
...
 
+- name: Set GEM_HOME
+  lineinfile: dest=~/.bash_profile line="export GEM_HOME=$HOME/.gem"
+
+- name: Set PATH
+  lineinfile: dest=~/.bash_profile line="export PATH=$PATH:$HOME/.gem/bin"
+
 - name: 'Install gems'
-  bundler: chdir=~/hammer-cli-foreman state=present executable=~/bin/bundle
+  bundler: chdir=~/hammer-cli-foreman state=present executable=~/bin/bundle gem_path=~/.gem
```
ekohl: Perhaps this should be split to multiple lines for readability.

roles/hammer_devel/tasks/hammer_install.yml
```diff
@@ -53,5 +53,11 @@
 - name: 'Install bundler'
   gem: name=bundler state=present
 
+- name: Set GEM_HOME
```
ekohl: This is a side effect of https://github.com/theforeman/forklift/pull/841 which switched to `rh-ruby25`. That's affected by https://bugzilla.redhat.com/show_bug.cgi?id=1613864#c5. See https://github.com/theforeman/puppet-katello_devel/pull/178#issuecomment-455719359 as well.

Perhaps we should do this as part of the SCL role to work around it. There we already deploy it in profile.d, maybe a better place? Maybe with that katello devel could even avoid using `.vendor` which in my experience led to a faster install.

akofink:
@ekohl take a look! This should solve all our problems (hopefully). :) I'm going to submit a puppet PR to test with this change and centos7-devel.

roles/ruby_scl/tasks/main.yml
```diff
@@ -10,10 +10,13 @@
...
 
-- name: 'Enable SCL at login'
+- name: Enable and configure SCL at login
   blockinfile:
     dest: /etc/profile.d/enable-{{ ruby_scl_version }}.sh
     create: yes
     block: |
       #!/bin/bash
       source scl_source enable {{ ruby_scl_version }}
+      export GEM_HOME=$HOME/.gem
```
ekohl: It may be sufficient to use `export GEM_HOME="$(ruby -e 'print Gem.user_dir')"` and drop the config modification part.

roles/ruby_scl/tasks/main.yml
```diff
@@ -10,10 +10,13 @@
...
+      export PATH=$PATH:$HOME/.gem/bin
+      bundle config path $HOME/.gem
```
akofink: This might not be necessary... (let me test)

roles/ruby_scl/tasks/main.yml
```diff
@@ -10,10 +10,13 @@
...
-- name: 'Enable SCL at login'
+- name: Enable and configure SCL at login
   blockinfile:
     dest: /etc/profile.d/enable-{{ ruby_scl_version }}.sh
     create: yes
     block: |
       #!/bin/bash
       source scl_source enable {{ ruby_scl_version }}
+      export GEM_HOME=$HOME/.gem
+      export PATH=$PATH:$HOME/.gem/bin
```
ekohl: I'm not opposed to this, but I always use `bundle exec <gem>` and then it's not needed. That said, I have no objections to people who do like this to be there.

akofink:
Updated to use `Gem.user_dir` in the env variables. I think this is good to go now.

roles/ruby_scl/tasks/main.yml
akofink: People have been complaining recently about the need for `bundle exec` after the scl update, so I think we should keep it.

PR #942, opened 2019-03-18
Support using fqdn as inventory name in etc_hosts

iNecas:
I'm using fqdn as inventory name for my hosts. Before this patch, I was
ending with etc hosts like this:

   1.2.3.4 my-host-example-com.example.com my-host-example-com

This was because the `etc_hosts` role haven't expected fqdn in inventory
name. After this patch, both short and long names in inventory are
supported.

### DIFF
--- a/roles/etc_hosts/tasks/hosts_file_record.yml
+++ b/roles/etc_hosts/tasks/hosts_file_record.yml
@@ -1,0 +1,27 @@
+---
+# expected vars:
+#  * ip
+#  * hostname
+#  * domain
+
+# the domain is not included in the hostname
+- name: "Add {{ hostname }} to hosts file: short and full"
+  lineinfile:
+    dest: /etc/hosts
+    regexp: ".*{{ hostname.replace('.', '-') }}$"
+    line: "{{ ip }} {{ hostname.replace('.', '-') }}.{{ domain }} {{ hostname.replace('.', '-') }}"
+    state: present
+  when: not hostname | regex_search(domain | regex_escape())
+  become: yes
+  tags: env_setup
+
+# the domain is part of the hostname
+- name: "Add {{ hostname }} to hosts file: full only"
+  lineinfile:
+    dest: /etc/hosts
+    regexp: ".*{{ hostname }}$"
+    line: "{{ ip }} {{ hostname }}"
+    state: present
+  when: hostname | regex_search(domain | regex_escape())
+  become: yes
+  tags: env_setup
--- a/roles/etc_hosts/tasks/main.yml
+++ b/roles/etc_hosts/tasks/main.yml
@@ -1,10 +1,10 @@
 ---
 - name: "Build hosts file"
-  lineinfile:
-    dest: /etc/hosts
-    regexp: ".*{{ item.replace('.', '-') }}$"
-    line: "{{ hostvars[item].ansible_host }} {{ item.replace('.', '-') }}.{{ etc_hosts_domain }} {{ item.replace('.', '-') }}"
-    state: present
+  include_tasks: "./hosts_file_record.yml"
+  vars:
+    ip: "{{ hostvars[item].ansible_host }}"
+    domain: "{{ etc_hosts_domain }}"
+    hostname: "{{ item }}"
   when: hostvars[item].ansible_host is defined
   become: yes
   with_items: "{{ groups['all'] }}"


iNecas:
@evgeni mind reviewing?

roles/etc_hosts/tasks/main.yml
```diff
@@ -1,10 +1,10 @@
 ---
 - name: "Build hosts file"
-  lineinfile:
-    dest: /etc/hosts
-    regexp: ".*{{ item.replace('.', '-') }}$"
-    line: "{{ hostvars[item].ansible_host }} {{ item.replace('.', '-') }}.{{ etc_hosts_domain }} {{ item.replace('.', '-') }}"
-    state: present
+  include_tasks: "./hosts_file_record.yml"
```
evgeni: no need to `./` here

roles/etc_hosts/tasks/main.yml
```diff
@@ -1,10 +1,10 @@
...
+  vars:
+    ip: "{{ hostvars[item].ansible_host }}"
+    domain: "{{ etc_hosts_domain }}"
+    hostname: "{{ item }}"
```
evgeni: if we pass `{{ item | regex_replace('\..*') }}` here instead (thus stripping the domain from the name), can we then avoid the duplicated code and complex searches/when conditions?

evgeni: wait, no, we can't as `item` will be something like `foreman-1.20.inecas.example.com`, right?

then `{{ item | regex_replace(etc_hosts_domain | regex_escape()) }}`?

roles/etc_hosts/tasks/hosts_file_record.yml
```diff
@@ -0,0 +1,27 @@
...
+#  * domain
+
+# the domain is not included in the hostname
+- name: "Add {{ hostname }} to hosts file: short and full"
+  lineinfile:
+    dest: /etc/hosts
+    regexp: ".*{{ hostname.replace('.', '-') }}$"
+    line: "{{ ip }} {{ hostname.replace('.', '-') }}.{{ domain }} {{ hostname.replace('.', '-') }}"
+    state: present
+  when: not hostname | regex_search(domain | regex_escape())
```
evgeni: would `hostname.find(domain) == -1` be easier?

roles/etc_hosts/tasks/hosts_file_record.yml
```diff
@@ -0,0 +1,27 @@
...
+  tags: env_setup
+
+# the domain is part of the hostname
+- name: "Add {{ hostname }} to hosts file: full only"
+  lineinfile:
+    dest: /etc/hosts
+    regexp: ".*{{ hostname }}$"
+    line: "{{ ip }} {{ hostname }}"
+    state: present
+  when: hostname | regex_search(domain | regex_escape())
```
evgeni: would `hostname.find(domain) != -1` be easier?

ekohl:
I'd actually be happy to nuke `/etc/hosts` handling and require users to ensure their own environment works. I thought by default vagrant or libvirt handles this or you can configure it to work way better by providing actual DNS. I've had issues where I shut down a VM, booted it later and it got a different IP.

evgeni:
Vagrant by default does not, no. Vagrant-libvirt does, sorta, but not without [my commit](https://github.com/vagrant-libvirt/vagrant-libvirt/commit/53539f6ad4f434f55145cb83f68b5c553b08f165) for the management network (which we use).

There is a [hostmanager plugin](https://github.com/devopsgroup-io/vagrant-hostmanager) for this, but we don't *require* it (and I'd prefer if we wouldn't in future).

ehelms:
@iNecas please revisit :)

iNecas:
Btw. I'm using this fully outside of the vagrant environment, and despite it's hackish, it's the simplest and most versatile solution that works everywhere (using it in libvirt na ovirt environments).

iNecas:
Should I invest into addressing the code comments, or are we in the meta-discussion right now whether this should get in in the first place?

evgeni:
Please address the comments, getting rid of the role is a longer task

evgeni:
@iNecas ping :)

ehelms:
@iNecas Any plans to update this?

iNecas:
I most probably will not get to finishing this one. Feel free to take over or close

ehelms:
@evgeni you commented the most on this one, preference on taking it over or closing?

evgeni:
Given I don't like the etc hosts role and would prefer not to extend it, I'm going to close it.

PR #951, opened 2019-03-25
Fixes #950 - Make luna-devel install plugins again

### LINKED ISSUE #950: luna-devel doesn't install plugins
As of 91ed38aae72780782368dfe70eafc851a83f7e19 , luna-devel box does not install various Foreman plugins it should, except for Katello.

## Steps to reproduce:
1. `cp vagrant/boxes.d/99-local.yaml.example vagrant/boxes.d/99-local.yaml`
2. `vagrant up centos7-luna-devel`
3. After box is provisioned, ssh into it and `ls ~`

## Actual results

Only two directories are present: `foreman` and `katello`.  
During provisioning, following will be displayed:

```
ok: [centos7-luna-devel] => changed=false 
  ansible_facts:
    foreman_installer_options_joined: ' --disable-system-checks --katello-devel-scl-ruby=rh-ruby25 --katello-devel-admin-password changeme'
```

## Expected results

All plugins are installed

## Additional info

Playbook luna-devel runs `katello_devel` playbook with some additional variables on top. `katello_devel` playbook essentially runs `katello_devel` role. Since b5e90f9e5a510df65e9b8e518efb34d24657aa86 this role specifies variables inside each dependent role, and since fe0964f4fb04b8ffd5fe0b6b0118a9a9fd0f7a20 it doesn't use vars anymore.

It seems that `foreman_installer_options_internal_use_only` set in luna-devel does not overwrite default value in role, and as a result, is not used.

mirzal:
Move foreman_installer_options_internal_use_only to vars, so it can be overwritten by value specified in playbooks/luna-devel.yml.

### DIFF
--- a/roles/katello_devel/meta/main.yml
+++ b/roles/katello_devel/meta/main.yml
@@ -12,7 +12,3 @@ dependencies:
     foreman_installer_scenario: katello-devel
     foreman_installer_additional_packages:
       - foreman-installer-katello
-    foreman_installer_options_internal_use_only:
-      - "--disable-system-checks"
-      - "--katello-devel-scl-ruby={{ ruby_scl_version }}"
-      - "--katello-devel-admin-password {{ foreman_installer_admin_password }}"
--- a/roles/katello/defaults/main.yml
+++ b/roles/katello/defaults/main.yml
@@ -1,0 +1,9 @@
+---
+foreman_server_repositories_katello: true
+foreman_server_repositories_ostree: true
+foreman_installer_scenario: katello
+foreman_installer_options_internal_use_only:
+  - --disable-system-checks
+  - "--foreman-initial-admin-password {{ foreman_installer_admin_password }}"
+foreman_installer_additional_packages:
+  - katello
--- a/roles/katello/meta/main.yml
+++ b/roles/katello/meta/main.yml
@@ -1,11 +1,3 @@
 ---
 dependencies:
   - role: foreman
-    foreman_server_repositories_katello: true
-    foreman_server_repositories_ostree: true
-    foreman_installer_scenario: katello
-    foreman_installer_options_internal_use_only:
-      - "--disable-system-checks"
-      - "--foreman-initial-admin-password {{ foreman_installer_admin_password }}"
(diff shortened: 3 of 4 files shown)


mirzal:
@evgeni @xprazak2 could you take a look at this and associated issue #950 ?

Bug is legitimate, fix is result of moving things around until bug appears to be fixed :)

roles/katello_devel/meta/main.yml
```diff
@@ -1,4 +1,9 @@
 ---
+vars:
```
evgeni: this will break `katello-devel` pipe, as `vars` don't belong into `meta`.
can you set these in `defaults/main.yml` instead please?

evgeni:
cc @akofink for awareness

roles/katello_devel/meta/main.yml
```diff
@@ -1,4 +1,9 @@
...
+  foreman_installer_options_internal_use_only:
+    - "--disable-system-checks"
```
akofink: quotes unnecessary

roles/katello_devel/meta/main.yml
mirzal: Done in 66ca393f53ff85915206cce67206f80425cd35f9

roles/katello_devel/meta/main.yml
mirzal: Done in 66ca393f53ff85915206cce67206f80425cd35f9

mirzal:
I force pushed new commit 66ca393f53ff85915206cce67206f80425cd35f9 .

It moves variables into `roles/katello/defaults/main.yml` instead of defining them in `vars` in `roles/katello/meta/main.yml`.

I did the same thing for both katello and katello_devel. This should fix both luna.yml and luna-devel.yml playbooks.

johnpmitsch:
@mirzal This is the likely cause of an error in the dev environment `ERROR: Unrecognised option '--foreman-initial-admin-password'`

johnpmitsch:
You can reproduce with `vagrant up centos7-katello-devel` with latest forklift

PR #952, opened 2019-03-25
pass FOREMAN_EXPECTED_VERSION to tests

evgeni:


pipelines/install_pipeline.yml
```diff
@@ -65,5 +65,7 @@
   become: true
   vars_files:
     - vars/{{ pipeline_type }}_base.yml
+  environment:
+    FOREMAN_EXPECTED_VERSION: "{{ foreman_expected_version | default('SKIP') }}"
```
ekohl: Empty strings should work:
```
ekohl@wisse ~ $ var="" ; if [[ -n $var ]] ; then echo test ; fi
ekohl@wisse ~ $ var="x" ; if [[ -n $var ]] ; then echo test ; fi
test
```

evgeni: that's what happens if you multi task hard and try to fix other peoples regex.

PR #957, opened 2019-04-10
Fix katello-devel and luna

mirzal:
Turns out PR #951 broke centos7-katello-devel boxes.

As far as I understand, the problem is in a way Ansible variable precedence works. When variable is defined at role parameter level, it takes precedence over almost everything. But when it is put in `defaults/main.yml` of role, then almost everything takes precedence over it - including `defaults` of dependent roles.

So we can have foreman installer options defined in `defaults` for `katello_devel`, which makes luna playbooks work, but katello-devel fail; or we can have it defined in katello_devel roles params, which makes katello-devel work, but also makes it impossible to override variables at playbook level and breaks luna.

This PR reverts 596ac89e8b5d015b88890bad5c2be41bcc56dd4b and takes another approach at luna playbooks - instead of including katello role, it duplicates it and adds variables as parameters. Main drawback is that we have some roles duplicated, but it seems to actually work this time.

I'm not huge fan of this solution, but this is about as far as my Ansible can take me. If there is better approach, I'm willing to implement it, but I need to be pointed in the right direction.

### DIFF
--- a/playbooks/luna-devel.yml
+++ b/playbooks/luna-devel.yml
@@ -1,16 +1,29 @@
 ---
-- import_playbook: katello_devel.yml
-  vars:
-    katello_devel_github_username: ''
-    foreman_installer_options_internal_use_only:
-      - "--disable-system-checks"
-      - "--katello-devel-github-username '{{ katello_devel_github_username }}'"
-      - "--katello-devel-enable-ostree=true"
-      - "--katello-devel-extra-plugins theforeman/foreman_remote_execution"
-      - "--katello-devel-extra-plugins theforeman/foreman_discovery"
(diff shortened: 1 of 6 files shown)


mirzal:
@johnpmitsch @akofink @evgeni Can you take a look at this PR?

akofink:
There's an issue with candlepin coming up properly during `rake db:seed`, but this PR doesn't cause or fix it. Re-running the installer (or vagrant provision) fixes the issue.

ehelms:
I am proposing https://github.com/theforeman/forklift/pull/958 as a replacement for this. I think its a bit simpler and achieves the goal with more re-use.

ehelms:
Closing given merge of #958

PR #958, opened 2019-04-11
Revert "Fixes #950 - Make luna-devel install plugins again (#951)"

### LINKED ISSUE #950: luna-devel doesn't install plugins

ehelms:
This reverts commit 596ac89e8b5d015b88890bad5c2be41bcc56dd4b.

### DIFF
--- a/playbooks/luna-devel.yml
+++ b/playbooks/luna-devel.yml
@@ -1,8 +1,7 @@
 ---
 - import_playbook: katello_devel.yml
   vars:
-    katello_devel_github_username: ''
-    foreman_installer_options_internal_use_only:
+    foreman_installer_options:
       - "--disable-system-checks"
       - "--katello-devel-github-username '{{ katello_devel_github_username }}'"
       - "--katello-devel-enable-ostree=true"
--- a/roles/katello/defaults/main.yml
+++ b/roles/katello/defaults/main.yml
@@ -1,9 +1,0 @@
----
-foreman_server_repositories_katello: true
-foreman_server_repositories_ostree: true
-foreman_installer_scenario: katello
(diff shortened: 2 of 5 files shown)


mirzal:
I like this approach better.

I provisioned katello-devel, luna and luna-devel boxes and each got right set of parameters in installer.

The only thing I noticed is that luna box got two parameters twice (as they are defined in both `foreman_installer_options` and `foreman_installer_options_internal_use_only`), but it doesn't matter.

PR #966, opened 2019-05-13
Update puppet_repositories to new layout

ekohl:
The Puppet repositories are moving around. Current releases are moved to the topp level while all release-archives.puppet.com is now used for EOL releases.

See https://groups.google.com/forum/#!topic/puppet-dev/19brLt5MuDw for the official announcement.

### DIFF
--- a/roles/puppet_repositories/tasks/puppetlabs-6-redhat.yml
+++ b/roles/puppet_repositories/tasks/puppetlabs-6-redhat.yml
@@ -8,7 +8,7 @@
 
 - name: 'Setup Puppet 6 Repository'
   yum:
-    name: https://yum.puppetlabs.com/puppet6/puppet6-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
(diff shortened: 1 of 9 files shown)


roles/puppet_repositories/tasks/puppetlabs-6-redhat.yml
```diff
@@ -8,7 +8,7 @@
 
 - name: 'Setup Puppet 6 Repository'
   yum:
-    name: https://yum.puppetlabs.com/puppet6/puppet6-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
+    name: https://yum.puppetlabs.com/puppet6-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
```
evgeni: this could be `yum.puppet.com` now?

ekohl: Updated

ekohl:
Added a few more changes.

PR #968, opened 2019-05-15
add a proxy to the upgrade pipeline

evgeni:


ehelms:
Generally looks good

ehelms:
@evgeni is this still blocked?

evgeni:
No, I think this should be OK now

evgeni:
But this might need updates to the latest changes to the installer role @ekohl did

evgeni:
Hmpf, the foreman maintain issue is fixed in 0.4.2+, but that's only in 1.22+ repos.

ekohl:
Perhaps you can split off the hosts changes in a separate PR and merge that so it doesn't get too stale?

evgeni:
yes, I actually might have found a real bug in our proxy testing which is the real source for the f-m breakage :D so testing that before splitting

pipelines/vars/katello_proxy.yml
```diff
@@ -0,0 +1,2 @@
+foreman_installer_additional_packages:
+  - foreman-installer-katello
```
ekohl: Don't we have a `foreman-proxy-content` meta package? Personally I'm ok with dropping that and fully depending on the installer though :)

evgeni:
Yes as it's still broken, working on it

pipelines/upgrade_pipeline.yml
```diff
@@ -31,12 +33,62 @@
...
+- hosts:
+    - "{{ forklift_proxy_name }}"
+  become: yes
+  vars_files:
+    - vars/upgrade_base.yml
+    - vars/{{ pipeline_type }}_proxy.yml
+  vars:
+    foreman_proxy_content_server: "{{ forklift_server_name }}"
+    foreman_installer_disable_system_checks: true
+    foreman_installer_options_internal_use_only:
```
ekohl: Should this be moved to the `_proxy.yml` file as well to mirror the `_server.yml` file?

evgeni: No (well, maybe). `vars_files` has higher priority than `vars`, so if we define it in `proxy.yml`, it gets also passed on upgrades, which is strictly speaking not needed.

ekohl: I was thinking of having a different variable name on the initial install and using that variable in the pipeline on the install step. We can also do it later, but I'd like to add a proxy to the regular foreman as well, even though we'll have to kill the foreman-proxy-content role for it.

evgeni: I'd prefer to defer that to another PR

evgeni:
This PR currently has multiple changes, that are useful on their own:
* make foreman_installer role not set params for proxy
* install additional packages depending on server vs proxy
* slight restructuring of the `hosts` entries to make the later "add proxy" changes easier

I could split them out, or you say you're happy as it is in one PR and we'll merge everything together.

pipelines/install_pipeline.yml
```diff
@@ -29,23 +29,27 @@
     - role: katello
```
ekohl: This is probably also incorrect and might pull in `katello` as a `foreman_installer_additional_packages` on the proxy.

Additionally we should probably skip the client repos on Foreman as well since those are supposed to be optional but that's for another PR.

ehelms:
While I tend not to be a fan of comments over code, the abstract nature of the pipelines is making them harder and harder for me to grok what is happening at any particular step. Could we opt for comments or breaking out groups of plays whereby the name of the file communicates what is happening?

ekohl:
Would it make sense to introduce meta-roles for this? `setup_box`, `pipeline_install` and `pipeline_upgrade` come to mind. Then the upgrade pipeline would be:
* `setup_box` on both boxes
* `pipeline_install` on server
* `pipeline_install` on proxy
* `pipeline_upgrade` on server if initial version != intermediate version
* `pipeline_upgrade` on proxy if initial version != intermediate version
* `pipeline_upgrade` on server if intermediate version != final version
* `pipeline_upgrade` on proxy if intermediate version != final version
* `foreman_testing` on server

Then it's essentially the same install so eventually it could be merged in a single pipeline where install just generates the same version info 3 times, depending on how well it would present the output.

evgeni:
@ekohl I've made the commits more logical, but didn't touch your meta-role idea yet.

@ehelms I've added name attributes to the plays, describing what they do, is that comment-y enough?

ehelms:
@evgeni Yep, thats a great first step and makes things much clearer! Not for this PR, as a near future action item I think it would useful to add in running the proxy tests between each point where the server is at say N and the proxy is at N-1. This would be a first step in supporting N server and N-1 proxies.

evgeni:
Not going to merge this on a Friday, esp because katello nightlies are currently broken because of the ping changes in core :/

ehelms:
Do we have a link to the ping issue?

ekohl:
https://github.com/theforeman/hammer-cli-foreman/pull/394 & https://github.com/Katello/hammer-cli-katello/pull/596 are open to fix them

ehelms:
Current nightly breakages are due to Candlepin 2.6.8 propagation and issue with foreman-proxy-certs-generate (https://community.theforeman.org/t/katello-nightly-rpm-pipeline-210-failed/15515/2)

evgeni:
created #1025 and #1026 for @ekohl's comments and going to merge this now as the pipes are green

PR #969, opened 2019-05-22
updated to check against new login page

ohadlevy:


bats/fb-test-foreman.bats
```diff
@@ -10,7 +10,7 @@ set -o pipefail
 }
 
 @test "check web app is up" {
-  curl -sk "https://localhost$URL_PREFIX/users/login" | grep login-form
+  curl -sk "https://localhost$URL_PREFIX/users/login" | grep login-page
```
ekohl: This needs to work with the old page as well since this is also used in the stable branches.

ekohl:
The content check is there to verify the right page is rendered. Another service could be serving the same URL with a HTTP 200. Granted, it's very light but so far it's served us well.

bats/fb-test-foreman.bats
```diff
@@ -10,7 +10,8 @@ set -o pipefail
 }
 
 @test "check web app is up" {
-  curl -sk "https://localhost$URL_PREFIX/users/login" | grep login-form
+  VERSION=`curl -sk "https://localhost$URL_PREFIX/status" | jq .version`
+  ! grep -v -H -x "$FOREMAN_EXPECTED_VERSION" "$VERSION"
```
ohadlevy: how about this instead?

tbrisker: Interesting approach, i completely forgot this endpoint even exists. Only issue is that in nightlies we don't have `FOREMAN_EXPECTED_VERSION` set iirc, so it would fail to find it.  Perhaps check `status` is `ok` instead since we check the version already earlier?

ohadlevy: @tbrisker updated.

bats/fb-test-foreman.bats
```diff
@@ -10,7 +10,7 @@ set -o pipefail
 }
 
 @test "check web app is up" {
-  curl -sk "https://localhost$URL_PREFIX/users/login" | grep login-form
+  [ `curl -sk "https://localhost$URL_PREFIX/status" | jq .status` == '"ok"' ]
```
ekohl: We don't have jq installed. I also don't think it's present in downstream where we don't have EPEL. That means this will still fail.

Laviro: maybe we can use something like: `curl -sk "http://localhost:5000/users/login" | grep -E 'login-form|login-page'`

tbrisker: https://github.com/theforeman/forklift/pull/970

PR #972, opened 2019-05-23
adjust apt repo grep string for PL repo

mmoll:


mmoll:
this should unbreak the deb pipeline

bats/fb-test-puppet.bats
```diff
@@ -19,7 +19,7 @@ fi
     run grep -q puppetlabs /etc/yum.repos.d/*.repo
     IS_NATIVE=$status
   elif tIsDebianCompatible ; then
-    run grep -q puppetlabs -R /etc/apt/sources.list*
+    run grep -q puppet\.com -R /etc/apt/sources.list*
```
sbernhard: Wouldn't it be better to check for both strings? I saw sources.list in which puppetlabs is used.

mmoll: see GH-966

PR #979, opened 2019-06-10
Add a split puppet pipeline

ekohl:


ekohl:
Includes https://github.com/theforeman/forklift/pull/1030 and https://github.com/theforeman/forklift/pull/1031. It now provisions correctly by using `puppet ssl bootstrap` in ansible.

pipelines/foreman_split_puppet.yml
```diff
@@ -0,0 +1,109 @@
...
+        - "--puppet-server-admin-api-whitelist localhost"
+        - "--puppet-server-admin-api-whitelist {{ forklift_puppet_name }}.{{ ansible_domain }}"
+        - "--puppet-server-admin-api-whitelist {{ forklift_proxy_name }}.{{ ansible_domain }}"
+        - "--puppet-server-ca-client-whitelist localhost"
+        - "--puppet-server-ca-client-whitelist {{ forklift_puppet_name }}.{{ ansible_domain }}"
+        - "--puppet-server-ca-client-whitelist {{ forklift_proxy_name }}.{{ ansible_domain }}"
+        - "--puppet-autosign-entries {{ forklift_server_name }}.{{ ansible_domain }}"
+        - "--puppet-autosign-entries {{ forklift_proxy_name }}.{{ ansible_domain }}"
+
+# TODO: set foreman's trusted hosts to [{{ forklift_puppet_name }}.{{ ansible_domain }}]
```
ehelms: What prevents this TODO?

pipelines/foreman_split_puppet.yml
```diff
@@ -0,0 +1,109 @@
...
+    - vars/puppet_base.yml
+  roles:
+    - role: forklift_versions
+      scenario: "{{ pipeline_type }}"
+      scenario_version: "{{ pipeline_version }}"
+    - role: foreman_installer
+      foreman_installer_options_internal_use_only:
+        - "--no-enable-foreman"
+        - "--no-enable-foreman-cli"
+        - "--no-enable-foreman-proxy"
```
ehelms: I wonder in these situations, would it be cleaner to add a puppet scenario to the installer?

pipelines/foreman_split_puppet.yml
```diff
@@ -0,0 +1,109 @@
...
+  roles:
+    - role: forklift_versions
+      scenario: "{{ pipeline_type }}"
+      scenario_version: "{{ pipeline_version }}"
+    - role: puppet_agent
+      puppet_agent_bootstrap_server: "{{ forklift_puppet_name }}.{{ ansible_domain }}"
+    - role: foreman_installer
+      foreman_installer_options_internal_use_only:
+        - "--no-enable-foreman"
+        - "--no-enable-foreman-cli"
```
ehelms: Great use case for getting https://github.com/theforeman/foreman-installer/pull/262 into the repository

pipelines/foreman_split_puppet.yml
ekohl: Haven't looked at automating it. It could probably be automated by calling `foreman-rake config -- -k trusted_hosts -v ...` but I was also thinking about a cleaner solution so I left it in for now.

ehelms:
Is this a typical deployment scenario? If a user has a separate puppet server, would they spin up a separate server with a smart proxy or use one on board the Foreman server or put one with the puppet server?

pipelines/foreman_split_puppet.yml
ekohl: Agreed, but that requires more changes. Currently no foreman documentation mentions the `--scenario` parameter so we need to go over it all if we start to ship it in `foreman-installer`. However, as we add more of these it's starting to make more sense. https://theforeman.org/manuals/1.23/index.html#3.2.3InstallationScenarios basically repeat a lot of these options.

ekohl:
> Is this a typical deployment scenario? If a user has a separate puppet server, would they spin up a separate server with a smart proxy or use one on board the Foreman server or put one with the puppet server?

Right now it isn't, but this was an interesting experiment to see how well it would work. Now that I know it can work this way, I'm also interested in doing the same in a katello scenario. It also serves as an interesting playground to see how Puppet Enterprise could be integrated into the stack.

PR #980, opened 2019-06-10
Add a test role for Pulp3

jlsherrill:


roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,97 @@
...
+    dest: /etc/httpd/conf.d/pulp_file.conf
+    owner: apache
+    group: apache
+
+- name: Deploy pulp_file.conf for https
+  template:
+    src: templates/pulp_file.conf.j2
+    dest: /etc/httpd/conf.d/05-foreman-ssl.d/pulp_file.conf
+    owner: apache
+    group: apache
```
jlsherrill: I'm thinking the type-specific stuff here will eventually be moved to its own role.

roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,97 @@
...
+    path: '/var/lib/pulp/artifact'
+    state: directory
+    owner: pulp
+    group: pulp
+
+- name: Deploy reverse proxy configuration
+  template:
+    src: templates/pulp-reverse-proxy.conf.j2
+    dest: /etc/httpd/conf.d/05-foreman-ssl.d/pulp-reverse-proxy.conf
+    owner: apache
```
ekohl: Unless apache needs to write to these files I'd say the owner/group should be root. That goes for all these conf.d files.

roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,97 @@
...
+  file:
+    path: /etc/pulp/
+    state: directory
+    owner: apache
+    group: pulp
+
+- name: Configure smart proxy pulp3
+  template:
+    src: templates/pulp3.yml
+    dest: /etc/foreman-proxy/settings.d/pulp3.yml
```
ekohl: This file is managed by the installer so that will conflict. I'd prefer to run the installer with the appropriate parameters so you're also sure that it's re-registered with the right features enabled.

roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,97 @@
...
+  file:
+    path: '/var/lib/pulp/artifact'
+    state: directory
+    owner: pulp
+    group: pulp
+
+- name: Deploy reverse proxy configuration
+  template:
+    src: templates/pulp-reverse-proxy.conf.j2
+    dest: /etc/httpd/conf.d/05-foreman-ssl.d/pulp-reverse-proxy.conf
```
ekohl: Won't the installer wipe these files?

ehelms: Sure. This is not intended to be a final solution. This is intended to give the Pulp 3 developers a working setup for now.

roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,97 @@
...
+    pulp_install_plugins:
+      pulp-file:
+        app_label: "file"
+        source_dir: "git+https://github.com/pulp/pulp_file.git#egg=pulp-file"
+      pulp-docker:
+        app_label: "docker"
+        source_dir: "git+https://github.com/pulp/pulp_docker.git#egg=pulp-docker"
+      pulp-ansible:
+        app_label: "ansible"
+        source_dir: "git+https://github.com/pulp/pulp_ansible.git#egg=pulp-ansible"
```
ehelms: What about setting all of these in defaults like some are currently set?

roles/katello_pulp3/tasks/main.yml
jlsherrill: as @ehelms mentioned, this is more temporary to get developers a way to quickly deploy pulp3

roles/katello_pulp3/tasks/main.yml
jlsherrill: Moving this to defaults didn't actually work, i believe due to https://github.com/ansible/ansible/issues/47023

parthaa:
Seems to work ok. Hit some issues with proxy only mode. Provision a capsule with the following configuration
```
zeta-proxy-devel:
    memory: 4000
    box: centos7
    ansible:
      playbook: 'playbooks/foreman_proxy_content_dev.yml'
      group: 'foreman-proxy-content'
      server: 'zeta'
```
After that provisioning, set up pulp3 on the box by checking out this PR and running
```
ansible-playbook ./playbooks/katello_pulp3.yml -l zeta-proxy-devel
```
Note you get an error in the middle of the provision saying something along ```/etc/httpd/conf.d/05-foreman-ssl.d/ not found```
Create that directory and install seems to run ok after.

Issue2:

 ``` /etc/httpd/conf.d/05-pulp-https.conf``` does not have the following line
```  IncludeOptional "/etc/httpd/conf.d/05-foreman-ssl.d/*.conf"```

Issue 3:
pstatus shows workers as failed
```
pulp-worker@1.service - Pulp RQ Worker
   Loaded: loaded (/usr/lib/systemd/system/pulp-worker@.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Sun 2019-07-07 16:59:47 UTC; 7h ago
  Process: 2855 ExecStart=/usr/local/lib/pulp/bin/rq worker -w pulpcore.tasking.worker.PulpWorker -n reserved-resource-worker-%i@%%h --pid=/var/run/pulp-worker-%i/reserved-resource-worker-%i.pid -c pulpcore.rqconfig (code=exited, status=200/CHDIR)
 Main PID: 2855 (code=exited, status=200/CHDIR)
```
That issue  wont get fixed until I manually create ```/var/run/pulp-worker-1``` and ``` /var/run/pulp-worker-2``` and change the ownership of that dir to ```pulp:pulp```

**Comment by ianballou** [comment]
I've set up a few pulp 3 katello boxes now and I've hit no issues other than the Candlepin Postgres one.

**Comment by johnpmitsch** [comment]
@ekohl @ehelms This is being used pretty extensively by pulp3 devs, since it's been at a standstill for a while, any way we could merge this and work out issues later?

File: ansible.cfg [hunk 6]
```diff
@@ -5,8 +5,9 @@ display_skipped_hosts = false
 verbosity = 1
 nocows = 1
 inventory = $PWD/inventories/
-roles_path = ./playbooks/galaxy_roles:./user_playbooks/roles:./roles
+roles_path = ./playbooks/galaxy_roles:./user_playbooks/roles:./roles:./ansible-pulp/roles
 retry_files_save_path = $HOME/.ansible/retry-files
 remote_user = vagrant
 host_key_checking = False
 force_color = 1
+remote_tmp = /tmp
```
ekohl: Why is this needed?

ansible.cfg
```diff
@@ -5,8 +5,9 @@ display_skipped_hosts = false
 verbosity = 1
 nocows = 1
 inventory = $PWD/inventories/
-roles_path = ./playbooks/galaxy_roles:./user_playbooks/roles:./roles
+roles_path = ./playbooks/galaxy_roles:./user_playbooks/roles:./roles:./ansible-pulp/roles
```
ekohl: I don't like this ansible-pulp specific thing. Could we generalize this somehow to accept any roles? I can also see a use case for foreman-ansible-modules. Ideally it could pull anything from galaxy.

playbooks/katello_pulp3.yml
```diff
@@ -0,0 +1,7 @@
+---
+- hosts: all
+  become: true
+  roles:
+    - katello_pulp3
+  environment:
+    DJANGO_SETTINGS_MODULE: pulpcore.app.settings
```
ekohl: Since you're repeating this, should this be part of the role and a variable?

roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,122 @@
+---
+- name: create static dir
+  file:
+    path: '/var/lib/pulp/static/'
+    state: directory
+    mode: '777'
```
ekohl: Is this really needed?

roles/katello_pulp3/tasks/main.yml
ekohl: Still, we now have 2 ways of doing this. If you do it via the installer, then you also automatically retrigger the registration which ensures the settings in the Foreman DB are also in sync. Here you don't.

roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,130 @@
...
+      pulp-docker:
+        app_label: "docker"
+        source_dir: "git+https://github.com/pulp/pulp_docker.git#egg=pulp-docker"
+      pulp-ansible:
+        app_label: "ansible"
+        source_dir: "git+https://github.com/pulp/pulp_ansible.git#egg=pulp-ansible"
+    pulp_settings:
+      secret_key: secret
+      content_host: "{{ ansible_fqdn }}"
+      databases:
```
sjha4: There are 2 config settings needed for ansible collections: https://github.com/pulp/pulp_ansible/blob/90f8c3fcd974682a43578ca47fee9cc8902007a1/.travis/ansible_playbook.yml

Would it make sense to add it here. The refresh_distribution task fails without these, specifically the ansible_api_hostname setting:
```
ansible_api_hostname: 'http://localhost:24817'
ansible_content_hostname: 'http://localhost:24816/pulp/content'
```

docs/development.md
```diff
@@ -307,3 +307,13 @@ vagrant up centos7-dynflow-devel
 ```
 
 In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
+
+## Pulp3 deployment
+
+To deploy pulp3 onto an exsiting box, within forklift run:
```
ekohl:
```suggestion
To deploy pulp3 onto an existing box, within forklift run:
```

roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,173 @@
...
+      pulp-file:
+        app_label: "file"
+        source_dir: "git+https://github.com/pulp/pulp_file.git#egg=pulp-file"
+      pulp-docker:
+        app_label: "docker"
+        source_dir: "git+https://github.com/pulp/pulp_docker.git#egg=pulp-docker"
+      pulp-ansible:
+        app_label: "ansible"
+        source_dir: "git+https://github.com/pulp/pulp_ansible.git#egg=pulp-ansible"
+#      pulp-rpm:
```
ianballou: add `prereq_role: "pulp.pulp_rpm_prerequisites"` indented under `pulp-rpm`

ianballou: Also want to note that the PR above got merged in.

pipelines/katello_with_pulp3.yml
```diff
@@ -0,0 +1,26 @@
...
+        memory: 10240
+  roles:
+    - forklift
+
+- hosts: katello-with-pulp3
+  become: true
+  roles:
+    - role: forklift_versions
+      scenario: "katello"
+      scenario_version: "3.12"
```
ekohl: Any reason this isn't using nightly? Perhaps at least update it to 3.13 now that it's GA.

playbooks/katello_pulp3.yml
jlsherrill: doesn't seem repeated anymore

roles/katello_pulp3/tasks/main.yml
jlsherrill: agreed its a problem, but this is just a stopgap until we can get packaging and installation via the installer

ansible.cfg
jlsherrill: @ehelms do you know?

ansible.cfg
jlsherrill: @ehelms thoughts?

jlsherrill:
This has been updated and now uses the changes in pulp 2.21 to adjust permissions.  There is one permission still needing adjusting due to a bug in pulp 2.21 that will be fixed in 2.21.1

roles/katello_pulp3/tasks/main.yml
ekohl: Actually, the installer does manage the file and has installer options to set them all. There's no reason to manually manage this file.

roles/katello_pulp3/tasks/main.yml
```diff
@@ -0,0 +1,154 @@
+---
+- name: correct /var/lib/pulp permissions (adding group writable), fixed in 2.21.1
+  file:
+    path: '/var/lib/pulp/'
+    mode: '774'
```
ekohl: 774 is an odd directory mode. Did you intend `775`?

ehelms:
The official installer work for this is starting with building out a [puppet-pulpcore](https://github.com/theforeman/puppet-pulpcore) module. All of the hard work in this PR will be used as input for that design, so thanks for all the work so far and please keep it up to date as you encounter changes so that we account for all the right work arounds and deployment nuances.

ekohl:
Given the work in nightly, is this still relevant?

evgeni:
@jlsherrill given the rather good katello-on-pulp3 coverage in nightly, is the separate test still useful?

jlsherrill:
bye bye PR, don't let the door hit you on the way out !    :wave:

PR #981, opened 2019-06-13
Allow qemu://system, default to qemu://session

akofink:
Related to https://community.theforeman.org/t/vagrant-issues-on-fedora-30/14176

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -232,6 +232,7 @@ module Forklift
         p.memory = box.fetch('memory').to_i * @settings['scale_memory'].to_i if box.fetch('memory', false)
         p.machine_virtual_size = box.fetch('disk_size') if box.fetch('disk_size', false)
         p.management_network_domain = create_domain(box) if p.respond_to?(:management_network_domain)
+        p.qemu_use_session = @settings['libvirt_qemu_use_session'] if @settings.key?('libvirt_qemu_use_session')
 
         box.fetch('add_disks', []).each do |disk|
           type = disk.fetch('type', 'raw')
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -234,16 +234,7 @@ module Forklift
         p.management_network_domain = create_domain(box) if p.respond_to?(:management_network_domain)
         p.qemu_use_session = @settings['libvirt_qemu_use_session'] if @settings.key?('libvirt_qemu_use_session')
 
-        box.fetch('add_disks', []).each do |disk|
-          type = disk.fetch('type', 'raw')
-          device = disk.fetch('device')
-          size = disk.fetch('size')
-          if type.nil? || device.nil? || size.nil?
-            raise "Error in add_disks configuration: type, device or size are missing #{disk}"
-          end
-
-          p.storage :file, :size => size, :type => type, :device => device
-        end
+        add_disks(box, p)
 
         merged_options(box, 'libvirt_options').each do |opt, val|
           p.instance_variable_set("@#{opt}", val)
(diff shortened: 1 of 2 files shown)


vagrant/lib/forklift/box_distributor.rb
```diff
@@ -232,6 +232,7 @@ def configure_libvirt(machine, box, networks = [])
         p.memory = box.fetch('memory').to_i * @settings['scale_memory'].to_i if box.fetch('memory', false)
         p.machine_virtual_size = box.fetch('disk_size') if box.fetch('disk_size', false)
         p.management_network_domain = create_domain(box) if p.respond_to?(:management_network_domain)
+        p.qemu_use_session = @settings.fetch('libvirt_qemu_use_session', false)
```
ekohl: I actually like the session since it requires less privileges. Perhaps we should use the default of the system, but allow overriding?
```suggestion
        p.qemu_use_session = @settings['libvirt_qemu_use_session'] if @settings.key?('libvirt_qemu_use_session')
```

akofink: Wait, how is that different logic than what I've written? If the setting doesn't exist, `fetch` will default to `false`

ekohl: Fedora 30 defaults to true and you don't cover that I think.

akofink: What do you mean? I tested this on F30, and adding the setting in settings.yaml overrides the default of `false`. Are you seeing other behavior?

ekohl: My point is that on Fedora 30 the built in default value is `true`. My reading of the current code is that if you do nothing, it will default to `false`. I would like it to default to the built in value.

parthaa:
Seems to work ok for me. ACK from me

vagrant/lib/forklift/box_distributor.rb
akofink: Ah, I see. I'd like it default to false because that makes it work the way it worked before F30 (no unexpected changes for our forklift users who upgraded from F29). This is especially important if you have a bunch of existing VMs and existing libvirt networks on qemu://system. Also remember that this is all for development, so security isn't really a concern IMO (the only reason you'd want to use qemu://session, afaik).

akofink: @ekohl convinced? ;)

ekohl: No :) My reasoning is that new users who haven't been migrated should work out of the box. Because of the user session you don't get a permission pop up and/or need to be added to the libvirt group

akofink:
Updated with `true` default to use session by default.

akofink:
Refactored a bit for rubocop

vagrant/lib/forklift/box_distributor.rb
```diff
@@ -232,17 +232,9 @@ def configure_libvirt(machine, box, networks = [])
         p.memory = box.fetch('memory').to_i * @settings['scale_memory'].to_i if box.fetch('memory', false)
         p.machine_virtual_size = box.fetch('disk_size') if box.fetch('disk_size', false)
         p.management_network_domain = create_domain(box) if p.respond_to?(:management_network_domain)
+        p.qemu_use_session = @settings.fetch('libvirt_qemu_use_session', true)
```
ekohl: My point was that not every system is using sessions by default, which is why I added the trailing `if ...` so we could have an undefined default. This will break on distros where the session is not set up by default.

akofink:
Updated with the trailing `if`. Thanks for the reviews, @ekohl!

akofink:
This one should be ready for merge @ekohl

akofink:
@parthaa care to ack this one? Seems @ekohl is out or something

akofink:
Updated with documentation in the readme for this setting

ekohl:
Thanks!

PR #984, opened 2019-06-25
Enable ostree plugin in luna

ares:


playbooks/luna.yml
```diff
@@ -21,6 +21,7 @@
       - "--enable-foreman-cli-remote-execution"
       - "--enable-foreman-cli-tasks"
       - "--enable-foreman-cli-templates"
+      - "--enable-ostree"
```
ekohl: Does this work? AFAIK we don't have an ostree class. Don't you mean:
```suggestion
      - "--katello-ostree true"
```

playbooks/luna.yml
```diff
@@ -21,6 +21,7 @@
...
     foreman_installer_additional_packages:
       - tfm-rubygem-foreman_virt_who_configure
```
ekohl: https://github.com/theforeman/foreman-installer/pull/371 should expose these too

ares:
`--katello-enable-ostree` is the right option I believe

playbooks/luna.yml
ares: I agree, can you add it there please? I think I could merge that PR in return :-)

ares:
fyi @rplevka

PR #986, opened 2019-06-27
Workaround for bats org delete failures

johnpmitsch:


ekohl:
https://github.com/Katello/katello/pull/8185 may fix the root cause

PR #988, opened 2019-07-03
remove the released repo when adding stage

evgeni:
otherwise we might have conflicting versions


evgeni:
mergin without travis as it's slooow

PR #989, opened 2019-07-03
Use variables for host names in pipelines

ekohl:


evgeni:
does that work? I've always thought `vars_files` is evaluated *after* the host selection, so the `hosts` entry would have undefined vars.

if it does: yay!

evgeni:
it indeed works

evgeni:
so I tested this, and it somehow deployed a katello on the proxy, instead of a proxy. investigating

pipelines/install_pipeline.yml
```diff
@@ -38,18 +41,21 @@
       scenario_version: "{{ pipeline_version }}"
     - foreman_installer
 
-- hosts: pipeline-{{ pipeline_type }}proxy-{{ pipeline_version }}-{{ pipeline_os }}
+- hosts: "{{ forklift_proxy_name }}"
   become: yes
+  vars_files:
+    - vars/install_base.yml
+    - vars/{{ pipeline_type }}_base.yml
```
evgeni: this overrides the scenario selection.

evgeni: yepp, re-running the pipe with the two type lines removed and it passes

ekohl: Removed now but perhaps we shouldn't set variables directly but rather have a `server_box_installer_parameters`. This would also allow us to use the same pipeline for both Foreman and Katello. Leaving that for a future iteration.

PR #992, opened 2019-07-15
Correct upgrade box names

ekohl:
45ada03a29a67d1458e814022dfe9e3eec46a3ec unified the box naming, but failed to define the proxy pipeline for upgrades. It also includes the -server and -proxy to be consistent with install pipelines.


tbrisker:
Looks like this led to a new error:
```
ERROR! The field 'hosts' has an invalid value, which includes an undefined variable. The error was: 'forklift_server_name' is undefined

The error appears to be in '/root/forklift/pipelines/upgrade_pipeline.yml': line 9, column 3, but may

- hosts: "{{ forklift_server_name }}"
  ^ here
We could be wrong, but this one looks like it might be an issue with
missing quotes. Always quote template expression brackets when they
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"

ekohl:
Where do you see this? I ran it locally and it could spin up a katello instance.

tbrisker:
https://ci.centos.org/job/foreman-katello-upgrade-nightly-test/475/console

ekohl:
One of those days that you missed a line in `git add` so locally it did work :( https://github.com/theforeman/forklift/pull/993 should really fix it.

PR #997, opened 2019-07-21
Always assign env to taxonomies in puppet bats

tbrisker:
Katello upgrade tests fail because the production environment doesn't
match the host taxonomies.


ekohl:
Ideally this would also be limited to versions since we shouldn't need this workaround, but we can leave that for the future.

PR #999, opened 2019-07-22
Use --no-colors in foreman installer

akofink:
Ansible's colors conflict with Foreman's. It's really obvious in the Jenkins output.

image

### DIFF
--- a/roles/foreman_installer/defaults/main.yml
+++ b/roles/foreman_installer/defaults/main.yml
@@ -8,6 +8,7 @@ foreman_installer_scenario_flag: '{{ "--scenario" if foreman_installer_scenario 
 foreman_installer_upgrade: False
 foreman_installer_command: foreman-installer
 foreman_installer_custom_hiera: ""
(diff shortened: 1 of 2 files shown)


pipelines/vars/katello_devel.yml
```diff
@@ -13,4 +13,4 @@ foreman_installer_additional_packages:
 foreman_installer_options_internal_use_only:
   - "--disable-system-checks"
   - "--katello-devel-enable-ostree=true"
```
ekohl: You could drop this since it's the default

akofink: But out of scope for this PR IMO

ekohl: True

akofink:
Updated to add `--no-colors` to foreman installer by default

akofink:
@ekohl good to merge?

akofink:
Updated (sorry for the delay)

PR #1000, opened 2019-07-30
Update versions config for 1.23

tbrisker:
Also dropped long-outdated versions

### DIFF
--- a/vagrant/config/versions.yaml
+++ b/vagrant/config/versions.yaml
@@ -1,22 +1,4 @@
 installers:
-  - foreman: '1.17'
-    katello: '3.6'
-    pulp: '2.15'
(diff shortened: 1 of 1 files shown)


tbrisker:
brought back 1.19

vagrant/config/versions.yaml
```diff
@@ -29,8 +11,8 @@ installers:
       - 'ubuntu1804'
 
   - foreman: '1.20'
-    katello: '3.9'
```
ekohl: The same goes for Katello 3.11: it needs Katello 3.9 in upgrades.

tbrisker: restored 3.9 too

PR #1003, opened 2019-08-02
Shut down puppet during host registration

jlsherrill:
very often we get unlucky and a puppet checkin happens
while the host is unregistered resulting in the inability
to register the hsot.  Shut down puppet during this time
to work around the issue

### DIFF
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -121,6 +121,11 @@ EOF
   tPackageExists subscription-manager || tPackageInstall subscription-manager
 }
 
+@test "disable puppet agent to prevent checkin from registering host to another org" {
+  systemctl is-active puppet || skip "Puppet is not active"
+  systemctl stop puppet
+}
+
 @test "delete host if present" {
   hammer host delete --name="`hostname`" || echo "Could not delete host"
 }
@@ -156,6 +161,11 @@ EOF
   echo "rc=${status}"
   echo "${output}"
   subscription-manager list --consumed | grep "${PRODUCT}"
+}
+
+@test "start puppet again" {
+  systemctl is-enabled puppet || skip "Puppet isn't enabled"
+  systemctl start puppet
 }
 
 @test "check content host is registered" {


bats/fb-content-katello.bats
```diff
@@ -158,6 +162,10 @@ EOF
   subscription-manager list --consumed | grep "${PRODUCT}"
 }
 
+@test "start puppet again" {
+ run systemctl start puppet
```
ekohl: Same here: probably needs `systemctl is-enabled`
```suggestion
  run systemctl start puppet
```

bats/fb-content-katello.bats
```diff
@@ -121,6 +121,10 @@ EOF
   tPackageExists subscription-manager || tPackageInstall subscription-manager
 }
 
+@test "disable puppet agent to prevent checkin from registering host to another org" {
+  run systemctl stop puppet
```
ekohl: This will fail if Puppet is not present and break. Probably needs some `systemctl is-active` love

jlsherrill:
@ekohl updated, this should be good to go and passed a local test

bats/fb-content-katello.bats
```diff
@@ -121,6 +121,12 @@ EOF
   tPackageExists subscription-manager || tPackageInstall subscription-manager
 }
 
+@test "disable puppet agent to prevent checkin from registering host to another org" {
+  if systemctl is-active puppet; then
+    run systemctl stop puppet
```
ekohl: This ignores the output. Is that what we want?

bats/fb-content-katello.bats
```diff
@@ -121,6 +121,12 @@ EOF
   tPackageExists subscription-manager || tPackageInstall subscription-manager
 }
 
+@test "disable puppet agent to prevent checkin from registering host to another org" {
+  if systemctl is-active puppet; then
```
ekohl: Maybe it's nice to clearly show if the service was running or not (same with starting):

```suggestion
  systemctl is-active puppet || skip "Not running"
```

ekohl:
An alternative is to use `setup` and `teardown` for this.

bats/fb-content-katello.bats
jlsherrill: i thought 'run' would capture the output if it failed, i'm not sure i get what you mean?

ekohl: https://github.com/sstephenson/bats#run-test-other-commands

Run allows you to use test a programs output, regardless of its exit code. It'll always exit with code 0 and stores the result in `$status` and the output in in `$output`. Since the code is 0, this test will silently fail. If you drop the `run`, then it'll automatically capture the output on failure in the test results.

jlsherrill:
thanks @ekohl updated

jlsherrill:
I'll run one more test and merge if its good

PR #1004, opened 2019-08-02
Allow disabling of forced private network

jlsherrill:
With the recent addition of qemu sessions into vagrant-libvirt
nfs shared folders does not work with private networks, so
we need a way to disable it

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -89,6 +89,7 @@ module Forklift
         configure_openstack_provider(machine, box)
         configure_google_provider(machine, box)
         configure_synced_folders(machine, box)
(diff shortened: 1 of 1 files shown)


jlsherrill:
I'm honestly not sure what the purpose of that line is, just that it is causing me issues now on fedora 30.  Would we prefer to just remove it?

jlsherrill:
its also worth noting that the 'nfs' option doesn't do this (although i use synced_folders for support for more than 1), so maybe we can juts remove it?

jlsherrill:
Updated to only configure the private network if private_ip is set, and removed the dependency on synced_folders with private_networking.  I assume if someone sets a private_ip they want a private network?    Let me know what you think

ekohl:
Makes much more sense to me.

PR #1005, opened 2019-08-06
Add integration tests using smoker

ekohl:
This adds a Fedora 30 box to the tests. This is chosen because it allows testing a remote scenario and guarantees it doesn't use anything on the system-under-test. This means it can also be used with containers or load balanced setups. It's also easy because we have Python 3, Chromium and Chromedriver all packaged up.

This is WIP because it's only the install pipeline and even then only tested on Foreman. It should be added to Katello as well. It also depends on https://github.com/theforeman/smoker/pull/2 being merged.


roles/smoker/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: 'Install browser'
+  become: true
+  package:
+    name: '{{ smoker_browser_packages }}'
+
+- name: 'Install Smoker via pytest_project'
```
evgeni: What's the reasoning behind importing separate task files from the role instead of the whole role?

ekohl: Both Robottelo and smoker need to do additional configuration. Technically smoker could drop its variables.json anywhere since it's an absolute path but Robottelo modifies files in the cloned repo. Since I started with Robottelo, I continued this way. And do sort of a callback where I provided a role/task with setup was even worse IMHO. If you know a clean way to do this in Ansible, I'd love to hear it.

ehelms:
Consider splitting out the introduction of pytest_project to its own PR to make consumption of the smoker aspect easier.

ehelms:
Note this will add some amount of overhead to our pipelines. Instead of a single box (centos 7) being downloaded, every test run must now download two test boxes.

ehelms:
My current error is:

```
TASK [pytest_project : Install requirements] ************************************************************
fatal: [pipeline-katello-smoker-nightly-centos7]: FAILED! => changed=false 
  cmd:
  - /home/vagrant/smoker/venv/bin/pip3
  - install
  - -r
  - /home/vagrant/smoker/requirements.txt
  msg: |2-
  
    :stderr: Could not open requirements file: [Errno 2] No such file or directory: '/home/vagrant/smoker/requirements.txt'
    You are using pip version 19.0.3, however version 19.2.3 is available.
    You should consider upgrading via the 'pip install --upgrade pip' command.
```

ekohl:
Indeed. I've merged the PR now to make integration easier.

ehelms:
To collect the test ouput, we really only need a small change to our collection playbook:

```
diff --git a/playbooks/collect_debug.yml b/playbooks/collect_debug.yml
index 1944067..d3d19b1 100644
--- a/playbooks/collect_debug.yml
+++ b/playbooks/collect_debug.yml
@@ -3,7 +3,7 @@
   become: true
   vars:
     bats_output_dir: '/root/bats_results'
-    bats_remote_dir: '/tmp/debug'
+    remote_dir: '/tmp/debug'
   roles:
     - sos_report
   tasks:
@@ -16,5 +16,10 @@
     - name: "Copy bats results"
       fetch:
         src: "{{ item.path }}"
-        dest: "{{ bats_remote_dir }}"
+        dest: "{{ remote_dir }}"
       with_items: "{{ bats_results.files }}"
+
+    - name: "Copy smoker results"
+      fetch:
+        src: "{{ ansible_env.HOME }}/smoker/junit.xml"
+        dest: "{{ remote_dir }}"
```

Now, I agree this does tie this role to the default location of smoker but you'll note we do the same for sos_report and bats so I think it's a fair compromise to get this off the ground. There is also a small change needed in the CI jobs:

https://github.com/theforeman/foreman-infra/pull/1168

ekohl:
So I've updated this. The box is now a centos7 box and uses the EL7.7 Python 3 to run its tests. This avoids the need to pull in an unrelated Fedora box. Since we're running within CentOS CI, it'll also be quickly available even when testing on Debian-based installs.

Currently the nightly pipelines are rather broken due to NodeJS changes so I'm not sure we should introduce more instability. Otherwise I'd like to merge this now.

ehelms:
Generally good, I as you say would wait to merge till we have a green nightly so we can isolate the changes.

pipelines/vars/install_base.yml
```diff
@@ -1,3 +1,7 @@
 forklift_name: "pipeline-{{ pipeline_type }}-{{ pipeline_version }}-{{ pipeline_os }}"
 forklift_server_name: "pipeline-{{ pipeline_type }}-server-{{ pipeline_version }}-{{ pipeline_os }}"
 forklift_proxy_name: "pipeline-{{ pipeline_type }}-proxy-{{ pipeline_version }}-{{ pipeline_os }}"
+forklift_smoker_name: "pipeline-{{ pipeline_type }}-smoker-{{ pipeline_version }}-{{ pipeline_os }}"
+smoker_box:
```
evgeni: this could be defined in `base.yml`, no? or do we need different smoker boxes for install and upgrade?

pipelines/vars/upgrade_base.yml
```diff
@@ -1,4 +1,8 @@
 forklift_name: "pipeline-upgrade-{{ pipeline_type }}-{{ pipeline_version}}-{{ pipeline_os }}"
 forklift_server_name: "pipeline-upgrade-{{ pipeline_type }}-{{ pipeline_version}}-{{ pipeline_os }}"
 forklift_proxy_name: "pipeline-upgrade-{{ pipeline_type }}-proxy-{{ pipeline_version}}-{{ pipeline_os }}"
-forklift_boxes: "{{ {forklift_server_name: server_box} }}"
+forklift_smoker_name: "pipeline-{{ pipeline_type }}-smoker-{{ pipeline_version }}-{{ pipeline_os }}"
+forklift_boxes: "{{ {forklift_server_name: server_box, forklift_smoker_name: smoker_box} }}"
+smoker_box:
```
evgeni: same comment as in install_base.yml

pipelines/vars/install_base.yml
ekohl: I don't see a `base.yml`. Are you suggesting to introduce one?

PR #1011, opened 2019-08-19
Revert "Install virt-who via the installer"

ekohl:
This reverts commit c270c58125cb75868272b2001e017ddc65dd34c4 which is needed because the luna pipeline in upgrades fails because 1.22 doesn't have this option yet. https://github.com/theforeman/foreman-installer/pull/380 is open for that.


evgeni:
as discussed on IRC: NACK
the fix will land in 1.22.1 soon (a week) and luna failures don't block anything at the moment, so I think it's OK to have a red upgrade pipe until we release 1.22.1

PR #1012, opened 2019-08-23
use unsafe writes for forklift boxes

evgeni:
this has a ~2× speedup in my tests (from ~60 to ~30 minutes) and should
be fine as those are throw-away VMs anyways


tbrisker:
how does this affect people using forklift devel boxes for a longer lifetime?

evgeni:
Only if they use setup forklift playbook and their system is crashing. Which imho should be seldom.

PR #1015, opened 2019-09-05
Skip available errata bats test until CentOS has newer subscription-m…

ehelms:
…anager

### DIFF
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -175,6 +175,7 @@ setup() {
 }
 
 @test "check available errata" {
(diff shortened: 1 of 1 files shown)


bats/fb-content-katello.bats
```diff
@@ -175,6 +175,7 @@ setup() {
 }
 
 @test "check available errata" {
+  skip
```
ekohl: Generally I like to have a comment why it's skipped. Ideally a BZ so we can see when it's fixed.

ehelms: Added

PR #1017, opened 2019-09-06
Fixes #1013 - Proxy dev fails using foreman start

### LINKED ISSUE #1013: Foreman proxy dev server doesn't work with foreman start
When you spin up a `centos7-proxy-devel` box, it doesn't work with `foreman start`. 

We [check the `server.pid` file](https://github.com/theforeman/forklift/blob/8b95b1a2d9cb78e2675025400b748344d23e14ea/roles/foreman_proxy_content/tasks/devel_install.yml#L3) which is not created with foreman start, so the playbook fails because "no server is running"

ianballou:
When starting Foreman with `bundle exec foreman start`, the server.pid file is not created for some reason.  This causes the `foreman_proxy_content_dev.yml` playbook to fail.  To fix this, we can also check for a running puma process with the [foreman] tag.

### DIFF
--- a/roles/foreman_proxy_content/tasks/devel_install.yml
+++ b/roles/foreman_proxy_content/tasks/devel_install.yml
@@ -1,15 +1,18 @@
 ---
-- name: 'get existing rails server pid'
-  command: cat {{ base_foreman_directory }}tmp/pids/server.pid
-  register: existing_rails_server
-  become_user: vagrant
-  ignore_errors: True
+- name: 'Get Server Hostname'
+  shell: hostname -f
   delegate_to: "{{ foreman_proxy_content_server }}"
+  register: server_fqdn
 
-- name: 'check for rails server'
-  fail: msg="No rails server detected running on {{ foreman_proxy_content_server }},
-             please run `rails s` on {{ foreman_proxy_content_server }}"
-  when: existing_rails_server.rc != 0
+- name: 'Query Foreman server'
+  uri:
+    url: "{{ 'https://' ~ server_fqdn.stdout }}"
+  ignore_errors: True
+  register: foreman_get
+
+- name: 'Check for rails server'
+  fail: msg="No rails server detected running on {{ foreman_proxy_content_server }}"
+  when: foreman_get.status != 200
 
 - name: 'Add group foreman'
   group: name=foreman state=present


roles/foreman_proxy_content/tasks/devel_install.yml
```diff
@@ -6,10 +6,17 @@
   ignore_errors: True
   delegate_to: "{{ foreman_proxy_content_server }}"
 
+- name: 'get existing rails server puma'
+  shell: ps -ef | grep -w "\[foreman\]"
```
ekohl: Could you rewrite this using [pgrep](https://linux.die.net/man/1/pgrep)?

ianballou:
@ekohl yeah I agree -- it would make more sense to just check the Foreman URL.  I'll try that out instead.

ianballou:
@ekohl I've updated it

roles/foreman_proxy_content/tasks/devel_install.yml
```diff
@@ -1,15 +1,18 @@
...
-  become_user: vagrant
-  ignore_errors: True
+- name: 'Get Server Hostname'
+  shell: hostname -f
   delegate_to: "{{ foreman_proxy_content_server }}"
+  register: server_fqdn
+
+- name: 'Query Foreman server'
+  uri:
+    url: "{{ 'https://' ~ server_fqdn.stdout }}"
```
ekohl: I'm wondering if we should set this as a variable/fact and use it later on. You can even move it to `install.yaml` and share it. It would even be nicer if the URL wasn't guessed but actually retrieved from the `foreman_url` setting.

ianballou: I did another task that was doing the same thing to get the server name, so saving the value higher up sounds good.  I'll look into retrieving that url value.

johnpmitsch:
> What is even the point of this task? This makes zero sense to me because the remote Foreman server could be a production setup as well. 

@ekohl this is for the devel install tasks, which is for a development setup where you have to manually start the server, which is why we added this check.

I do agree that we should check if its running based on the URL to keep this not tied with the how the dev starts the server

I'll test out soon!

ianballou:
Revisiting after being gone last week, I've tried a few ways of saving and sharing the fact in `install.yaml` but it doesn't end up being accessible for whatever reason.  I'm thinking for now that this solution will cover our issues with using the proxy dev box, but we can revisit and re-factor later.

ianballou:
Looks like just running `curl http://localhost:3000/api/v2/ping` on the Foreman server box is a much simpler way to see if the server is running without having to build the FQDN url.  It's tested to work when HTTPS is in use too.  @ekohl do you have any thoughts about checking like this?

ekohl:
Oh, I now saw your comment. Yes, the proxy is probably ok as well and that's already explicitly configured. Something other things rely on anyway. I'd be happy to merge a follow up PR.

PR #1027, opened 2019-09-26
Move foreman-initial-admin-password to install.yaml

ekohl:
Rather than keeping it as a default value, this moves it to install.yaml which has more information about whether it really needs to be there.  Besides scenarios that are not foreman-proxy-content (which doesn't include the foreman class), it also doesn't make sense on upgrades since it's the initial password.

A nice benefit is that it isn't spread over the various playbooks anymore.

### DIFF
--- a/pipelines/upgrade_pipeline.yml
+++ b/pipelines/upgrade_pipeline.yml
@@ -95,6 +95,7 @@
     - vars/{{ pipeline_type }}_server.yml
   vars:
     foreman_installer_upgrade: True
+    foreman_installer_options_internal_use_only: []
   roles:
     - role: forklift_versions
       scenario: "{{ pipeline_type }}"
@@ -119,6 +120,7 @@
     - vars/{{ pipeline_type }}_proxy.yml
   vars:
     foreman_installer_upgrade: True
+    foreman_installer_options_internal_use_only: []
     foreman_proxy_content_upgrade: True
     foreman_proxy_content_server: "{{ forklift_server_name }}"
   roles:
@@ -141,6 +143,7 @@
   become: yes
   vars:
     foreman_installer_upgrade: True
+    foreman_installer_options_internal_use_only: []
   vars_files:
     - vars/upgrade_base.yml
     - vars/{{ pipeline_type }}_base.yml
@@ -164,6 +167,7 @@
   become: yes
   vars:
     foreman_installer_upgrade: True
+    foreman_installer_options_internal_use_only: []
     foreman_proxy_content_upgrade: True
     foreman_proxy_content_server: "{{ forklift_server_name }}"
   vars_files:
--- a/pipelines/upgrade_pipeline.yml
+++ b/pipelines/upgrade_pipeline.yml
@@ -93,9 +93,6 @@
     - vars/{{ pipeline_type }}_base.yml
     - vars/{{ pipeline_type }}_base_release.yml
     - vars/{{ pipeline_type }}_server.yml
-  vars:
-    foreman_installer_upgrade: True
-    foreman_installer_options_internal_use_only: []
   roles:
     - role: forklift_versions
       scenario: "{{ pipeline_type }}"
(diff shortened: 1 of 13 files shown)


roles/foreman_installer/tasks/install.yml
```diff
@@ -12,6 +12,11 @@
 - debug:
     msg: "Foreman installer version {{ foreman_installer_version_file['content'] | b64decode }}"
 
+- name: 'Set initial admin password'
+  set_fact:
+    foreman_installer_options_internal_use_only: "{{ [ '--foreman-initial-admin-password', '{{ foreman_installer_admin_password }}' ] + foreman_installer_options_internal_use_only }}"
+  when: foreman_installer_scenario != "foreman-proxy-content" && not foreman_installer_upgrade|bool
```
evgeni:
```suggestion
  when:
    - foreman_installer_scenario != "foreman-proxy-content" 
    - && not foreman_installer_upgrade|bool
```

ekohl: I am locally running into this as well which is exactly why this is a draft :)

ekohl:
I think this is ready for review now.

pipelines/upgrade_pipeline.yml
```diff
@@ -106,6 +104,8 @@
     - foreman_server_repositories
     - foreman_client_repositories
     - role: foreman_installer
+      foreman_installer_upgrade: True
+      foreman_installer_options_internal_use_only: []
```
evgeni: why do you reset all internal options?

ekohl: One problem I kept running into was that it's adding the initial install options to proxies on upgrades. Since we don't change options on upgrade (and the actual roles add the upgrade statements), I thought we should clear them out.

evgeni: Feels like the role should ignore them then?

ekohl: Not always. In the future we might want to add testing scenarios where we enable a module that was introduced in 1.next so the role is not the right layer.

ekohl: Btw, another solution I thought of was to change the param to `server_installer_options` and explicitly set that in the initial install. Perhaps we're also getting to a point where a simulated inventory is actually nice. I'm not sure our vagrant inventory plugin exposes the variables, but that might be a better place.

evgeni: The problem with moving these to the inventory is that the inventory is evaluated for the whole playbook once, so it can't change mid-way (there are ways to refresh it, but I'm not a fan).

ekohl: But we can store custom variables. `server_installer_options` is not used elsewhere. In the playbook during the initial run we can do `foreman_installer_options_internal_use_only: "{{ server_installer_options }}"` just where we need it.

ekohl:
I don't remember if this PR actually passed. I think a bats test failed when I ran it but didn't investigate.

evgeni:
> I don't remember if this PR actually passed. I think a bats test failed when I ran it but didn't investigate.

Oh, OK, then I'll run a few tests before merging.

evgeni:
So I tried this, and it failed :)

```
TASK [foreman_installer : set_fact] ok: [pipeline-upgrade-katello-3.13-centos7] => changed=false 
  ansible_facts:
    - --upgrade
    - --certs-update-all

TASK [foreman_installer : Read installer version] 
ok: [pipeline-upgrade-katello-3.13-centos7] => changed=false 
  content: MS4yMi4xCg==
  encoding: base64
  source: /usr/share/foreman-installer/VERSION

TASK [foreman_installer : debug] 
ok: [pipeline-upgrade-katello-3.13-centos7] => 
  msg: |-
    Foreman installer version 1.22.1

TASK [foreman_installer : Clear initial admin password option] ***************************************************************************************************************************************************************************************************************************
ok: [pipeline-upgrade-katello-3.13-centos7] => changed=false 
  ansible_facts:
    foreman_installer_admin_password_option: ''

TASK [foreman_installer : Join options] ok: [pipeline-upgrade-katello-3.13-centos7] => changed=false 
  ansible_facts:
    foreman_installer_options_joined: ' '

TASK [foreman_installer : Run installer] 
fatal: [pipeline-upgrade-katello-3.13-centos7]: FAILED! => changed=true 
  cmd: |-
    foreman-installer -v --no-colors --disable-system-checks --scenario katello
```

It called the installer without --upgrade, and boom :)

ekohl:
Closing since another solution was implemented. I'm also thinking that we shouldn't add module specific parameters to the role.

PR #1029, opened 2019-10-03
Update production.md

Zefool:
Without this step, deployment will likely fail. At least my boxes don't have a vagrant user.


ekohl:
Thanks!

PR #1030, opened 2019-10-05
Update vagrant inventory to return groups

ekohl:
According to the documentation this avoids individual calls to the inventory script with --host improving performance.


evgeni:
test failure unrelated

PR #1031, opened 2019-10-05
Add a puppet_agent role

ekohl:


roles/puppet_agent/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+    state: "{{ puppet_group_state }}"
+  when: puppet_group_manage
+
+- name: puppet ssl bootstrap
+  when: puppet_bootstrap_manage|bool
+  block:
+    # This makes sure 'sudo puppet' works which is needed for bootstrapping
+    - name: setup sudo secure_path
+      copy:
+        content: "Defaults secure_path = \"/sbin:/bin:/usr/sbin:/usr/bin:/opt/puppetlabs/bin\"\n"
```
evgeni:
```suggestion
        content: 'Defaults secure_path = "/sbin:/bin:/usr/sbin:/usr/bin:/opt/puppetlabs/bin"\n'
```

roles/puppet_agent/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+  when: puppet_package_manage
+
+- name: ensure puppet group
+  group:
+    name: "{{ puppet_group_name }}"
+    state: "{{ puppet_group_state }}"
+  when: puppet_group_manage
+
+- name: puppet ssl bootstrap
+  when: puppet_bootstrap_manage|bool
```
evgeni: you use `|bool` here, but not with the others? why?

roles/puppet_agent/tasks/main.yml
ekohl: I tried that, but then it literally outputs a `\n` in the file and you break sudo.

roles/puppet_agent/tasks/main.yml
ekohl: Likely a leftover because I wasn't sure if it was needed or not. Ansible variable handling is the worst.

roles/puppet_agent/tasks/main.yml
evgeni: ugh. ok

roles/puppet_agent/tasks/main.yml
ekohl: Fixed

roles/puppet_agent/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
```
ehelms: Our typical best practice has been to scope all variables here with the role name. I think switching them from `puppet_` to `puppet_agent` would be good.

roles/puppet_agent/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
...
+puppet_package_name: puppet-agent
+puppet_package_state: installed
+puppet_package_manage: true
+
+puppet_group_name: puppet
```
ehelms: What other group names might someone use?

roles/puppet_agent/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
...
+puppet_group_state: present
+puppet_group_manage: true
```
ehelms: What situations lead to not wanting this group managed?

roles/puppet_agent/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
...
+
+puppet_bootstrap_manage: true
+puppet_bootstrap_server: ""
```
ehelms: You can leave this empty and do a `is not defined` check. That is how I typically see it in Ansible.

roles/puppet_agent/defaults/main.yml
ekohl: Good point. This is a leftover because the role started as `puppet` but I decided `puppet_agent` was more appropriate.

roles/puppet_agent/defaults/main.yml
ekohl: I strongly dislike hardcoded values.

roles/puppet_agent/defaults/main.yml
ekohl: Again, I like things to be configurable. It's from my Puppet experience: configurable but a sane default.

roles/puppet_agent/defaults/main.yml
ekohl: I dislike that pattern because then I don't know what variables I can change. It's also how foreman_ansible discovers variables: read defaults.

ekohl:
@ehelms I've renamed variables. Please have a look again.

roles/puppet_agent/defaults/main.yml
ehelms: Right, what I meant was you can leave the value empty:

```
puppet_bootstrap_server:
```

roles/puppet_agent/defaults/main.yml
ehelms: The downside is over specifying configuration. In part my questions are to make sure we focus on specifying what users should be configuring, and usings `vars` to put things that are internally specified without hardcoding the values. I tend to also find that over specifying configuration leads to hard to read code. So in this case, I am asking myself, do people change the puppet_agent group? Do they ever want to not manage it? This stuff also seems like it tends to be rarely if ever changed and thus a sort of pre-optimization.

PR #1032, opened 2019-10-06
Update travis configuration

ekohl:
This now uses a bionic image which has a package for ansible-lint. The base image also stopped shipping ruby and bundler so we explicitly install it.

It also adds two commits with changes to vagrant to ensure the Ruby tests are triggered. These changes by themselves are also useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,12 @@
-sudo: false
 language: python
 python:
   - 2.7
-install:
-  - |
-    if git diff origin/master --name-only | grep -q "vagrant"; then
-      rvm install 2.2.4
-    fi
-
-  - |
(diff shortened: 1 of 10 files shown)


ekohl:
There is a preinstalled Ruby 2.4 to 2.6 which could also be used but it now takes about 10 to 15 seconds to install ruby + ansible-lint.

.travis.yml
```diff
@@ -1,17 +1,12 @@
...
-    fi
+dist: bionic
+addons:
+  apt:
+    packages:
+      - ansible-lint
+      - ruby-bundler
 script:
   - |
     if git diff origin/master --name-only | grep -q "vagrant"; then
```
evgeni: imho we should run these unconditionally too, but out of scope here

PR #1035, opened 2019-10-17
Add a rpm-packaging box

mdellweg:
This box installs the dependencies to run add_gem_package.sh and clones
foreman-packaging in the users context.

### DIFF
--- a/playbooks/rpm_packaging.yml
+++ b/playbooks/rpm_packaging.yml
@@ -1,0 +1,38 @@
+---
+- hosts: all
+  tasks:
+    - name: Install dependencies for packageing tooling
+      yum:
+        name: "{{ package_list }}"
+        state: present
+      vars:
+        package_list:
+          - git-core
+          - git-annex
+          - rubygem-nokogiri
+          - rpmspectool
+          - rubygem-gem2rpm
+          - make
+          - gcc
+          - redhat-rpm-config
+          - ruby-devel
+          - curl-devel
+          - gem
+          - python3-click
+          - python3-ruamel-yaml
+          - crudini
+      become: true
+
+    - name: Install gem-compare
+      gem:
+        name: gem-compare
+        user_install: false
+        state: present
+      become: true
+
+    - name: Clone foreman-packaging
+      git:
+        repo: https://github.com/theforeman/foreman-packaging.git
+        dest: /home/vagrant/foreman-packaging
+        version: rpm/develop
+...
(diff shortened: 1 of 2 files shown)


playbooks/rpm_packaging.yml
```diff
@@ -0,0 +1,38 @@
...
+          - redhat-rpm-config
+          - ruby-devel
+          - curl-devel
+          - gem
+          - python3-click
+          - python3-ruamel-yaml
+          - crudini
+      become: true
+
+    - name: Install gem-compare
```
ekohl: I use https://copr.fedorainfracloud.org/coprs/ekohl/gem-compare/ and saves the need for having `gem` and all `-devel` packages.

playbooks/rpm_packaging.yml
```diff
@@ -0,0 +1,38 @@
+---
+- hosts: all
+  tasks:
+    - name: Install dependencies for packageing tooling
```
ekohl:
```suggestion
    - name: Install dependencies for packaging tooling
```

playbooks/rpm_packaging.yml
```diff
@@ -0,0 +1,32 @@
+---
+- hosts: all
+  tasks:
+    - name: Enable Fedora Copr for gem-compair
```
ehelms: s/gem-compair/gem-compare

playbooks/rpm_packaging.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Install dependencies for packaging tooling
+      yum:
+        name: "{{ package_list }}"
+        state: present
+      vars:
+        package_list:
+          - git-core
+          - git-annex
+          - rubygem-nokogiri
+          - rubygem-gem-compare
```
ekohl: Looks like I locally depend on an ancient package still installed from Fedora 24 and COPR doesn't have repoclosure so it currently fails.

mdellweg: Yes. The question is, should we update the rpm, or drop the top commit from this PR?

ekohl: I'd also be ok changing the bumping script to not hard depend on gem-compare. Its primary benefit to me has been showing differences in the requirements. When we automatically update the dependencies, the biggest benefit is seeing if files were added/removed. The build will fail in most cases.

ehelms:
What is the resolution here?

mdellweg:
I can still remove the last commit. It will compile gem-compare then.
If there is a benefit, i think, we should keep the gem-compare dependency.

PR #1037, opened 2019-10-23
add Debian/buster

mmoll:


mmoll:
untested and I have no idea, if there's anything missing :upside_down_face:

PR #1039, opened 2019-10-23
install gpg on Debian/buster

mmoll:


mmoll:
I guess newer versions of Ansible will auto-handle this with `apt_key`, for now I'd just like to duplicate this to get things going...

PR #1045, opened 2019-11-01
Adding a way to copy custom files to a dev box

johnpmitsch:
This allows devs to specify custom files (like dotfiles) to be copied over to `~/vagrant` when spinning up or provisioning a development environment. These files can be added to `roles/customize_home/files` and they will be copied over. If that folder does not exist, the copy step is skipped, so a developer can choose to ignore this feature too.

To test:

1. create `roles/customize_home/files`
2. Add files there
3. Spin up a `centos7-katello-devel` or `centos7-katello-devel-stable` environment after copying over the box definition
4. Make sure the files are copied over to `~/vagrant` on the devel environment


playbooks/katello_stable.yml
```diff
@@ -0,0 +1,3 @@
+- hosts: all
+  roles:
+    - customize_home
```
ehelms: The name of this playbook seems at odds with what it's actually doing. I would assume this would install a stable Katello.

.gitignore
```diff
@@ -15,6 +15,7 @@ user_playbooks
 playbooks/galaxy_roles
 requirements.yml
 inventories/local_*
+roles/customize_home/files/
```
ehelms: What do you think about instead supporting these two options:

 1) a git repository that can be synced
 2) user specifying a variable file that includes a list of all files on their control node to copy over, that file could live in a `user_variables` folder at the top level to match the `user_playbooks` directory.

playbooks/katello_stable.yml
johnpmitsch: This is for the roles that are run after a katello stable VM is provisioned, any suggestions on other names?

.gitignore
johnpmitsch: > a git repository that can be synced

I could see a use case for the git repository (that is where I keep my custom dotfiles and setup scripts), let me look into that.

> user specifying a variable file that includes a list of all files on their control node to copy over, that file could live in a user_variables folder at the top level to match the user_playbooks directory.

My thought was, to keep things simple, you can just symlink anything on your machine to `/roles/customize_home/files`. I worry about adding too much complexity here before we have gotten any feedback, but I'll think it over.

johnpmitsch:
@ehelms I updated to be able to use a git repository as well. Please take a look at the documentation for instructions on how to use. So now a dev can put files in a specified git repo and/or the local `roles/customize_home/files` path.

This should be flexible enough to cover developer workflows without being overly complicated (and also easy to ignore if you don't want this functionality)

roles/customize_home/tasks/clone_specified_repo.yml
```diff
@@ -0,0 +1,29 @@
+- name: Ensure temporary home directory does not exist
+  file:
+    path: "{{ ansible_env.HOME }}/tmp_custom_home/"
+    state: absent
+
+- name: Clone specified repo into temporary custom home directory
+  git:
+    repo: "{{ custom_home_git_repo }}"
+    dest: "{{ ansible_env.HOME }}/tmp_custom_home/"
```
johnpmitsch: I cloned into a temporary directory because git won't allow you to clone into a non-empty existing directory

.gitignore
ehelms: Can this then be moved elsewhere? I don't like ignoring a directory that deep in our roles as I've had bad experiences with it as a developer.

playbooks/katello_stable.yml
```diff
@@ -0,0 +1,4 @@
+# For roles to run after a stable Katello development image is provisioned
+- hosts: all
```
ehelms: How about something akin to `setup_user_home` or `setup_user_config` or `configure_user_home_directory` as the playbook name?

playbooks/katello_stable.yml
ehelms: Off the top of my head `setup_user_development_environment`

roles/customize_home/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+- name: Check for specified files to copy to home directory
+  local_action: stat
+                path={{ role_path }}/files/
```
ehelms: I tend to find this format more obvious:

```
local_action:
  module: stat
  path: {{ role_path }}/files
```

.gitignore
ehelms: Sorry, by elsewhere I mean a directory we ignore at the top level like `user_dotfiles` or `user_devel_environment_files`

docs/development.md
```diff
@@ -134,6 +134,37 @@ in your boxes.yaml entry to configure this on box creation.
 #### Additional info
 - `foreman start` can be run with `rails` or `webpack` to start just that server: i.e. `bundle exec foreman start rails`. This can be used to start servers separately.
 
+### Customizing the Development Environment
+
+#### Custom local files
+
+You can have files automatically copied over to the `vagrant` user's home directory when spinning up a `centos7-katello-devel` or `centos7-katello-devel-stable` box. To do so, follow these steps:
```
ehelms: Technically this is not tied to vagrant and should work in any development setup right?

roles/customize_home/tasks/clone_specified_repo.yml
```diff
@@ -0,0 +1,29 @@
+- name: Ensure temporary home directory does not exist
+  file:
+    path: "{{ ansible_env.HOME }}/tmp_custom_home/"
+    state: absent
+
+- name: Clone specified repo into temporary custom home directory
+  git:
+    repo: "{{ custom_home_git_repo }}"
```
ehelms: customize_home_git_repo

ehelms:
Any idea if this covers the use cases for most developers? If I consider my own setup, then this would have to be re-factored to a different design to support:

 * file list
 * commands to run
 * packages to install

However, there is a good use case for supporting a git repo based design as it encourages (and educates) developers not to build environments that are hard-coded but rather to build development configurations that can survive hardware issues or porting to multiple environments. 

My own workflow is (poorly named git repo though):

```
git clone https://github.com/ehelms/vim-settings
cd vim-settings
./bootstrap.sh
```

johnpmitsch:
> Any idea if this covers the use cases for most developers? If I consider my own setup, then this would have to be re-factored to a different design to support:
> 
> * file list
> * commands to run
> * packages to install

I see your point, and I've thought of adding a way to specify packages to install or having a scripts directory where any file in it is automatically executed. But these are also specific tasks and are assuming devs want these things. The nice thing about copying a whole git repository's contents to the home directory is that you can put whatever you want in it. It's trivial to make a script in the repo that installs the packages you want, calls 20 different commands, and makes your terminal prompt neon-pink.  You can then run that easily as its already copied to your dev box.

So I think starting with a broad solution like this is for the best. However, I'm not opposed to adding more specific tasks after this change, especially if many devs are doing the same tasks after spinning up a dev box.

> However, there is a good use case for supporting a git repo based design as it encourages (and educates) developers not to build environments that are hard-coded but rather to build development configurations that can survive hardware issues or porting to multiple environments.

Agreed! This encourages good practices around storing your configuration in a versioned repo.

> My own workflow is (poorly named git repo though):
> 
> ```
> git clone https://github.com/ehelms/vim-settings
> cd vim-settings
> ./bootstrap.sh
> ```

I had a similar setup in my git repo, it was pretty easy to refactor to put the dotfiles in the root level and modify the scripts a bit. I would rather have a standard paradigm that is easy to adapt to than adapt to everyone's existing workflows and introduce more complexity. But I'm open to suggestions

I'm don't think this is a complete solution, ideally we get to a point where every dev can run one command and have a dev box set up exactly how they like it. This is beneficial not only for ease of use and time saved, but it encourages devs to keep an up-to-date environment and makes the decision to spin up a new environment an easy one. I think we can iterate on improvements to get to this point, I can start a community discussion asking for more feedback after this initial change and see what sort of post-spin-up steps everyone takes.

johnpmitsch:
@ehelms Thinking about this more, if we set a `boostrap_script` variable that would execute a script copied from the git repo or local directory, that would make the setup even more automated and allow devs a lot of flexibility in what they did with that script.

ehelms:
@johnpmitsch Agreed! On the whole, I do see where you are going with this PR and I think the concept is great. The part I worry about most is if in creating a workflow, it makes it harder for developers later on given they may end up tightly coupled to their environments. Creating easy to use but prescriptive of best practices methods provides the mechanism for users while pushing them to a place that, while it may require a little work, in the end they are in a better place.

That is where I like the idea of developer creates a git repository with everything they need, and our tasks clone that repository. From reading your feedback, it does sound like we could support two default ideas:

 1) copy any dotfiles found in the git repository to $HOME
 2) if a `bootstrap.sh` script (or whatever we want to call it) is found, it is executed

There could be some secrets (e.g. SSH keys) on a user's box that cannot be stored in git easily that a user may want to transfer. That is a bit tougher, and perhaps we can be specific about those instances to help users out.

johnpmitsch:
@ehelms I updated to add a way to execute a bootstrap script. I still have to do the documentation and update the other parts we discussed.

It uses the `custom_bootstrap_script` variable which defaults to running `bootstrap` from the home directory (so this file would be in the root dir of your git repo or the local folder). I kept off the `.sh` to keep it language-agnostic. Users can specify their own path with that variable. It will just have to be structured like `myscriptfolder/myscript.py` (for example) and not have a leading `./`. I couldn't figure out how to concatenate two file paths in ansible so my current approach is to just make sure this is documented.

> There could be some secrets (e.g. SSH keys) on a user's box that cannot be stored in git easily that a user may want to transfer. That is a bit tougher, and perhaps we can be specific about those instances to help users out.

@jturel mentioned a way to set this up so it happens automatically. If you have the docs for this (iirc it was in github's docs), maybe we can link it in our docs in forklift as a recommendation!

johnpmitsch:
@ehelms updated!

jturel:
> @jturel mentioned a way to set this up so it happens automatically. If you have the docs for this (iirc it was in github's docs), maybe we can link it in our docs in forklift as a recommendation!
> 

Did you want to include that in this PR? Those docs are here https://developer.github.com/v3/guides/using-ssh-agent-forwarding/

johnpmitsch:
> Did you want to include that in this PR? Those docs are here https://developer.github.com/v3/guides/using-ssh-agent-forwarding/

@jturel thanks, I added it in!

docs/development.md
```diff
@@ -134,6 +141,48 @@ in your boxes.yaml entry to configure this on box creation.
...
+  hostname: centos7-katello-devel-stable.example.com
+  ansible:
+    playbook: 'playbooks/setup_user_devel_environment.yml'
+    variables:
+      customize_home_git_repo: https://github.com/johnpmitsch/config_settings
+```
+
+The contents of the specified repo will be copied to the home directory excluding the `.git` folder and `.gitignore` file.
+
+A file named `bootsrap` will be looked for in the home directory and executed if found. This can be used to run any commands you would like to during provisioning. Be sure to make the file executable and use the proper language [shebang](https://en.wikipedia.org/w/index.php?title=Shebang_(Unix)) at the top. A different script location can specified by using the `custom_bootstrap_script` variable. The `custom_bootstrap_script` variable needs to be specified in the format `path/to/my/script.sh` (don't prepend with `./`).
```
ehelms: s/bootsrap/bootstrap

docs/development.md
```diff
@@ -134,6 +141,48 @@ in your boxes.yaml entry to configure this on box creation.
...
+centos7-katello-devel-stable:
+  box_name: katello/katello-devel
+  hostname: centos7-katello-devel-stable.example.com
+  ansible:
+    playbook: 'playbooks/setup_user_devel_environment.yml'
+    variables:
+      customize_home_git_repo: https://github.com/johnpmitsch/config_settings
+```
+
+The contents of the specified repo will be copied to the home directory excluding the `.git` folder and `.gitignore` file.
```
ehelms: If I supply a bootstrap script, does this still perform the copy (I would assume it should not)? That should be made clear to the user.

playbooks/setup_user_devel_environment.yml
```diff
@@ -0,0 +1,4 @@
+# For roles to run after a stable Katello development image is provisioned
```
ehelms: Can we drop the comment given this playbook is generic to any environment?

roles/customize_home/defaults/main.yml
```diff
@@ -0,0 +1 @@
+custom_bootstrap_script: 'bootstrap'
```
ehelms: Should be `customize_home_custom_bootstrap_script` to follow namespacing conventions.

roles/customize_home/tasks/clone_specified_repo.yml
```diff
@@ -0,0 +1,39 @@
...
+    path: "{{ ansible_env.HOME }}/tmp_custom_home/"
+    state: absent
+
+- name: Check custom bootstrap script exists
+  stat:
+    path: "{{ ansible_env.HOME }}/{{ custom_bootstrap_script }}"
+  register: bootstrap_script_path
+
+# The script is already on the machine, so using 'command' instead of 'script', which copies the script from the Ansible controller
+- name: Run custom bootstrap script
```
ehelms: If I have a custom script, I expect that script to be run and control everything that happens.

.gitignore
```diff
@@ -15,6 +15,7 @@ user_playbooks
 playbooks/galaxy_roles
 requirements.yml
 inventories/local_*
+user_devel_env_files/*
```
ehelms: I saw this is included but with a .gitkeep... I think I am less inclined to cloud the top level _committed_ directories to this repository and don't consider it too much to ask a developer to simply create this folder. If we do keep it, maybe shorten it down to `user_env_files` given this is applicable for use on any machine and any setup.

ehelms: The example would be, I sometimes want to setup `root` home for some activities with some of my `vim` configuration.

docs/development.md
johnpmitsch: @ehelms the order of operations is:

- Git repo's contents is dumped in to the target user's home dir
- .git and .gitignore files are removed if present
- the specified bootstrap script script is run, or if not specified, a `bootstrap` script is run if found. If nothing is specified and not `bootstrap` script present, do nothing.

So if I understand the question correctly, the copy always happens regardless of whether the bootstrap script is present. Does that answer your question? I can clarify in the docs, but trying to understand the comment a little better.

roles/customize_home/tasks/clone_specified_repo.yml
johnpmitsch: @ehelms can you elaborate on this? The script is run on the machine after being copied over.

.gitignore
johnpmitsch: sure, we can keep it uncommitted and ask the user to create it

johnpmitsch:
> My concern with this is that Vagrant itself also provides a way to sync files. It can even use NFS so they're actually live. However, the whole topic of customizing a workspace is very personal. Not sure what an actual good practice is.

@ekohl I know as a developer, I like keeping my non-sensitive environment setup in a git repo. Anecdotally, I know others do the same. So having a way to copy over the files and run commands specified automatically from a git repo makes life way easier.

The goal is to have a one-command devel environment with a guaranteed successful installation that is customized the way a dev specifies. I detailed out some of the reasons for this in a previous comment, its not only for ease of use and time, but encouragement to keep a fresh, up-to-date devel environment around. Currently, I know some devs have 6+ month old dev environments they are using. The reasoning generally is a) I'm not sure if it will spin up successfully (which was the motivation for the stable box) and b) I have to set it up the way I like it again (hence this PR)

Just wanted to share some of the goals behind this PR and the other recent changes. I'm open to approaching things differently in Vagrant/Ansible, but I ask that we keep these goals in mind that are not only from me, but other devs :)

ekohl:
I can appreciate that. I'll also share my (long term) vision of how we should have this.

I'd like to provide multiple boxes. That means you can use a production box when you don't care about that project. This heavily relies on some actual changes to the projects. Then we can also make it easy to use a local git clone without a VM just for foreman (with plugins).

One example of how I'd like to get there is that for Pulp 3 is that there's no need for a shared anything. You can deploy a pulp.example.com box (either production style, kubernetes or dev checkout). Then you configure the foreman-proxy with the URLs pointing there.

Once we drop Pulp 2, I hope Katello no longer has the concept of an internal proxy which means we can also deploy a foreman-proxy.example.com.

Easy deployment of a candlepin.example.com box should be much easier once qpid is out of the picture.

All of this relies on an improved certificate setup as well.

Note that all of the above is also where I would like productions setups to (be able to) evolve to. It allows scaling out application servers. When services are separated and only loosely coupled, installing them is also easier. Perhaps that means we can look at alternative installation methods (ansible, containers, mixed).

Of course I realize we're far from that point so I'm ok with changes that might not be ideal. Just wanted to share my long term goal.

johnpmitsch:
> Of course I realize we're far from that point so I'm ok with changes that might not be ideal. Just wanted to share my long term goal.

@ekohl Thanks for sharing, its good to know where you want things to head. That sounds like a beneficial setup and I would be curious to learn more as development progresses on that. I think these solutions that benefit developers can evolve alongside these changes. There definitely will be a lot of potential with loosely coupled moving parts!

docs/development.md
ehelms: That does clarify and might cause errors. I suppose I can test. I sorta liked the idea of its either or but not both. My bootstrap script for example symlinks my dotfiles so when I git pull updates, they are automatically in place. I can understand if I am at a bit of an advanced use case point,

roles/customize_home/tasks/clone_specified_repo.yml
ehelms: It is related to my previous comment around order of operations. I tend to look at it as, if I provide a script, the script should be all that happens and thus I have complete control over everything. Whereas if no script is supplied, I expect the simple case of copying dotfiles.

docs/development.md
johnpmitsch: > That does clarify and might cause errors. I suppose I can test. I sorta liked the idea of its either or but not both. My bootstrap script for example symlinks my dotfiles so when I git pull updates, they are automatically in place. I can understand if I am at a bit of an advanced use case point,

That is one thing that did come to mind, I actually do the same (or did, before I updated for this PR). However, you could `vagrant provision` to apply the latest updates in the git repo - especially with the stable box, this will only run the customization steps since its a pre-installed image. Its a different workflow, but I think for the better than trying to manage symlinked dotfiles on a devel box.

We get some feedback from devs once this is merged and see if there are concerns or asks in this area. Its definitely more of a "convention over configuration" approach, but I think trying to adapt to everyone's workflows could be a slippery slope and add add too much complexity.

roles/customize_home/tasks/clone_specified_repo.yml
johnpmitsch: @ehelms hopefully I addressed this above, but if you still have concerns maybe we can poll some other devs for input.

johnpmitsch:
@ehelms I updated this and I tested out the successful runs of the following:

- Specify nothing: Box spins up with no custom config
- Just a `user_devel_env_files` directory:  files are copied over
- Just a github repo directory: files copied over and bootstrap script ran
- both local directory and github repo: files copied from both locations and bootstrap script ran

docs/development.md
johnpmitsch: @ehelms  Also, I want to add, the current procedure isn't exclusive to dotfiles, thats just the intended usage so it makes for a good example. It is just dumping a git repo into the home directory and running a script if its present, so you can make that work for you however you want. 

It seems like you are thinking of replicating the cloning of the repo and symlinking the dotfiles strategy. So the big difference is the copying vs. the cloning. I think the copying is simpler (especially since you can re-provision to updated) so it would be nice to start out that way

docs/development.md
```diff
@@ -134,6 +141,51 @@ in your boxes.yaml entry to configure this on box creation.
...
+
+A file named `bootsrap` will be looked for in the home directory and executed if found. This can be used to run any commands you would like to during provisioning. Be sure to make the file executable and use the proper language [shebang](https://en.wikipedia.org/w/index.php?title=Shebang_(Unix)) at the top. A different script location can specified by using the `customize_home_bootstrap_script` variable. The `customize_home_bootstrap_script` variable needs to be specified in the format `path/to/my/script.sh` (don't prepend with `./`).
```
ehelms: The word 'bootstrap' is mispelled in the first line.

johnpmitsch: @ehelms thanks, missed that one, updated!

Issue #1026, opened 2019-09-26
skip foreman client repos in foreman install pipeline


evgeni:
Additionally we should probably skip the client repos on Foreman as well since those are supposed to be optional but that's for another PR.

_Originally posted by @ekohl in https://github.com/theforeman/forklift/pull/968/files_

PR #1049, opened 2019-11-06
don't enable client repos on foreman pipelines

evgeni:
this is only needed for katello/luna

Fixes: #1026

### DIFF
--- a/pipelines/upgrade_pipeline.yml
+++ b/pipelines/upgrade_pipeline.yml
@@ -29,7 +29,8 @@
     - selinux
     - etc_hosts
     - foreman_server_repositories
-    - foreman_client_repositories
+    - role: foreman_client_repositories
+      when: pipeline_type != 'foreman'
     - update_os_packages
     - haveged
     - disable_firewall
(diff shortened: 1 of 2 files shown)


pipelines/upgrade_pipeline.yml
```diff
@@ -130,7 +132,8 @@
       scenario: "{{ pipeline_type }}"
       scenario_version: "{{ forklift_upgrade_version_intermediate }}"
     - foreman_server_repositories
-    - foreman_client_repositories
+    - role: foreman_client_repositories
+      when: pipeline_type != 'foreman'
```
ehelms: Would it make sense to move this down into the testing section given the client repositories are only used for that aspect? I agree that this check still needs to be here but it more localizes the use of these.

evgeni: Yeah, I thought about that and then put it off as "don't do two changes in one go", but if you also thought that, I'll go ahead and do that.

evgeni: done

PR #1056, opened 2019-11-12
Add configurable box type loading to speed up environment

ehelms:
The idea here is to define 'profiles' for the Vagrant environment in Forklift to only display the types of boxes a particular user or CI system may care about. This reduce the giant list of options when running `vagrant status` and speeds up any calls that use `vagrant status` such as Forklift's role.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ By default, the boxes are set with `example.com` domain.
 If you're using NetworkManager, [this advanced DNS configuration](https://m0dlx.com/blog/Automatic_DNS_updates_from_libvirt_guests.html)
 allows completely automated dns resolution using dnsmasq from host to guest and guest to guest.
 
-You can disable hostmanager in `settings.yaml` by setting `hostmanager_enabled` option.
+You can disable hostmanager in `vagrant/settings.yaml` by setting `hostmanager_enabled` option.
 
 ### Adding Custom Boxes
 
@@ -165,8 +165,8 @@ static:
     management_network_address: 172.23.99.0/24
 ```
 
-Example with openstack provider:  
-You will need to install vagrant openstack provider. For more information click [here](https://github.com/ggiamarchi/vagrant-openstack-provider).  
+Example with openstack provider:
+You will need to install vagrant openstack provider. For more information click [here](https://github.com/ggiamarchi/vagrant-openstack-provider).
 Do not forget to set openstack API credentials.
 To use openstack provider as default look [here](https://www.vagrantup.com/docs/providers/default.html).
 
@@ -234,7 +234,8 @@ with-nfs:
 
 ### Customize Deployment Settings
 
-Some settings can be customized for the entirety of the deployment, they are:
+
+Some settings can be customized for the entirety of the deployment by copying `vagrant/settings.yaml.example` to `vagrant/settings.yaml` and add, remove or updating:
 
  * memory: Memory to give boxes by default unless specified by a box
  * cpus: Number of CPUs to give boxes by default unless specified by a box
@@ -245,7 +246,18 @@ Some settings can be customized for the entirety of the deployment, they are:
  * domain: domain for your hosts, you can override this per-box by configuring your box with a domain directly
  * libvirt_options, virtualbox_options, rackspace_options, openstack_options, google_options: custom options for the various providers
 
-To customize any of these, copy `settings.yaml.example` to `settings.yaml` and add, remove or update the ones you wish to change'
+#### Customize Available Boxes
+
+The list of available boxes can be customized by setting an exclude list in `vagrant/settings.yaml`. This allows faster `vagrant status` calls as well as reducing the the scope of boxes a user sees to tailor to their use cases. To specify boxes to exclude add the following to `vagrant/settings.yaml`, for example, to remove fips, fedora and any Foreman 1.2X boxes from view:
+
+```
+boxes:
+  exclude:
+    - "katello" # exclude any box containing "katello"
+    - "ubuntu1604-foreman-1\\.24" # exclude only the box "ubuntu1604-foreman-1.24". Notice the escaped '.' character to match the specific character instead of any single character
+    - "^centos7-fips" # exclude any box that starts with "centos7-fips"
+    - "foreman-1\\.(?:[2][0-3])" # exclude any foreman-1.20 to foreman-1.23 version box
+```
 
 ### Post Install Playbooks
 
--- a/vagrant/lib/forklift/box_factory.rb
+++ b/vagrant/lib/forklift/box_factory.rb
@@ -5,6 +5,7 @@ require 'erb'
 require 'yaml'
 
 require_relative 'compat'
+require_relative 'settings'
 
 module Forklift
   class BoxFactory
@@ -37,6 +38,10 @@ module Forklift
 
     def process_boxes!(boxes)
       boxes.each do |name, box|
+        box_config = Settings.new.settings.key?('boxes') ? Settings.new.settings['boxes'] : {}
+
+        next if box_config.key?('exclude') && box_config['exclude'].any? { |exclude| name.match(/#{exclude}/) }
+
         box['name'] = name
         box = layer_base_box(box)
 
--- a/vagrant/Gemfile
+++ b/vagrant/Gemfile
@@ -7,6 +7,7 @@ group :test do
   gem 'mocha'
   gem 'rake'
   gem 'rubocop'
+  gem 'json'
 end
 
 # rubocop:enable Naming/FileName
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative 'settings'
+
 module Forklift
   class BoxDistributor
 
@@ -21,8 +23,6 @@ module Forklift
     end
 
     def settings
-      overrides = {}
-      settings_file = File.join(__dir__, '..', '..', 'settings.yaml')
       default_settings = {
         'memory' => 6144,
         'cpus' => 2,
@@ -36,7 +36,7 @@ module Forklift
         'hostmanager_enabled' => true
       }
 
-      overrides = YAML.load_file(settings_file) if File.exist?(settings_file)
+      overrides = Settings.new.settings
 
       default_settings.merge(overrides)
     end
--- a/vagrant/lib/forklift/settings.rb
+++ b/vagrant/lib/forklift/settings.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Forklift
+  class Settings
+
+    attr_reader :settings
+
+    def initialize
+      settings_file = File.join(__dir__, '..', '..', 'settings.yaml')
+      @settings = File.exist?(settings_file) ? YAML.load_file(settings_file) : {}
+    end
+
+  end
+end
--- a/vagrant/settings.yaml.example
+++ b/vagrant/settings.yaml.example
@@ -16,3 +16,7 @@ domain: 'example.com'
 google_project_id: "YOUR_GOOGLE_CLOUD_PROJECT_ID"
 google_client_email: "YOUR_SERVICE_ACCOUNT_EMAIL_ADDRESS"
 google_json_key_location: "/path/to/your/private-key.json"
+
+boxes:
+  exclude:
+    - fips


ehelms:
To enable a profile, either a user copies an existing profile into `vagrant/config/boxes.yaml` or creates a symlink.

ekohl:
I had a slightly different idea. Perhaps we should ship those boxes files in `boxes.available.d` and let users symlink to `boxes.enabled.d`. A small wrapper could be added as well. It would reduce the need for config duplication. Thoughts?

ehelms:
For me, the box files are only half the picture. They only define the base boxes, and thus while you may be able to cut out some of the OSes, you still get every version of Foreman, every FIPS combination, all the random infra or non-standard environment boxes that may mean nothing to you as a user.

ehelms:
I'll also see if various stakeholders have thoughts on the ideas: @evgeni @johnpmitsch @jlsherrill @akofink

johnpmitsch:
@ehelms How does one use this? Could you explain a bit more?

vagrant/config/boxes.all.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - fedora29
+  - fedora30
+
+foreman_versions:
+  - '1.20'
+  - '1.21'
+  - '1.22'
+  - '1.23'
+  - '1.24'
+  - 'nightly'
```
akofink: nit: these don't need to be quoted

evgeni: True, but 1.20, 1.30 etc needs to be, and I must admit I like consistency :)

vagrant/config/boxes.ci.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - ubuntu1604
+  - ubuntu1804
+
+foreman_versions:
+  - 1.20
+  - 1.21
+  - 1.22
+  - 1.23
+  - 1.24
+  - nightly
```
akofink: @evgeni so these don't work?

vagrant/config/boxes.all.yaml
akofink: They shouldn't need to be quoted either afaik

evgeni: 1.20 is parsed as a float and becomes 1.2 :(

vagrant/config/boxes.ci.yaml
evgeni: They will, as long as they can't be miss-parsed as a float.

akofink:
@ehelms you could make this even more configurable by offering a `include/exclude` option and removing the custom fips/devel options. Something like this:

```yaml
operating_systems:
  - centos7

foreman_versions:
  - 1.20
  - 1.21
  - 1.22
  - 1.23
  - 1.24
  - nightly

include:
  - devel
exclude:
  - nightly
  - fips
```

vagrant/config/boxes.all.yaml
akofink: Ouch, yeah that's a gotcha :)

vagrant/config/boxes.ci.yaml
akofink: @ehelms let's quote these for consistency

ehelms: Yea.. I did that for the all case but since this was a draft I got lazy elsewhere :)

ehelms:
@johnpmitsch The idea is that you can use one of the existing "profiles" or create your own as long as its symlinked to `boxes.yaml` or created by the user. The user can then specify certain options that reduce the amount of boxes seen in `vagrant status` call. This can reduce the time to run vagrant status from 15 seconds down to 1-2 depending on how limited you setup the environment. This means you can also scope what you want to see locally (e.g. hide all debian boxes, or hide old Foreman versions)

ehelms:
@akofink Is your idea with include/exclude that any "term" included in an exclude or include list would look for that term in the box name and include or exclude it?

akofink:
@ehelms yes, that's right. It looks like that's how the `fips` and `devel` options work as written, so why not generalize it?

vagrant/config/boxes.ci.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+fips: false
+devel: false
```
ekohl: devel is not a built in, unlike fips so IMHO doesn't belong here.

vagrant/config/boxes.ci.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+foreman_versions:
+  - 1.20
+  - 1.21
+  - 1.22
+  - 1.23
+  - 1.24
+  - nightly
+
+fips: false
```
ekohl: We could also list fips as `centos7-fips` in `operating_systems`

vagrant/config/boxes.ci.yaml
```diff
@@ -0,0 +1,17 @@
+operating_systems:
+  - centos7
+  - debian9
+  - debian10
+  - ubuntu1604
+  - ubuntu1804
+
+foreman_versions:
```
ekohl: In CI we always know which OS + version we're testing. CI also relies on the pipelines which AFAIK always defines its own boxes. So we can actually generate the `operating_systems` on the fly in Jenkins and leave out all `foreman_versions`.

johnpmitsch:
@ehelms I really like the idea and I like that we can improve the `vagrant status` load time. Works well in my testing. Here are a few thoughts and questions that come to mind:

- How will keeping these files up-to-date and pulling latest changes work with the user customization?
- To remove boxes, I have to either remove them or comment them out, if I want to add them back, I'll have to look up what the boxes are if I removed them. Maybe having a boolean flag or enabled/disabled list would help with that so the disabled boxes stay in the file and can be easily switched back to enabled
- I know you probably didn't add docs since this is a draft PR, but please do document this even if it seems intuitive to use.

I would aim to keep it something that a dev doesn't have to worry about if they don't want to, it seems to be this way currently, which is good. In general, +1 to the idea!

vagrant/lib/forklift/box_factory.rb
```diff
@@ -37,6 +38,10 @@ def load_box_file(file)
 
     def process_boxes!(boxes)
       boxes.each do |name, box|
+        next unless @box_config['operating_systems'].any? { |os| name.include?(os) }
+        next if !@box_config['fips'] && name.include?('fips')
+        next if !@box_config['devel'] && name.include?('devel')
```
ekohl: As said above: devel is not a built in box and is always in the local config. No sense in built in filtering.

vagrant/lib/forklift/box_factory.rb
```diff
@@ -37,6 +38,10 @@ def load_box_file(file)
 
     def process_boxes!(boxes)
       boxes.each do |name, box|
+        next unless @box_config['operating_systems'].any? { |os| name.include?(os) }
+        next if !@box_config['fips'] && name.include?('fips')
```
ekohl: `centos7-fips` is actually an OS and can be filtered out that way.

vagrant/lib/forklift/box_loader.rb
```diff
@@ -35,5 +35,15 @@ def versions
       YAML.load_file("#{@root_dir}/config/versions.yaml")
     end
 
+    def box_config
+      config = if File.exist?("#{@root_dir}/config/boxes.yaml")
+                 "#{@root_dir}/config/boxes.yaml"
+               else
+                 "#{@root_dir}/config/boxes.all.yaml"
```
ekohl: I'd like to avoid a `all` config because it's an additional thing to maintain. IMHO the code should handle when there's no config. The alternative is a special value for `operating_systems` and `foreman_versions` where instead of a list you can pass in the string `all`.

johnpmitsch:
@ehelms I haven't fully fleshed out this idea, but what do you think about using a file that will exclude boxes? This would be kept out of version control and configured by the user. If its not there, all the boxes will be shown. There also could be niceties like excluding `all` of a certain type of box or similar so its kept updated.

The advantage being its kept out of version control so won't need to be rebased and the "master list" of boxes somewhere else would stay updated.

vagrant/config/boxes.ci.yaml
```diff
@@ -0,0 +1,17 @@
+operating_systems:
```
evgeni: the CI doesn't use predefined boxes, only `versions.yaml` and then generates boxes based on that on the fly. so we don't need that file/profile at all (IMHO)

vagrant/lib/forklift/box_loader.rb
```diff
@@ -35,5 +35,15 @@ def versions
       YAML.load_file("#{@root_dir}/config/versions.yaml")
     end
 
+    def box_config
+      config = if File.exist?("#{@root_dir}/config/boxes.yaml")
```
evgeni: just so that I understand this PR correctly:
the idea is to provide filters (based on os, foreman version and certain properties of the box) ontop of what is generated by building a matrix from `versions.yaml`, correct?
and then we provide multiple files that could be used as "profiles" (well, presets for those filters)?

I think I would find it easier if those filters be just exclude ones the user can set in their `settings.yaml`, like (this is what I'd probably set it to):
```
exclude_os:
  - centos7-fips
  - debian9
  - ubuntu1604
  - ubuntu1804

exclude_foreman:
  - '1.20'
  - '1.21'
```

Maybe mixed with the idea from @ekohl to have certain configurations marked as "unsupported" and thus not displayed either unless the user sets `show_unsupported` to true -- this would eliminate the need for `exclude_foreman` *for me*.

ekohl:
Perhaps we need to reshape `settings.yaml` a bit. Currently it's describing a box default but not the global boxes.

ehelms:
The feedback I gather from folks they want is a simple place to indicate excludes to customize their environments. I can live with that. I will however explain a bit again my hypothesis just in case it sparks further thought.

I always feel like their are two sets of users: power users and developers. The power users are most of us here commenting on this PR, know the ins and outs of Forklift, use it either daily or it's used in daily workflows we care about. Creating an include/exclude list of things we are about makes sense. 

Then I see developers as those in the ecosystem who come to Forklift because they need to either replicate a failure on a released or nightly version, or they want a development environment. In which to me, they just want to spin up the environments they care about, have it be quick and just work.

With the "profiles" I was attempting to address this latter group by allowing both the full power of customization along with here are a few typical customization's to get someone started. 

There is probably conversation to be had around Forklift serving many masters and possibly simplifying that, I, however, don't want to cloud this PR with that discussion. We can have it on an issue if folks like.

evgeni:
But won't those non-power-users just use `boxes.all` and thus not benefit from any changes anyways?

johnpmitsch:
> I always feel like their are two sets of users: power users and developers. The power users are most of us here commenting on this PR, know the ins and outs of Forklift, use it either daily or it's used in daily workflows we care about. Creating an include/exclude list of things we are about makes sense.

My favorite tools to use are ones that are very easy to get up and running and require no configuration or decisions from me to do so, but they also allow me to configure things after the fact.

I think if I was a developer who wants to reproduce a bug, I want as few configuration steps to get a box up and running. Ideally, we keep it this way and don't add any more steps to the `Quickstart` flow. I really wouldn't care about what boxes show if I'm just getting started with forklift or use it occasionally.

However, for those of use who live in Forklift every day, its nice to have more advanced configuration, and we do care about the time it takes to show `vagrant status` and filtering boxes will be a really nice feature. So having a configuration that one can alter if they want to will be nice. 

So my  suggestion would be for an "out of sight, out of mind" solution until someone has a need for it. My thought was a `.boxesignore` file or similar solution. Just my two cents :)

akofink:
+1 to convention over configuration for new users and configuration for power users or automation environments. I think this PR adequately solves the problem and addresses that concern. Other solutions presented also seem fine but do not offer significant advantages over the current implementation IMO.

evgeni:
So comparing this PR to #1044, I prefer this, as it's more flexible and allows more granular selection what is shown (and what not).

As for tools that don't require configuration, I think both the proposed "exclude list" and this PR (which is mostly an "include list") would get to the same result if we define the default behaviour well enough (either "all", as this PR does, or "what most users care about" [which needs to be defined]).

I think the main question for me would be: who is to benefit here. If it's the average user who clones the repo just to spin up a box (to repro a bug or whatever), then the current "all" default yields zero benefit.

johnpmitsch:
If we are planning to use an include list, I am still wondering about this:

> How will keeping these files up-to-date and pulling latest changes work with the user customization?

Seems like an exclude/ignore list that can match patterns (for example, `debian*`) would allow the most flexibility and would stay up-to-date as boxes are added, but I have no idea if this is feasible or not.

ehelms:
This has been updated to a simple `exclude` only supported list specified in `settings.yaml`. That gives everyone the full power for now to ad as they wish.

vagrant/lib/forklift/box_factory.rb
```diff
@@ -37,6 +38,10 @@ def load_box_file(file)
 
     def process_boxes!(boxes)
       boxes.each do |name, box|
+        box_config = Settings.new.settings['boxes']
+
+        next if box_config.key?('exclude') && box_config['exclude'].any? { |exclude| name.include?(exclude) }
```
johnpmitsch: @ehelms how do you feel about allowing regex with `name.match(/#{exclude}/)`? That would allow for some more fine tuning over what is allowed, but still work for a simple include.

ehelms: Done.

README.md
```diff
@@ -247,6 +247,18 @@ Some settings can be customized for the entirety of the deployment, they are:
...
+
+The list of available boxes can be customized by setting an exclude list in `settings.yaml`. This allows faster `vagrant status` calls as well as reducing the the scope of boxes a user sees to tailor to their use cases. To specify boxes to exclude add the following to `settings.yaml`, for example, to remove fips, fedora and any Foreman 1.2X boxes from view:
+
+```
+boxes:
+  exclude:
+    - fips
+    - fedora
+    - '1.2'
+```
```
johnpmitsch: @ehelms should we mention or add an example with regex here now that its an option?

akofink: See, this is kind of an issue now IMO - this `'1.2'` exclude in the example is going to match a box I have named (i.e.) `centos7-152.example.com`. I don't think that's expected or desired, do you?

johnpmitsch: its still possible to match "1.2" with "1\\.2"
```
2.6.3 :017 > "centos7-152.example.com".match(/#{"1\\.2"}/)
 => nil 
2.6.3 :018 > "centos1.2.example.com".match(/#{"1\\.2"}/)
 => #<MatchData "1.2"> 
```
but yes, we should mention that the expressions are treated as ruby-flavor regex and add examples. 

I would prefer to keep regex to be able to fine tune the exact boxes that should be ignored. My thought is to be able to specify exactly what you want without side effects, such as a range of versions or any box that starts with "foo" but allow boxes that include "foo" in the middle of the name. This does come with handling special characters but I think it could be documented and examples can be added.

akofink: I'd lean towards the 'Keep it Simple' mantra. If this is a deal breaker for your use case, we could allow advanced users to specify a regex literal (`/^foo/`) if they want regex but keep the defaults easy to use and understand. After all, people tend to not read docs unless the system behaves in a non-intuitive way.

johnpmitsch: I think it just comes down to preference. A user would have to read the docs to even know about the feature so I think either will would work fine with good documentation. 

@ehelms any thoughts?

ehelms: If you can provide an example regex I can add that and it would help to settle the discussion.

johnpmitsch: @ehelms  Here is an example showing off simple to more specific matching:
```yaml
boxes:
  exclude:
    - "katello" # exclude any box containing "katello"
    - "ubuntu1604-foreman-1\\.24" # exclude only the box "ubuntu1604-foreman-1.24". Notice the escaped '.' character to match the specific character instead of any single character
    - "^centos7-fips" # exclude any box that starts with "centos7-fips"
    - "foreman-1\\.(?:[2][0-3])" # exclude any foreman-1.20 to foreman-1.23 version box
```

README.md
```diff
@@ -247,6 +247,18 @@ Some settings can be customized for the entirety of the deployment, they are:
 
 To customize any of these, copy `settings.yaml.example` to `settings.yaml` and add, remove or update the ones you wish to change'
 
+#### Customize Available Boxes
+
+The list of available boxes can be customized by setting an exclude list in `settings.yaml`. This allows faster `vagrant status` calls as well as reducing the the scope of boxes a user sees to tailor to their use cases. To specify boxes to exclude add the following to `settings.yaml`, for example, to remove fips, fedora and any Foreman 1.2X boxes from view:
```
johnpmitsch: Can we be specific and say "`vagrant/settings.yaml` found in forklift's root directory"?

johnpmitsch:
seems like there is an error when there is no `settings.yaml` file present:

```
[jomitsch@jomitsch forklift{pr1056}]$ vagrant status
Vagrant failed to initialize at a very early stage:

There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /home/jomitsch/sat-deploy/forklift/Vagrantfile
Line number: 43
Message: NoMethodError: undefined method `key?' for nil:NilClass
```

ehelms:
Updated docs and fixed up the code to not fail if no settings

README.md
ehelms: Let's roll with regex for now since you can do simple and complex and see where it takes us.

PR #1058, opened 2019-11-18
Fix os register condition

xprazak2:


xprazak2:
Correct, this is needed for the new Ansible version, it should work on older versions as well.

PR #1067, opened 2019-12-06
added pulpcore repository to staging_repos.yml

wbclark:
our goal is to install pulpcore from RPMs not from PyPI so it is necessary to have this repo available on our boxes

### DIFF
--- a/roles/katello_repositories/tasks/staging_repos.yml
+++ b/roles/katello_repositories/tasks/staging_repos.yml
@@ -40,6 +40,16 @@
   when:
     - katello_repositories_pulp_version == "nightly"
 
+- name: 'Add Pulpcore repository'
+  yum_repository:
+    name: pulpcore-repository
+    description: Pulpcore repository for Katello {{ katello_repositories_version }}
+    baseurl: "https://koji.katello.org/releases/yum/katello-{{ katello_repositories_version }}/pulpcore/el{{ ansible_distribution_major_version }}/x86_64/"
+    gpgcheck: no
+    enable: yes
+  when:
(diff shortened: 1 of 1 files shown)


roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -40,6 +40,16 @@
   when:
     - katello_repositories_pulp_version == "nightly"
 
+- name: 'Add Pulpcore repository'
+  yum_repository:
+    name: pulpcore-repository
+    description: Pulpcore nightly repository
+    baseurl: "https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el{{ ansible_distribution_major_version }}/x86_64/"
```
ekohl: This is the release repo, not staging. Take a look at the above koji repos.

wbclark: @ekohl got it, thanks

roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -40,6 +40,16 @@
   when:
     - katello_repositories_pulp_version == "nightly"
 
+- name: 'Add Pulpcore repository'
+  yum_repository:
+    name: pulpcore-repository
+    description: Pulpcore nightly repository
+    baseurl: "https://koji.katello.org/releases/yum/katello-nightly/pulpcore/el{{ ansible_distribution_major_version }}/x86_64/"
```
ekohl: nightly should be a parameter is a parameter. Otherwise you can't use the 3.15 repos once they show up.

wbclark: @ekohl I'm now using the parameter in the baseurl and description

PR #1076, opened 2019-12-19
add centos8 box

evgeni:


vagrant/boxes.d/00-centos.yaml
```diff
@@ -30,3 +30,7 @@ boxes:
     scenarios:
       - foreman
       - katello
+
+  centos8:
```
ekohl: Odd indenting?

evgeni: what's odd here?

ekohl: Looks like Github renders it incorrectly in diff mode and that was incorrect.

PR #1080, opened 2020-01-02
test docker content with podman

jlsherrill:


jlsherrill:
@ekohl as requested!

bats/fb-content-katello.bats
```diff
@@ -266,3 +266,10 @@ setup() {
   skip # because of above
   puppet agent --test && grep -q Lorem /tmp/dummy
 }
+
+@test "try fetching docker content" {
+  yum install -y podman
```
ekohl: Ideally the helper would be used

```suggestion
  tPackageInstall podman
```

bats/fb-content-katello.bats
```diff
@@ -266,3 +266,10 @@ setup() {
...
+  podman login $(hostname -f) -u admin -p changeme
```
ekohl: Not blocking for this PR, but there's a lot of `hostname -f` calls in this file. Perhaps the `setUp` method should be expanded to use a variable similar to how the DNS tests run (not sure if it's actually needed to export):
https://github.com/theforeman/forklift/blob/4bf20dfc93241ec03f151138c920fbc6781f55ed/bats/fb-proxy-dns.bats#L73-L83

jlsherrill: went ahead and did this

PR #1081, opened 2020-01-03
Add a squid box

ekohl:
The recommended approach with this is to create an isolated network with only IPv6 ULA.

```xml
<network ipv6="yes">
  <name>ipv6-isolated</name>
  <uuid>d3c8973f-b0e6-42a4-a95d-46657fb9bc8c</uuid>
  <bridge name="virbr3" stp="on" delay="0"/>
  <mac address="52:54:00:18:fc:77"/>
  <ip family="ipv6" address="fd00::" prefix="64">
  </ip>
</network>
```

The result is that the Squid box (per example) is in both the management network with IPv4 connectivity and in the ipv6-isolated network with the fd00::/64 ULA network.

I did notice the network sometimes disappeared and I had to manually change the BOOTPROTO from dhcp to none in /etc/sysconfig/network-scripts/ifcfg-eth1. Without that, NetworkManager
took the interface offline when it gave up when IPv4 DHCP failed.

Now a box can be provisioned with only a nic in ipv6-isolated. Since squid out of the box allows ULA networks, it will work out of the box.  On the isolated box, the proxy can be set as follows:

```bash
export http_proxy="http://[fd00::5054:ff:fe72:1e34]:3128"
export https_proxy="http://[fd00::5054:ff:fe72:1e34]:3128"
```

Modify the IP as needed.

### DIFF
--- a/vagrant/boxes.d/99-local.yaml.example
+++ b/vagrant/boxes.d/99-local.yaml.example
@@ -81,3 +81,18 @@ centos7-katello-devel-stable:
   hostname: centos7-katello-devel-stable.example.com
   ansible:
     playbook: 'playbooks/setup_user_devel_environment.yml'
+
+# This box's intended use is dual homed where one interface is in the
+# management network and the other in an IPv6-only network. If that's using an
+# ULA range or link local, it should have permission to request resources by
+# default. This depends on a manually defined network with a bridge named
+# virbr3.
+centos7-squid:
+  box: centos7
+  ansible:
+    playbook: 'playbooks/squid.yml'
+  networks:
+    - type: 'public_network'
+      options:
+        dev: 'virbr3'
+        type: 'bridge'
--- a/playbooks/squid.yml
+++ b/playbooks/squid.yml
@@ -1,0 +1,5 @@
+---
+- hosts: all
+  become: yes
+  roles:
+    - squid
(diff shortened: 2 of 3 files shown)


evgeni:
can such a network be created by vagrant-libvirt directly? maybe with something like

https://github.com/theforeman/forklift/blob/be56c7cf1bdaafa1e42e2b4d669e02988de9535c/vagrant/boxes.d/02-kubevirt.yaml#L8-L11

ekohl:
The problem is that vagrant-libvirt can't do IPv6-only now. https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1021 does exist but it hasn't seen a lot of activity and I don't think it's really ready for merge.

evgeni:
Well, you clearly did your homework. Go on then ;)

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -81,3 +81,13 @@ centos7-katello-devel-stable:
   hostname: centos7-katello-devel-stable.example.com
   ansible:
     playbook: 'playbooks/setup_user_devel_environment.yml'
+
+squid:
```
evgeni:
```suggestion
centos7-squid:
```

ekohl:
Updated. Also added a comment above the box.

PR #1084, opened 2020-01-10
Hardcode the smoker user's homedir

ekohl:
The smoker directory is in ~/smoker for non-root, but collection runs as root. In our infrastructure it's safe to assume that user is vagrant.


zjhuntin:
ACK

PR #1085, opened 2020-01-10
Use the correct destination directory

ekohl:
623ef516264aa07024616599bac2b944cea89c33 refactored the remote directory to be common, but this was missed. Since no results were copied before, this didn't show up. After e29944991e26182ea282c73991396f93437a51ab was merged, results were copied and it started to fail.


pcreech:
ack

PR #1091, opened 2020-01-13
Skip docker content on foreman 1.18

pcreech:


bats/fb-content-katello.bats
```diff
@@ -273,6 +273,9 @@ setup() {
   if [[ $(echo "${FOREMAN_VERSION}\n1.20" | sort --version-sort | tail -n 1) == "1.20" ]] ; then
```
ekohl: Isn't this also sufficient?
```suggestion
  if [[ $(echo "${FOREMAN_VERSION}\n1.20" | sort --version-sort | tail -n 1) != $FOREMAN_VERSION ]] ; then
```

PR #1092, opened 2020-01-13
use printf instead of echo

pcreech:
echo doesn't handle newlines


evgeni:
> That bash code worked for me locally.

Yeah, so did for me. But in some environments you need `echo -e`, which `printf` is a good replacement for :)

PR #1095, opened 2020-01-14
Add Foreman smoker box

johnpmitsch:
Adds the ability to spin up a box with pytest and smoker and run tests on a foreman/katello instance.

UPDATE: I added tasks to spit out the command and also alias it, so you can ignore the following :)

This currently fails when smoker is run, but smoker will be set up on the box and ready to run when the foreman/katello URL resolves properly.

I think in this case it is actually best to not to run smoker and rather just set it up.

The goal of this box is to have a way for developers to run the tests on a system locally. So I'm leaning towards either printing out the command needed to run or setting up an alias on the box for the expected command. This allows developers to ssh into the box and iterate on smoker tests and/or run them locally if trying to debug a failure in infrastructure.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -16,6 +16,7 @@ This covers how to setup and configure a development environment using the Forkl
  * [Capsule Development](#capsule-development)
  * [Client Development](#client-development)
  * [Dynflow](#dynflow-development)
+ * [Smoker](#smoker)
 
 ## Development Environment Deployment
 
@@ -379,3 +380,15 @@ vagrant up centos7-dynflow-devel
 ```
 
 In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
+
+## Smoker
+
+The testing tool [smoker](https://github.com/theforeman/smoker) can be set up with the `centos7-foreman-smoker` box and tests can be run against a separate Foreman/Katello instance.
+
+To use:
+1. Ensure that you have a running instance of Foreman/Katello.  
+2. Follow the example box definition in `vagrant/boxes.d/99-local.yaml.example` for `centos7-foreman-smoker` and update the `smoker_base_url` variable. With `pytest_run_tests` set to false, smoker tests will not be run by the playbook, but the box will be set up with pytest and the smoker repository will be cloned to the `vagrant` user's home directory.
+3. Run `vagrant up centos7-foreman-smoker`. A debug message will print showing the command to run smoker tests and the alias that has been set up. The alias is defined in `~/.bash_profile` on the box itself.
+4. You can then ssh into the smoker box. Ensure the hostname of the Foreman/Katello instance can be reached by the smoker box.
+5. From the smoker box, run tests as the vagrant user using the alias or running pytest commands manually. To change the testing options, please see [the smoker documentation](https://github.com/theforeman/smoker) and modify the alias or manually run pytest commands as necessary.
+
--- a/playbooks/smoker.yml
+++ b/playbooks/smoker.yml
@@ -1,0 +1,7 @@
+- hosts: all
+  roles:
+    - role: epel_repositories
+      become: true
+    - role: etc_hosts
+      become: true
+    - smoker
--- a/roles/pytest_project/defaults/main.yml
+++ b/roles/pytest_project/defaults/main.yml
@@ -19,3 +19,5 @@ pytest_project_junit_output: junit.xml
 pytest_project_markers:
 pytest_project_command_args:
 pytest_project_ignore_errors: no
+pytest_run_tests: true
+pytest_project_alias: ""
--- a/roles/pytest_project/tasks/local_env.yml
+++ b/roles/pytest_project/tasks/local_env.yml
@@ -1,0 +1,14 @@
+- name: "Create alias for testing command"
+  lineinfile: 
+    dest: ~/.bash_profile 
+    line: "alias {{ pytest_project_alias }}=\"{{ pytest_project_command }}\""
+  when: pytest_project_alias != ""
+
+- debug:
+    msg: |
+      pytest is installed and testing can be run with
+      {{ pytest_project_command }}
+
+- debug:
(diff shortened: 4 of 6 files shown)


johnpmitsch:
To review, please follow the instructions in the README, as those are the instructions a developer would use to use the smoker box, and point at a foreman/katello instance. 

One test will currently fail, but the tests themselves should execute successfully

docs/development.md
```diff
@@ -379,3 +380,15 @@ vagrant up centos7-dynflow-devel
...
 
 In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
+
+## Smoker
+
+The testing tool [smoker](https://github.com/theforeman/smoker) can be set up with the `centos7-foreman-smoker` box and tests can be run against a separate Foreman/Katello instance.
+
+To use:
+1. Follow the example box definition in `vagrant/boxes.d/99-local.yaml.example` and update the `smoker_base_url` variable. 
+2. Run `vagrant up centos7-foreman-smoker`. With `pytest_local_env` set to true, the box will be set up with pytest and the smoker repository will be cloned to the `vagrant` user's home directory, but the tests will not be run by the playbook.
```
jturel: The notes about pytest_local_env should be part of step 1 because you'll want to be aware of and potentially modify it before running `vagrant up`

docs/development.md
```diff
@@ -379,3 +380,15 @@ vagrant up centos7-dynflow-devel
...
+3. A debug message will print showing the command to run smoker tests and the alias that has been set up. The alias is defined in `~/.bash_profile` on the box itself.
+4. Ensure that you have a running instance of Foreman/Katello and the hostname of this instance can be reached by the smoker box.
```
jturel: I think this should be moved possibly to (1) or to paragraph under the Smoker heading

johnpmitsch:
@jturel thanks for the review, I updated the docs with your suggestions

@ekohl mind taking another look as well?

roles/pytest_project/tasks/local_env.yml
```diff
@@ -0,0 +1,11 @@
+- name: 'Alias rake'
```
ekohl: rake sounds incorrect

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -96,3 +96,13 @@ centos7-squid:
...
+
+centos7-foreman-smoker:
+  box: centos7
+  ansible:
+    playbook: 'playbooks/smoker.yml'
+    variables:
+      smoker_base_url: "CHANGEME" # For example: "https://centos7-katello-devel-stable.example.com"
+      pytest_project_alias: "foreman_smoker"
+      pytest_local_env: true
+
```
ekohl: Odd empty line?

roles/pytest_project/tasks/run.yml
```diff
@@ -13,3 +13,8 @@
   ignore_errors: "{{ pytest_project_ignore_errors }}"
   args:
     chdir: "{{ pytest_project_directory }}"
+  when: not pytest_local_env
+
+- include_tasks: local_env.yml
+  when: pytest_local_env
+
```
ekohl: Odd empty line?

roles/pytest_project/tasks/local_env.yml
```diff
@@ -0,0 +1,11 @@
...
+  lineinfile: 
+    dest: ~/.bash_profile 
+    line: "alias {{ pytest_project_alias }}=\"{{ pytest_project_command }}\""
+
+- debug:
+    msg: |
+      pytest is installed and testing can be run with
+      {{ pytest_project_command }}
+      which is aliased as the command {{ pytest_project_alias }}
+
```
ekohl: Odd empty line?

playbooks/smoker.yml
```diff
@@ -0,0 +1,8 @@
+- hosts: all
+  roles:
+    - role: epel_repositories
+      become: true
+    - role: etc_hosts
+      become: true
+    - smoker
+
```
ekohl: Odd empty line?

roles/pytest_project/defaults/main.yml
```diff
@@ -19,3 +19,5 @@ pytest_project_junit_output: junit.xml
 pytest_project_markers:
 pytest_project_command_args:
 pytest_project_ignore_errors: no
+pytest_local_env: false
```
ekohl: I think the name is not very descriptive. Would it make more sense to add a more explicit `pytest_run_tests: true` instead?

roles/pytest_project/defaults/main.yml
```diff
@@ -19,3 +19,5 @@ pytest_project_junit_output: junit.xml
...
+pytest_project_alias: "pytest-project"
```
ekohl: Perhaps this should support an empty string to skip setting up the alias. Perhaps that should even be the default

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -96,3 +96,13 @@ centos7-squid:
       options:
         dev: 'virbr3'
         type: 'bridge'
+
+centos7-foreman-smoker:
```
ekohl: Rather than a separate box, does it make sense to include it by default in the devel box? Technically there's no need for a separate box other than guaranteeing that there's separation between services. We capture that in our pipelines and can cover it in reviews.

johnpmitsch: I would rather keep a separate box for a few reasons:
1) If the steps in these roles are failing for any reason that will mean the devel box will fail to spin up
2) Being able to spin up different boxes and run smoker tests on them locally. I could see that being helpful for multiple development environments, foreman/katello/luna environments, and releases.
3) We can stay flexible as we figure out what the expectations are of the environment(s) that smoker will run on.

johnpmitsch:
@ekohl updated!

ekohl:
Thanks!

PR #1102, opened 2020-02-05
Make sure ror_52 is not present in bats tests

zjhuntin:


zjhuntin:
Related to https://github.com/theforeman/foreman-packaging/pull/4652

evgeni:
not merging, as the packaging PR is still open

PR #1104, opened 2020-02-07
Add rh-postgresql12 role and to katello_devel

ehelms:


ansible.cfg
```diff
@@ -2,7 +2,6 @@
 callback_whitelist = profile_tasks,profile_roles
 stdout_callback = yaml
 display_skipped_hosts = false
-verbosity = 1
```
rahulbajaj0509: not sure why we need to remove this one though.

roles/postgresql_scl/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+- name: 'Install SCL repository'
```
rahulbajaj0509: should we add a `---` in the beginning  of the playbook?

rahulbajaj0509:
@ehelms in the above PR, we are installing the packages, do we need to also enable their services? so that next time if system reboots or starts then the service would be on.

Let me know your thoughts.

ehelms:
The installer should handle the actual PG service. This is just to configure the environment to be int he context of the SCL. This is updated

PR #1107, opened 2020-02-13
Add new version of Foreman & Katello to version.yaml

zjhuntin:


vagrant/config/versions.yaml
```diff
@@ -55,6 +55,18 @@ installers:
...
       - 'ubuntu1804'
 
+  - foreman: '2.0'
+    katello: '3.15'
+    pulp: '2.21'
+    puppet: 6
+    boxes:
+      - 'centos7'
+      - 'centos7-fips'
+      - 'debian9'
```
ekohl: This should be dropped since it's not supported. Looks like this should also be reflected in nightly

vagrant/config/versions.yaml
```diff
@@ -55,6 +55,18 @@ installers:
...
+      - 'debian10'
+      - 'ubuntu1604'
```
ekohl: Same here

zjhuntin:
Bonus points!

PR #1115, opened 2020-02-24
Tune the Puppetserver in the devel scenario

ekohl:
This tunes the Puppetserver for a minimal deployment where very few to no clients are expected. This keeps the memory usage of the devel box a bit in check. By doing this via the default answers file, it shouldn't affect users who've already specified other options on the command line.


ekohl:
Note I haven't tested this since I don't have a devel box. @johnpmitsch @jturel mind taking a look?

ekohl:
@johnpmitsch do you have any numbers on memory usage before and after?

johnpmitsch:
@ekohl This is what I am seeing for the box with nothing running on it:
```
[vagrant@centos7-katello-devel ~]$ free -h
              total        used        free      shared  buff/cache   available
Mem:           8.4G        3.8G        908M         98M        3.7G        3.6G
Swap:          2.0G        520K        2.0G
```

PR #1116, opened 2020-02-27
Drop upgrade flag for Foreman 2.1+

ehelms:
Results from the change at https://github.com/theforeman/foreman-installer/pull/470

### DIFF
--- a/roles/foreman_installer/tasks/install.yml
+++ b/roles/foreman_installer/tasks/install.yml
@@ -4,14 +4,6 @@
     msg: 'foreman_installer_options must be an array'
   when: foreman_installer_options is string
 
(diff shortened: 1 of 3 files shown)


roles/foreman_installer/tasks/install.yml
```diff
@@ -4,11 +4,6 @@
     msg: 'foreman_installer_options must be an array'
   when: foreman_installer_options is string
 
-- name: 'Read installer version'
-  slurp:
-    src: /usr/share/foreman-installer/VERSION
-  register: foreman_installer_version_file
-
 - debug:
```
ekohl: Perhaps this debug should move along with the reading?

PR #1117, opened 2020-02-28
Add playbook to test puppet-pulpcore

wbclark:
This playbook uses puppet-pulpcore to deploy a standalone pulp3 server, which is useful both for testing changes to puppet-pulpcore or if you need a standalone pulp3 environment to play with.

### DIFF
--- a/playbooks/puppet-pulpcore.yml
+++ b/playbooks/puppet-pulpcore.yml
@@ -1,0 +1,96 @@
+# This playbook uses https://github.com/theforeman/puppet-pulpcore to deploy a standalone pulp3 server.
+# It is useful both for testing changes to puppet-pulpcore as well as for setting up a pulp3 playground environment.
+#
+# You can edit the task "Clone puppet-pulpcore" to point to a different remote or branch as necessary.
+#
+# You can edit the task "Deploy test.pp" if you want to test the module using some different params.
+#
+# To use the playbook, first create a box, e.g. centos7: `$ vagrant up centos7`
+# Then run the playbook on the box you created: `$ ansible-playbook -l centos7 playbooks/puppet-pulpcore.yml`
+---
+- hosts: all
+  become: true
+  roles:
+    - puppet_repositories
+    - epel_repositories
+
+  tasks:
+    - name: "Enable pulpcore repository"
+      yum_repository:
+        name: pulpcore-nightly
+        description: Pulpcore nightly YUM repo
+        baseurl: https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el7/x86_64/
+        gpgcheck: no
+        enabled: yes
+
+- hosts: all
+  tasks:
+
+    - name: 'Install packages'
+      yum:
+        name:
+          - ruby-devel
+          - rubygem-bundler
+          - libxml2-devel
+          - ruby
+          - rake
+          - gcc
+          - wget
+          - git
+          - nss
+          - puppet
+          - glib2
+          - centos-release-scl-rh
+        state: "present"
+      become: true
+
+    - name: "Install librarian-puppet"
+      gem:
+        name: librarian-puppet
+        state: present
+
+    - name: "Clone puppet-pulpcore"
+      git:
+        repo: https://github.com/theforeman/puppet-pulpcore.git
+        dest: "/home/vagrant/puppet-pulpcore"
+        version: master
+
+    - name: "Deploy Puppetfile"
+      copy:
+        content: "forge \"https://forgeapi.puppetlabs.com\"\nmetadata"
+        dest: "/home/vagrant/puppet-pulpcore/Puppetfile"
+
+    - name: "Install Puppet modules"
+      shell: "/home/vagrant/bin/librarian-puppet install"
+      args:
+        chdir: "/home/vagrant/puppet-pulpcore"
+      remote_user: "vagrant"
+
+    - name: "Link pulpcore puppet module"
+      file:
+        state: link
+        src: "/home/vagrant/puppet-pulpcore"
+        dest: "/home/vagrant/puppet-pulpcore/modules/pulpcore"
+
+    - name: "Deploy test.pp"
+      copy:
+        content: |
+          class { 'postgresql::globals':
+            version              => '10',
+            client_package_name  => 'rh-postgresql10-postgresql-syspaths',
+            server_package_name  => 'rh-postgresql10-postgresql-server-syspaths',
+            contrib_package_name => 'rh-postgresql10-postgresql-contrib-syspaths',
+            service_name         => 'postgresql',
+            datadir              => '/var/lib/pgsql/data',
+            confdir              => '/var/lib/pgsql/data',
+            bindir               => '/usr/bin',
+          }
+          class { pulpcore:
+            postgresql_manage_db => false,
+            postgresql_db_ssl    => true,
+          }
+        dest: "/home/vagrant/test.pp"
+
+    - name: "Run puppet apply"
+      become: true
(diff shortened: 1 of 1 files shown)


sthirugn:
> $ vagrant up centos7
> $ ansible-playbook -l centos7 playbooks/puppet-pulpcore

I think it would help to add the above instructions as comments in the playbook itself.

wbclark:
> > $ vagrant up centos7
> > $ ansible-playbook -l centos7 playbooks/puppet-pulpcore
> 
> I think it would help to add the above instructions as comments in the playbook itself.

That's a good idea. I will add it there and maybe shorten the commit message a bit.

playbooks/puppet-pulpcore.yml
```diff
@@ -0,0 +1,96 @@
...
+  roles:
+    - puppet_repositories
+    - epel_repositories
+
+  tasks:
+    - name: "Enable pulpcore repository"
+      yum_repository:
+        name: pulpcore-nightly
+        description: Pulpcore nightly YUM repo
+        baseurl: https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el7/x86_64/
```
ehelms: This limits this to EL7, with EL8 on the horizon you can make this pull the dist from Ansible facts to work on both.

playbooks/puppet-pulpcore.yml
```diff
@@ -0,0 +1,96 @@
...
+#
+# To use the playbook, first create a box, e.g. centos7: `$ vagrant up centos7`
+# Then run the playbook on the box you created: `$ ansible-playbook -l centos7 playbooks/puppet-pulpcore.yml`
+---
+- hosts: all
+  become: true
+  roles:
+    - puppet_repositories
+    - epel_repositories
+
```
ehelms: This line could be dropped here

playbooks/puppet-pulpcore.yml
```diff
@@ -0,0 +1,96 @@
...
+      yum_repository:
+        name: pulpcore-nightly
+        description: Pulpcore nightly YUM repo
+        baseurl: https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el7/x86_64/
+        gpgcheck: no
+        enabled: yes
+
+- hosts: all
+  tasks:
+
```
ehelms: This line could be dropped

playbooks/puppet-pulpcore.yml
```diff
@@ -0,0 +1,96 @@
...
+      file:
+        state: link
+        src: "/home/vagrant/puppet-pulpcore"
+        dest: "/home/vagrant/puppet-pulpcore/modules/pulpcore"
+
+    - name: "Deploy test.pp"
+      copy:
+        content: |
+          class { 'postgresql::globals':
+            version              => '10',
```
ehelms: We deploy nightly with PG 12, could we here as well?

playbooks/puppet-pulpcore.yml
```diff
@@ -0,0 +1,96 @@
...
+            client_package_name  => 'rh-postgresql10-postgresql-syspaths',
+            server_package_name  => 'rh-postgresql10-postgresql-server-syspaths',
+            contrib_package_name => 'rh-postgresql10-postgresql-contrib-syspaths',
+            service_name         => 'postgresql',
+            datadir              => '/var/lib/pgsql/data',
+            confdir              => '/var/lib/pgsql/data',
+            bindir               => '/usr/bin',
+          }
+          class { pulpcore:
+            postgresql_manage_db => false,
```
ehelms: Where is the database coming from in this instance? Or are you experimenting with these options in your local testing?

playbooks/puppet-pulpcore.yml
```diff
@@ -0,0 +1,96 @@
...
+
+    - name: "Deploy Puppetfile"
+      copy:
+        content: "forge \"https://forgeapi.puppetlabs.com\"\nmetadata"
+        dest: "/home/vagrant/puppet-pulpcore/Puppetfile"
+
+    - name: "Install Puppet modules"
+      shell: "/home/vagrant/bin/librarian-puppet install"
+      args:
+        chdir: "/home/vagrant/puppet-pulpcore"
```
ehelms: This is a good candidate for a variable given it's used in multiple places.

playbooks/puppet-pulpcore.yml
```diff
@@ -0,0 +1,96 @@
...
+      remote_user: "vagrant"
```
ehelms: This and the home directory tie it to use with vagrant environments. Could we use root or generalize to allow others?

ehelms:
The general pattern here of cloning a puppet module and configure an environment to run puppet apply could be nice to extract out to a role to allow this pattern for any module.

ekohl:
After our discussion yesterday, is this still relevant?

For context, I was urging to use the acceptance tests we have in combination with the vagrant backend. The benefit is that any test you write to verify functionality becomes a regression test we also run in our CI.

sthirugn:
Two things:
- I didn't know that we can run acceptance tests locally. Please provide any doc links to do that when you get a chance.
- Even with acceptance test running locally, this kind of playbook/method suggested by eric here will still be beneficial for the early phases of development of the module like how we did for pulp3.

ekohl:
> I didn't know that we can run acceptance tests locally. Please provide any doc links to do that when you get a chance.

Our acceptance tests have some defaults:

https://github.com/theforeman/foreman-installer-modulesync/blob/79498a77e3244bd3c726a72bdc895ca078d2f2e7/moduleroot/spec/spec_helper_acceptance.rb.erb#L1-L6

Personally I use [beaker-vagrant](https://github.com/puppetlabs/beaker-vagrant) (note this is not the RH beaker project) with libvirt instead of docker, so in my `.bashrc` I have `export BEAKER_HYPERVISOR=vagrant_libvirt` but it's just a regular env var. Libvirt are actual VMs so there's better isolation. The docker containers are a bit of a hack since they need to be privileged to run systemd. You also get SELinux so a better representation of the actual deployment.

In the defaults you can see that it defaults to centos 7 which for now is a safe default. Then you run it via:
```shell
# Full suite
bundle exec rake beaker
# Single test
bundle exec rspec spec/acceptance/my_spec.rb
```

If you want to keep the machine after running it, you can use `BEAKER_destroy=no`. Note that a machine will be provisioned by default, even if it exists. There's `BEAKER_provision=no` for that.

So in short what I usually end up doing:

```shell
export BEAKER_HYPERVISOR=vagrant_libvirt
BEAKER_destroy=no bundle exec rspec spec/acceptance/basic_spec.rb
```

Then it fails. If I need to log into the machine, I can do that:

```shell
cd .vagrant/beaker_vagrant_files/centos7-64\{hostname\=centos7-64.example.com\}
vagrant ssh
```

Now I modify some files (using my editor with all tooling set up right) and rerun the test using the same machine:
```
BEAKER_provision=no BEAKER_destroy=no bundle exec rspec spec/acceptance/basic_spec.rb
```

Acceptance tests use [serverspec](https://serverspec.org/) to verify resources. Doing so actually creates a nice regression test suite.

We currently don't have any multi node tests, but that can be done and used to verify remote postgresql installs.

> Even with acceptance test running locally, this kind of playbook/method suggested by eric here will still be beneficial for the early phases of development of the module like how we did for pulp3.

I disagree because the acceptance test workflow is just more flexible. It's integrated into the regular tooling that's already in place. Building it in forklift is reinventing a workflow that already exists I would like to avoid wasting time on that.

Note we're currently using beaker. Puppetlabs is developing [litmus](https://github.com/puppetlabs/puppet_litmus) but I feel it's not ready yet. I'm following it and when the time is right, I'll make a push to switch over.

ehelms:
Closing in favor of using acceptance tests in the puppet modules

PR #1118, opened 2020-03-02
fix #1025: Stop installing katello on proxies

### LINKED ISSUE #1025: make install pipeline not use katello role on proxy
This is probably also incorrect and might pull in `katello` as a `foreman_installer_additional_packages` on the proxy.

Additionally we should probably skip the client repos on Foreman as well since those are supposed to be optional but that's for another PR.

_Originally posted by @ekohl in https://github.com/theforeman/forklift/pull/968/files_

ehelms:


pipelines/vars/katello_base.yml
```diff
@@ -1,5 +1,8 @@
 foreman_server_repositories_katello: true
 foreman_installer_disable_system_checks: true
+foreman_installer_scenario: katello
```
evgeni: we set both in katello_server.yml, so you don't even have to add it to base :)

pipelines/install_pipeline.yml
```diff
@@ -25,10 +25,11 @@
     - role: disable_firewall
     - role: foreman_server_repositories
     - role: update_os_packages
-    - role: foreman
```
ekohl: If you no longer include these, you can drop the `foreman_installer_skip_installer: true` part on line 15.

ekohl: In https://github.com/theforeman/forklift/blob/master/pipelines/upgrade_pipeline.yml we do include the foreman-installer role so we get the `yum install` part of both boxes in parallel. I'd like to have both pipelines follow the same pattern if possible

ehelms:
Addressed comments

ekohl:
I also removed a redundant var. @evgeni mind giving it another look?

pipelines/upgrade_pipeline.yml
```diff
@@ -25,15 +23,15 @@
...
       scenario_version: "{{ forklift_upgrade_version_start }}"
-    - umask
-    - selinux
-    - etc_hosts
-    - foreman_server_repositories
-    - update_os_packages
-    - haveged
-    - enable_ipv6
-    - disable_firewall
-    - foreman_installer
```
evgeni: wasn't the idea of @ekohl's comment to have `foreman_installer` in both pipelines, so the RPM step is done in parallel on both machines, instead of removing it in both and thus not doing parallel at all?

ekohl:
In https://github.com/theforeman/forklift/pull/1120 I've taken a similar approach, but taken a step further. This should also make it easier in the future to do foreman upgrade testing, which would have prevented issues in 2.0.0-rc1 and 2.0.0-rc2.

ekohl:
Looks like there is an issue with the system checks on the proxy box that's not passed correctly. Will look into it later.

evgeni:
this is obsolete now

PR #1123, opened 2020-03-09
Fix nonexistent deep_merge!

ehelms:


ekohl:
(just make rubocop happy)

vagrant/lib/forklift/box_factory.rb
```diff
@@ -45,8 +45,8 @@ def process_boxes!(boxes)
         box = layer_base_box(box)
         box['name'] = name
 
-        if @boxes[name]
-          @boxes[name].deep_merge!(box)
+        if @boxes.key?(name) # rubocop:disable Style/ConditionalAssignment
```
ekohl: I'm not sure rails can disable inline like this. IIRC you also need to enable it again.
Alternative:
```ruby
@boxes[name] = @boxes.key?(name) ? deep_merge(@boxes[name], box) : box
```

PR #1124, opened 2020-03-11
Only run --upgrade on proxy upgrade less than Foreman 2.1

ehelms:


roles/foreman_installer/tasks/install.yml
```diff
@@ -4,14 +4,17 @@
     msg: 'foreman_installer_options must be an array'
   when: foreman_installer_options is string
 
+- debug:
```
ekohl: Is this still needed?

roles/foreman_installer/tasks/main.yml
```diff
@@ -9,6 +9,9 @@
 - debug:
     msg: "Foreman installer version {{ foreman_installer_version_file['content'] | b64decode }}"
 
+- set_fact:
```
ekohl: Perhaps this should be done before the debug. The debug can then use this variable.

roles/foreman_proxy_content/tasks/upgrade.yml
```diff
@@ -1,16 +1,16 @@
 ---
 - include_tasks: certs_generate.yml
 
+- set_fact:
+    foreman_installer_options_internal_use_only:
```
ekohl: This just wipes the internal options. The upgrade pipeline also sets these so I'm not sure that's wise.

evgeni:
this slightly differs from the solution in #1143 ;)

PR #1127, opened 2020-03-12
Revert use of dig which is Ruby 2.3+ only

ehelms:
CI systems are CentOS 7 based and thus Ruby 2.0.

### DIFF
--- a/bin/ansible-vars
+++ b/bin/ansible-vars
@@ -11,4 +11,4 @@ loader.load!
 
 names = $*
 boxes = names.any? ? loader.boxes.select { |name, box| names.include?(name) } : loader.boxes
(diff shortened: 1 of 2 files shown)


ehelms:
A follow on would be to have Rubocop target Ruby 2.0 and/or find a way to have Travis test python and Ruby 2.0. Perhaps Github actions is better suited?

evgeni:
@ehelms yeah, sounds sane to me. will you make cards/issues for that?

ekohl:
> Perhaps Github actions is better suited?

AFAIK GHA is going to drop Python 2 soon.

evgeni:
No, they reverted that decision

ekohl:
Good to know

evgeni:
Source: https://github.com/actions/virtual-environments/issues/399#issuecomment-594091483

PR #1135, opened 2020-03-16
Skip katello specific upgrade code

ekohl:
Foreman has never depended on foreman-maintain and isn't even available on Debian. It also doesn't have --upgrade, --certs-update-all and --disable-system-checks parameters.

### DIFF
--- a/roles/foreman_installer/tasks/upgrade.yml
+++ b/roles/foreman_installer/tasks/upgrade.yml
@@ -1,6 +1,7 @@
 ---
 - name: 'Stop services'
   command: foreman-maintain service stop
+  when: foreman_installer_scenario != 'foreman'
 
 - name: 'Update packages'
   import_role:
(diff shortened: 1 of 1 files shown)


roles/foreman_installer/tasks/upgrade.yml
```diff
@@ -1,16 +1,19 @@
...
-  when: foreman_installer_version_file['content'] | b64decode is version('2.1', '<')
+  when:
+    - foreman_installer_version_file['content'] | b64decode is version('2.1', '<')
+    - foreman_installer_scenario != 'foreman'
 
 - name: 'Run installer upgrade'
   import_tasks: "install.yml"
   vars:
-    foreman_installer_disable_system_checks: true
+    foreman_installer_disable_system_checks: foreman_installer_scenario != 'foreman'
```
ehelms: Does this work outside of curly brackets?

ekohl: I'll admit I only tested this with a Foreman pipeline, but there it evaluated to `False` and got it to upgrade me to 2.1.0-develop (other than that the upgrade itself failed due to HTTP 503).

PR #1136, opened 2020-03-19
Fix katello upgrade parameter passing

ekohl:
35ef518717d6eea3c9d393c27efaa3a7fa63ba82 introduced this, but uses incorrect syntax. It was only tested on Foreman where it evaluates to false. That works there. In the katello case it's actually needed and that broke.


ekohl:
To be clear: I still have a pipeline running with this, but I believe it's past the point that previously failed.

PR #1137, opened 2020-03-25
Split install and upgrade playbooks into separate plays

ehelms:
This splits the install and upgrade pipelines into separate plays that can still be run in the traditional all-in-one method. The idea is to set the stage for two things:

 * installer PR testing where I will need the power to specify when to inject a scratch built installer at the right point in the upgrade pipeline
 * update our Jenkins pipelines to have a stage per playbook for cleaner output and identification of stage that failed


roles/forklift_versions/defaults/main.yml
```diff
@@ -1,2 +1,2 @@
-forklift_versions_file: "{{ playbook_dir | dirname }}/vagrant/config/versions.yaml"
+forklift_versions_file: "{{ lookup('env', 'PWD') }}/vagrant/config/versions.yaml"
```
ekohl: Why is this needed? In my experience this failed because the directory changed. Should we use `forklift_directory` here?

ehelms:
I had a similar thought of expanding some things out to common steps. That could either just be symlinks or import those playbooks. I've already used the ability to retry failed steps in my daily work and it was nice not to have to re-spin the entire thing or comment code out.

roles/forklift_versions/defaults/main.yml
ehelms: This existing code puts on a restriction that playbooks can only be a single directory deep. I seem to recall PWD causing some issue in the past but not what issue. So far my testing has been all successful including my testing inside `ci.centos.org`.

evgeni:
I like that, and I have no idea/opinion about the `pwd` thing.

PR #1138, opened 2020-03-26
Fix the katello devel pipeline

ekohl:
My pipeline is still running, but creating the PR already.

### DIFF
--- a/roles/customize_home/tasks/main.yml
+++ b/roles/customize_home/tasks/main.yml
@@ -3,7 +3,6 @@
   local_action:
     module: stat
     path: "{{ playbook_dir | dirname }}/user_devel_env_files/"
-  become: no
   register: custom_home
 
 - name: Copy files to home directory
--- a/pipelines/katello_devel_pipeline.yml
+++ b/pipelines/katello_devel_pipeline.yml
@@ -1,6 +1,5 @@
 - hosts: localhost
   vars_files:
-    - vars/katello_base.yml
     - vars/katello_devel.yml
   roles:
     - forklift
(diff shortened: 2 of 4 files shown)


ekohl:
@johnpmitsch I'm removing a lot of variables that I think are redundant. Would you mind verifying? The goal is to get https://ci.theforeman.org/job/katello-devel-forklift-test/ green

ekohl:
And for what it's worth, it works for me.

johnpmitsch:
> The goal is to get https://ci.theforeman.org/job/katello-devel-forklift-test/ green

This pipeline is a little less important with the stable devel box, not that it invalidates the PR or anything, its just worth noting. The primary use of this pipeline was to check if the devel box is spinning up successfully if you were to spin one up, but now devs can just use the stable box. Also, the creation of the stable box serves as sort of a pipeline itself. If a stable box vagrant image was created with packer, the traditional counterpart will spin up successfully as its using the same playbook. This means if you want to spin up centos7-katello-devel, you could check https://app.vagrantup.com/katello/boxes/katello-devel and see if a vagrant image was created recently. If so, it will very likely spin up fine. Though this is a bit less intuitive, and right now the vagrant image creation process is running on my workstation (see https://github.com/johnpmitsch/vagrant_cloud_scripts), so the logs aren't visible to anyone except me. But on the other hand, it makes me wonder if the de-facto Katello devel environment should be `centos7-katello-devel-stable`?

We can keep the job for now and I'll review the PR, I just wanted to share my thoughts as things have changed quite a bit since we've created the job and we may want to remove or change it.

roles/customize_home/tasks/main.yml
```diff
@@ -3,7 +3,6 @@
   local_action:
     module: stat
     path: "{{ playbook_dir | dirname }}/user_devel_env_files/"
-  become: no
```
johnpmitsch: why change this? https://github.com/theforeman/forklift/commit/308e4b91aa3ef650076db670b2f3582961d91e78

ekohl: it's redundant from a few lines above and ansible was not happy with that.

johnpmitsch: @ekohl Is the issue described and fixed in https://github.com/theforeman/forklift/commit/308e4b91aa3ef650076db670b2f3582961d91e78 no longer a concern?

ekohl: It is, but it wasn't working. 3557d6692858eac4d52273bb8d464880ecc1dc00 added a proper boolean but failed to remove the incorrect one.

ehelms:
Merging as this job is visible to everyone and can give us insight still. The use of the stable box and its existence in the ecosystem needs to be elevated. We are ending up in another split-community situation.

PR #1140, opened 2020-04-01
Do not use verbose foreman-installer logging by default

ehelms:
We should rely on the normal error reporting from the installer for debugging and improve it if it does not meet our needs.

### DIFF
--- a/roles/foreman_installer/defaults/main.yml
+++ b/roles/foreman_installer/defaults/main.yml
@@ -2,7 +2,7 @@
 foreman_installer_additional_packages: []
 foreman_installer_admin_password: changeme
 foreman_installer_skip_installer: False
-foreman_installer_verbose: True
+foreman_installer_verbose: False
 foreman_installer_scenario: foreman
 foreman_installer_scenario_flag: '{{ "--scenario" if foreman_installer_scenario != "" else "" }}'
 foreman_installer_upgrade: False


ekohl:
My only issue is that we currently only store sosreports and on Debian those don't contain `/var/log/foreman-installer` which makes debugging them nearly impossible. IMHO we should first solve that. It may be ok in the short term to manually patch sosreport:
```patch
--- /usr/share/sosreport/sos/plugins/foreman.py.old	2020-04-01 10:31:55.130506712 +0000
+++ /usr/share/sosreport/sos/plugins/foreman.py	2020-04-01 10:33:20.873832549 +0000
@@ -14,16 +14,16 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 
-from sos.plugins import Plugin, RedHatPlugin
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
-class Foreman(Plugin, RedHatPlugin):
+class Foreman(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     """Foreman/Satellite 6 systems management
     """
 
     plugin_name = 'foreman'
     profiles = ('sysmgmt',)
-    packages = ('foreman-debug')
+    packages = ('foreman-debug',)
 
     def setup(self):
         cmd = "foreman-debug"
```

PR #1141, opened 2020-04-07
load the installer version in (install|upgrade).yml

evgeni:
loading it in main.yml implied we always call main.yml, but this is not
true in the proxy upgrade scenario where the proxy role calls
install.yml only

### DIFF
--- a/roles/foreman_installer/tasks/upgrade.yml
+++ b/roles/foreman_installer/tasks/upgrade.yml
@@ -7,6 +7,8 @@
   import_role:
     name: update_os_packages
 
+- include_tasks: installer_version.yml
+
 - set_fact:
     foreman_installer_options_internal_use_only: "{{ [ '--upgrade', '--certs-update-all' ] + foreman_installer_options_internal_use_only }}"
   when:
--- a/roles/foreman_installer/tasks/install.yml
+++ b/roles/foreman_installer/tasks/install.yml
@@ -1,4 +1,6 @@
 ---
+- include_tasks: installer_version.yml
+
 - name: 'Check if foreman_installer_options is a string'
   fail:
     msg: 'foreman_installer_options must be an array'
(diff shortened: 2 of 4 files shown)


evgeni:
@ehelms this should fix https://ci.centos.org/job/foreman-katello-upgrade-nightly-test/786/console

```
TASK [foreman_installer : Replace certs-tar and --inital-admin parms with legacy options] ***
fatal: [pipeline-upgrade-katello-proxy-nightly-centos7]: FAILED! => 
  msg: |-
    The conditional check 'foreman_installer_version_file['content'] | b64decode is version('1.21', '<')' failed. The error was: error while evaluating conditional (foreman_installer_version_file['content'] | b64decode is version('1.21', '<')): 'foreman_installer_version_file' is undefined
  
    The error appears to be in '/root/forklift/roles/foreman_installer/tasks/install.yml': line 11, column 3, but may
  
    - name: 'Replace certs-tar and --inital-admin parms with legacy options'
      ^ here
```

roles/foreman_installer/tasks/upgrade.yml
```diff
@@ -7,6 +7,8 @@
   import_role:
     name: update_os_packages
 
+- include_tasks: installer_version.yml
```
evgeni, line 10: interestingly, I think the "add `--upgrade`" logic in `upgrade.yml` was broken before this, as the version was read before we upgraded the packages

evgeni: ah, it was not, as we upgrade the installer package before calling `upgrade.yml`, clever.

ekohl: I'd like to kill this task eventually. Does that mean that `update_os_packages` is redundant and can be dropped?

evgeni: No, it does not mean that. Unless you want a machine with upgraded installer, but not upgraded everything else ;)

PR #1143, opened 2020-04-07
let foreman_proxy_content role handle the whole upgrade 

evgeni:


pipelines/upgrade_pipeline.yml
```diff
@@ -4,7 +4,9 @@
 - import_playbook: upgrade/03-install_server.yaml
 - import_playbook: upgrade/04-install_proxy.yaml
 - import_playbook: upgrade/05-server_to_intermediate.yaml
+  when: forklift_upgrade_version_start != forklift_upgrade_version_intermediate
```
ekohl, line 7: One issue is that in CI it doesn't actually run this playbook but rather all individual ones

evgeni: oh. right. because we are clever.

evgeni: how about now?

evgeni: (it wouldn't have worked anyways, as the vars are only loaded inside the playbook)

PR #1146, opened 2020-04-13
EL8 selinux handling and ignore foreman-release-scl

ehelms:


roles/selinux/tasks/main.yml
```diff
@@ -1,19 +1,20 @@
...
+      when: ansible_distribution_major_version == '7'
 
-- name: 'Set selinux state'
-  selinux:
-    policy: targeted
-    state: "{{ selinux_state }}"
-  tags:
-    - env_setup
-  when: ansible_os_family == 'RedHat'
+    - name: 'Ensure libselinux-python is installed'
```
ekohl, line 8:
```suggestion
    - name: 'Ensure selinux packages are installed'
```

bats/fb-test-puppet.bats
```diff
@@ -54,15 +54,15 @@ fi
     modpath=/etc/puppet/environments/production/modules
   fi
 
-  if [ ! -d $modpath/ntp ]; then
-    puppet module install -i $modpath -v 4.2.0 puppetlabs/ntp
+  if [ ! -d $modpath/chrony ]; then
+    puppet module install -i $modpath -v 0.3.2 aboe/chrony
```
ehelms, line 58: Note I could not find NTP on EL8 as it is on EL7 so I switched this to chrony.

ekohl: Does it make sense to use something simple like motd? OTOH, the module looks simple enough and supports what we need. puppetlabs-motd looks more complex.

ehelms:
I could go either way. This is a drop in replacement as far as I can tell for what we had so it keeps that consistency.

bats/fb-test-puppet.bats
```diff
@@ -79,12 +79,12 @@ fi
...
-  id=$(hammer --csv puppet-class list --search 'name = ntp' | tail -n1 | cut -d, -f1)
+  id=$(hammer --csv puppet-class list --search 'name = chrony' | tail -n1 | cut -d, -f1)
   pc_ids=$(hammer host update --help | awk '/class-ids/ {print $1}')
   hammer host update $pc_ids $id --name $(hostname -f)
 }
 
 @test "apply class with puppet agent" {
   puppet agent -v -o --no-daemonize
-  grep -i puppet /etc/ntp.conf
+  grep -i puppet /etc/chrony.conf
```
ekohl, line 89: Turns out this module doesn't have the word `puppet` in the resulting file causing the tests to fail. To be clear, the test was to check for https://github.com/puppetlabs/puppetlabs-ntp/blob/29929d1f425e40134c51b9d11a87c0f71fcffab6/templates/ntp.conf.epp#L1 in the resulting file.

PR #1147, opened 2020-04-19
use correct chrony.conf

mmoll:


mmoll:
completely untested :bomb:

tbrisker:
Looks like the issue isn't the distro but rather the file that doesn't contain the string 'puppet' - https://github.com/theforeman/forklift/pull/1146#pullrequestreview-396275130

mmoll:
@tbrisker Both issues need fixing, AFIACT.

ehelms:
Could we do a more puppety test where we stop the service or remove a file, run puppet and then check that it exists again? Or do that form the get go?

ehelms:
Let's revert for now to keep EL7 going. https://github.com/theforeman/forklift/pull/1148

ehelms:
This is the chrony module (https://github.com/voxpupuli/puppet-chrony)  so we could propose adding the template header to it and then use it?

ekohl:
I also considered something like that, but relying on some complex module might be too much. Publishing a dummy module that statically manages something is much easier. It can be as simple as creating `/file-managed-by-puppet`

ehelms:
Proposed set of changes here: https://github.com/theforeman/foreman-infra/pull/1341

evgeni:
@mmoll closing in favor of @ehelms solution? :)

mmoll:
Ja. :)

PR #1150, opened 2020-04-22
Updates needed to run Katello on EL8

ehelms:


roles/katello_repositories/tasks/main.yml
```diff
@@ -2,6 +2,9 @@
 - name: "Set up {{ katello_repositories_environment }} repositories"
   include_tasks: "{{ katello_repositories_environment }}_repos.yml"
 
+- command: dnf module -y enable pki-core
```
evgeni, line 5: So I wanted to suggest to use the `dnf` module, but even after knowing about its bugs (https://github.com/ansible/ansible/issues/56504) I could not manage to enable the pki-core module with it.

evgeni: However, I wonder why this does not get autoenabled when we install candlepin?

ehelms: Just following up here, the pki-core module does not have a default profile which is the key aspect of modules getting auto-enabled.

PR #1154, opened 2020-04-27
Use --puppet-class-ids explicitly for puppet bats test

ehelms:


ekohl:
For historical reasons, the commit that introduced it: https://github.com/theforeman/foreman-bats/commit/609c22c80ea92a9c4d1f998d690d4b9e745d33bf

PR #1155, opened 2020-04-28
Include module_hotfixes when using staging repositories

ehelms:


ehelms:
There is no builtin module option for this config option so this was one way I figured how to do it.

ehelms:
Seems to ignore from my testing.

PR #1160, opened 2020-05-13
Add Foreman 2.1 to vagrant configs

ehelms:


pipelines/install/03-install_server.yaml
```diff
@@ -11,3 +11,5 @@
       scenario: "{{ pipeline_type }}"
       scenario_version: "{{ pipeline_version }}"
     - foreman_installer
+  post_tasks:
+    - shell: foreman-installer -v --katello-pulp-expose-db true
```
ekohl, line 15: This looks like another change that's not yet ready

ehelms: Doh, thought I cleaned my branch.

PR #1161, opened 2020-05-19
fix foreman version check in verify packages

evgeni:


ekohl:
This will break until https://github.com/theforeman/foreman-packaging/pull/5257 is resolved and picked into 2.*.

evgeni:
that's the intention, well, we should fix it, and then merge this :)

ekohl:
Just making sure that people who didn't follow our IRC conversation also know and don't just merge this :)

evgeni:
2.0 and 2.1 are fixed

PR #1167, opened 2020-06-01
Ansible lint cleanups

ekohl:
It's odd our ansible-lint didn't catch this. I suspect ansible-lint simply isn't doing anything right now.

### DIFF
--- a/playbooks/runcible.yml
+++ b/playbooks/runcible.yml
@@ -36,10 +36,10 @@
       become: true
 
     - name: "Install RVM gpg key"
-      shell: "gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3"
+      command: "gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3"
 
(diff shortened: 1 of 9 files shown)


playbooks/runcible.yml
```diff
@@ -36,10 +36,10 @@
       become: true
 
     - name: "Install RVM gpg key"
-      shell: "gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3"
+      command: "gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3"
 
     - name: "Install RVM"
-      shell: "curl -sSL https://get.rvm.io | bash -s stable --ruby"
+      command: "curl -sSL https://get.rvm.io | bash -s stable --ruby"
```
evgeni, line 42: a `command` can't contain a pipe

ekohl: Bleh. Should we just remove this playbook? I'm not sure anyone really uses it. Especially now that all the focus is on removing Pulp 2.

ehelms: Yes, you can remove this playbook.

PR #1173, opened 2020-07-02
Update bats to support no Pulp 2

ehelms:
This is also part of EL8 support which does not have Pulp 2.


bats/fb-content-katello.bats
```diff
@@ -211,32 +215,44 @@ setup() {
 }
 
 @test "install katello-agent" {
+  tSkipIfPulp2
```
ekohl, line 218: The naming of this feels a bit odd. This reads like it should skip katello-agent tests if it's Pulp 2 but the implementation suggests it's skipping unless Pulp 2 is present

ehelms:
Hah, I forgot the 'No' in there

bats/foreman_helper.bash
```diff
@@ -17,3 +17,13 @@ tForemanVersion() {
     fi
   ) | cut -d. -f1-2
 }
+
+Pulp2() {
```
evgeni, line 21:
```suggestion
tIsPulp2() {
```

bats/foreman_helper.bash
```diff
@@ -17,3 +17,13 @@ tForemanVersion() {
...
+  tPackageExists pulp-server
+}
+
+tSkipIfNoPulp2() {
+  if ! tIsPulp2; then
+   skip "Puppet content not available with Pulp 3 only"
```
evgeni, line 27: You're using this in more cases than Puppet, so we either should have a more generic message here, or multiple `skip` functions with different explicit messages.

evgeni:
the wording changes and removal of the always-skipped tests can go in a separate PR if you want, as they are not related anyways.

bats/foreman_helper.bash
ehelms: Now it supports multiple cases.

bats/fb-content-katello.bats
```diff
@@ -245,30 +261,20 @@ setup() {
 }
 
 @test "publish content view" {
+  tSkipIfNoPulp2 "Puppet content"
```
evgeni, line 264: publish and promote should still work with pulp3 only?

bats/fb-content-katello.bats
```diff
@@ -61,22 +61,26 @@ setup() {
   curl http://$HOSTNAME/pulp/isos/${ORGANIZATION_LABEL}/Library/custom/${PRODUCT_LABEL}/${REPOSITORY_LABEL}/1.iso > /dev/null
 }
 
-@test "create a docker repository" {
+@test "create a container repository" {
   hammer repository create --organization="${ORGANIZATION}" --docker-upstream-name="fedora/ssh" --url=https://registry-1.docker.io/ \
     --product="${PRODUCT}" --content-type="docker" --name "${DOCKER_REPOSITORY}" | grep -q "Repository created"
```
evgeni, line 66: should this then also be `CONTAINER_REPOSITORY`?

bats/fb-content-katello.bats
ehelms: They should, but this publish and promote were targeted as follow ons to adding the Puppet content so they are not really testing anything new.

bats/foreman_helper.bash
```diff
@@ -17,3 +17,13 @@ tForemanVersion() {
...
   ) | cut -d. -f1-2
 }
+
+tIsPulp2() {
+  tPackageExists pulp-server
+}
+
+tSkipIfNoPulp2() {
+  if ! tIsPulp2; then
+   skip "${1} not available with Pulp 3 only"
```
ekohl, line 27:
```suggestion
   skip "${1} unavailable without Pulp 2"
```

bats/fb-content-katello.bats
evgeni: oooh, right. should the test name then be adjusted to "$action content view with puppet content" or something?

PR #1174, opened 2020-07-07
move box exclusion to an own function and call it once

evgeni:
if we filter in `process_boxes!`, we might also filter out our base
boxes and then our real boxes end up without a `box_name` and thus can't
be created by vagrant at all.

instead let's reject all excluded boxes in a separate step

### DIFF
--- a/vagrant/lib/forklift/box_factory.rb
+++ b/vagrant/lib/forklift/box_factory.rb
@@ -30,10 +30,11 @@ module Forklift
     end
 
     def filter_boxes!
-      box_config = Settings.new.settings.key?('boxes') ? Settings.new.settings['boxes'] : {}
+      box_config = Settings.new.settings['boxes']
+      return unless box_config&.key?('exclude')
 
       @boxes.reject! do |name, _box|
-        box_config.key?('exclude') && box_config['exclude'].any? { |exclude| name.match(/#{exclude}/) }
+        box_config['exclude'].any? { |exclude| name.match(/#{exclude}/) }
       end
     end
 
--- a/vagrant/lib/forklift/box_factory.rb
+++ b/vagrant/lib/forklift/box_factory.rb
@@ -29,6 +29,14 @@ module Forklift
       end
     end
 
+    def filter_boxes!
+      box_config = Settings.new.settings.key?('boxes') ? Settings.new.settings['boxes'] : {}
+
+      @boxes.reject! do |name, _box|
+        box_config.key?('exclude') && box_config['exclude'].any? { |exclude| name.match(/#{exclude}/) }
+      end
+    end
+
     private
 
     def load_box_file(file)
@@ -38,10 +46,6 @@ module Forklift
 
     def process_boxes!(boxes)
       boxes.each do |name, box|
-        box_config = Settings.new.settings.key?('boxes') ? Settings.new.settings['boxes'] : {}
-
-        next if box_config.key?('exclude') && box_config['exclude'].any? { |exclude| name.match(/#{exclude}/) }
-
         box = layer_base_box(box)
         box['name'] = name
 
--- a/vagrant/lib/forklift/box_factory.rb
+++ b/vagrant/lib/forklift/box_factory.rb
@@ -31,8 +31,10 @@ module Forklift
 
     def filter_boxes!
       box_config = Settings.new.settings['boxes']
-      return unless box_config&.key?('exclude')
+      # rubocop:disable Style/SafeNavigation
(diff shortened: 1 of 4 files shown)


vagrant/test/lib/box_loader_test.rb
```diff
@@ -10,10 +10,24 @@ def test_load
...
+    excludes = { 'boxes' => { 'exclude' => ['fedora'] } }
+    Forklift::Settings.any_instance.stubs(:settings).returns(excludes)
+    locations = get_locations('00-fedora.yaml', '03-packaging.yaml')
+    loader = Forklift::BoxLoader.new(nil, locations)
+    loader.load!
+
+    assert loader.boxes
+    assert loader.boxes['rpm-packaging']
+    assert_equal 'rpm-packaging', loader.boxes['rpm-packaging']['name']
+    assert_equal 'fedora/30-cloud-base', loader.boxes['rpm-packaging']['box_name']
```
evgeni, line 28: this could also be a simple `assert` as all we care about is that it's set at all (which it wasn't before)

vagrant/lib/forklift/box_factory.rb
```diff
@@ -29,6 +29,14 @@ def add_boxes!(box_file)
       end
     end
 
+    def filter_boxes!
+      box_config = Settings.new.settings.key?('boxes') ? Settings.new.settings['boxes'] : {}
+
+      @boxes.reject! do |name, _box|
+        box_config.key?('exclude') && box_config['exclude'].any? { |exclude| name.match(/#{exclude}/) }
+      end
```
ekohl, line 37:
```suggestion
      box_config = Settings.new.settings['boxes']
      return unless box_config && box_config.key?('exclude')

      @boxes.reject! do |name, _box|
        box_config['exclude'].any? { |exclude| name.match(/#{exclude}/) }
      end
```

vagrant/lib/forklift/box_factory.rb
```diff
@@ -29,6 +29,15 @@ def add_boxes!(box_file)
       end
     end
 
+    def filter_boxes!
+      box_config = Settings.new.settings['boxes']
+      return unless box_config&.key?('exclude')
```
ehelms, line 34: Not seen that syntax before, what does it mean?

evgeni: Don't fail if `box_config` is `nil` and thus has no `key?`

http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

ekohl: This is Ruby 2.4+ I believe and unsupported on EL7, right?

vagrant/test/lib/box_loader_test.rb
evgeni: @ehelms opinions here?

vagrant/lib/forklift/box_factory.rb
evgeni: 2.3+ I think, but yes.

And before you ask: Vagrant on EL7 uses Ruby 2.4 that it bundles:

```console
# /opt/vagrant/embedded/bin/ruby --version
ruby 2.4.9p362 (2019-10-02 revision 67824) [x86_64-linux]
```

ehelms: That would be problematic for at least one of my setups. Let's use the old school method :)

vagrant/test/lib/box_loader_test.rb
ehelms: This looks good.

vagrant/lib/forklift/box_factory.rb
ekohl: @evgeni https://github.com/theforeman/forklift/blob/master/bin/ansible-vars would break

evgeni: Ewwww. OK

vagrant/lib/forklift/box_factory.rb
```diff
@@ -29,6 +29,17 @@ def add_boxes!(box_file)
       end
     end
 
+    def filter_boxes!
+      box_config = Settings.new.settings['boxes']
+      # rubocop:disable Style/SafeNavigation
```
ekohl, line 34: This suggests we should set our Rubocop rules to target 2.0 compliance.

evgeni: probably, yeah.

PR #1179, opened 2020-07-08
migrate to GH actions

evgeni:


evgeni:
example run: https://github.com/evgeni/forklift/runs/849422491

.github/workflows/main.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: vagrant
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.6'
```
ekohl, line 34: I think I tried to get Ruby 2.0 going as well but that's pretty much impossible nowadays in CI.

.github/workflows/main.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+      - name: Install ansible-lint
+        run: pip install ansible-lint
+      - name: Run tests
+        run: |
+          ansible-lint -r rules/ playbooks/*
+          ansible-lint -r rules/ roles/*
```
ekohl, line 22: Have you considered using https://github.com/marketplace/actions/ansible-lint

.github/workflows/main.yml
evgeni: We've been running on 2.5 before (that's what Bionic has by default) and I honestly don't care about 2.0 in CI.
If it ever breaks, it's probably easier to update the nodes we still run on 2.0 from EL7 to something more modern (via SCL or whatever).

.github/workflows/main.yml
evgeni: Not yet, no

.github/workflows/main.yml
ekohl: I think in CentOS CI we can request EL8 nodes instead. https://github.com/theforeman/foreman-infra/blob/612b8ec679506a4c3a2f3025e9655bf322945383/puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml#L6-L10 currently doesn't specify anything but https://github.com/redhat-openstack/ansible-role-ci-centos#role-variables suggests we can set release to 8. That would be a start.

.github/workflows/main.yml
evgeni: Now I did, and it uses an older version of ansible-lint and meh.

ekohl: I also tried it in the past and didn't really saw a benefit. At least last time it always needed to build the container so it wasn't faster.

playbooks/setup_forklift.yml
```diff
@@ -37,5 +37,5 @@
 
     - name: 'Configure Forklift'
       copy:
-        content: "{{ forklift_config | to_nice_yaml }}"
+        content: "{{ forklift_config | to_json | from_json | to_nice_yaml }}"
```
ekohl, line 40: Why is this needed?

evgeni: `ansible-lint` (and only that, `ansible-playbook` is fine) fails to parse the playbook otherwise. I'll open a bug against `ansible-lint`, but so far this workaround was the cheapest to get it running.

ekohl: Mind adding that as a comment, ideally with a link to the ansible-lint issue?

evgeni: done

PR #1180, opened 2020-07-08
add molecule tests for forklift_versions

evgeni:
includes #1179

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,22 @@ jobs:
           ansible-lint -r rules/ playbooks/*
           ansible-lint -r rules/ roles/*
 
+  molecule:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+      - name: Install molecule
+        run: pip install molecule
+      - name: Run tests
+        run: |
+          for roledir in roles/*/molecule; do
+            pushd $(dirname $roledir)
+            molecule test
+            popd
+          done
+
   vagrant:
     runs-on: ubuntu-latest
     defaults:
--- a/roles/forklift_versions/molecule/default/molecule.yml
+++ b/roles/forklift_versions/molecule/default/molecule.yml
@@ -1,0 +1,9 @@
+---
+driver:
+  name: delegated
+  options:
+    managed: False
+    ansible_connection_options:
+      ansible_connection: local
+platforms:
+ - name: instance
--- a/roles/forklift_versions/.yamllint
+++ b/roles/forklift_versions/.yamllint
@@ -1,0 +1,33 @@
+---
+# Based on ansible-lint config
+extends: default
+
+rules:
+  braces:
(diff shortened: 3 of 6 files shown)


evgeni:
example run: https://github.com/evgeni/forklift/runs/849448248?check_suite_focus=true

.github/workflows/main.yml
```diff
@@ -21,6 +21,22 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+      - name: Install molecule
+        run: pip install molecule
+      - name: Run tests
+        run: |
+          for folder in roles/*/molecule; do
+            pushd $(dirname $folder)
```
ekohl, line 35: Generally not a fan of `folder`
```suggestion
          for directory in roles/*/molecule; do
            pushd $(dirname $directory)
```

roles/forklift_versions/molecule/default/molecule.yml
```diff
@@ -0,0 +1,13 @@
...
+  options:
+    managed: False
+    ansible_connection_options:
+      ansible_connection: local
+platforms:
+ - name: instance
+#provisioner:
+#  name: ansible
+#verifier:
+#  name: ansible
```
ekohl, line 13: Not a fan of commented "code". Is this needed in the future?

roles/forklift_versions/molecule/default/INSTALL.rst
```diff
@@ -0,0 +1,22 @@
+*******
+Docker driver installation guide
+*******
+
+Requirements
+============
+
+* Docker Engine
```
ekohl, line 8: Is it possible to run with podman instead of docker?

evgeni: yeah, or no container at all, like I actually do here, so removed that file.

roles/forklift_versions/molecule/default/molecule.yml
evgeni: if it is, it can be added.

evgeni:
The `.yamllint` file we can easily move to an central place and symlink. The other files don't really feel like *much* boilerplate. `molecule.yml` might be, depending on whether we end up having identical/similar executors. A symlink might work here too.

ehelms:
For us molecule newbs, could you expand on how it's intended to be used in a project and how that informs this design choice?

evgeni:
So molecule is a "test executor" for Ansible, probably comparable with Beaker for Puppet.
At a minimum, a test has two steps: converge (run the role using ansible) and verify (do some asserts, somehow). This can be extended to install dependencies, start up services, whatever is needed. It can also run tests in a container (docker and podman are natively supported) or on a regular local or remote host (like I do above).

Instead of molecule you could (in this particular case) also just take the verify playbook and execute it with `ansible-playbook` (essentially how we do it in FAM right now). This would certainly work for `forklift_versions` as that role is essentially just "set some facts" but would not work that well if we wanted to add tests for roles that actually modify systems.

PR #1182, opened 2020-07-10
call yamllint from a central place

evgeni:


evgeni:
@ekohl there we go, de-boilerplated yamllint :)

evgeni:
A symlink you need to remember, the config is there. Plus we can migrate more from molecule.yaml there when we see that stuff is generic.

PR #1185, opened 2020-07-20
minor changes to Ansible instructions

douglasawh:
I don't think tying to a specific version of foreman and katello in the docs is a good idea, unless someone is going to update it for every version. Also, some changes to the local deployment. Obviously, this is not the place for full Ansible instructions, but if people are new to repos, the fact that 'yum install ansible' does not work without EPEL is an easy blocker to avoid.

### DIFF
--- a/docs/production.md
+++ b/docs/production.md
@@ -71,7 +71,7 @@ This use case assumes that you are on a box that has SSH access to the target bo
 
 For a release version in production:
 
-    ansible-playbook -l <target-host> playbooks/katello.yml -e foreman_repositories_version=1.17 -e katello_repositories_version=3.6
+    ansible-playbook -l <target-host> playbooks/katello.yml -e foreman_repositories_version=VERSIONYOUWANT -e katello_repositories_version=VERSIONYOUWANT
 
 For nightly production:
 
@@ -85,13 +85,14 @@ After installing a Katello server, you could then spin up a Capsule with the ass
 ### Local Deployment
 
 1. ssh to target machine **as root**
-2. Install Ansible -- `yum install ansible`
+2. Install Ansible -- `yum install ansible` (this will require the EPEL repo)
 3. Clone this repository - `git clone https://github.com/theforeman/forklift.git`
 4. Enter the repository - `cd forklift`
+5. Make sure DNS is set up properly. This is a Foreman requirement, not an Ansible requirement.
 
 For a release version in production:
 
-    ansible-playbook -l localhost playbooks/katello.yml -e foreman_repositories_version=1.17 -e katello_repositories_version=3.6
+    ansible-playbook -l localhost playbooks/katello.yml -e foreman_repositories_version=VERSIONYOUWANT -e katello_repositories_version=VERSIONYOUWANT
 
 For nightly production:
 
--- a/docs/production.md
+++ b/docs/production.md
@@ -52,7 +52,7 @@ Start the installation for CentOS 7:
 
     vagrant up centos7-katello-3.4
 
-
+This will create a libvirt based virtual machine running the Katello server on CentOS.
 
 ## BYOB
 
--- a/docs/production.md
+++ b/docs/production.md
@@ -85,7 +85,7 @@ After installing a Katello server, you could then spin up a Capsule with the ass
 ### Local Deployment
 
 1. ssh to target machine **as root**
-2. Install Ansible -- `yum install ansible` (this will require the EPEL repo)
+2. Install Ansible -- `yum install ansible` (for example from the EPEL repo)
 3. Clone this repository - `git clone https://github.com/theforeman/forklift.git`
 4. Enter the repository - `cd forklift`
 5. Make sure DNS is set up properly. This is a Foreman requirement, not an Ansible requirement.
--- a/docs/production.md
+++ b/docs/production.md
@@ -52,7 +52,7 @@ Start the installation for CentOS 7:
 
     vagrant up centos7-katello-3.4
 
-This will create a libvirt based virtual machine running the Katello server on CentOS.
+This will create a libvirt based virtual machine running the Katello server on CentOS. 
 
 
 ## BYOB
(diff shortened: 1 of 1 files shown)


docs/production.md
```diff
@@ -71,7 +71,7 @@ This use case assumes that you are on a box that has SSH access to the target bo
 
 For a release version in production:
 
-    ansible-playbook -l <target-host> playbooks/katello.yml -e foreman_repositories_version=1.17 -e katello_repositories_version=3.6
+    ansible-playbook -l <target-host> playbooks/katello.yml -e foreman_repositories_version=VERSIONYOUWANT -e katello_repositories_version=VERSIONYOUWANT
```
ekohl, line 74: Perhaps it should be `WANTED_FOREMAN_VERSION` and `WANTED_KATELLO_VERSION` to indicate they're not the same value? While on the subject, perhaps it should also mention that it's up to the user to make sure the versions are compatible

douglasawh: @ekohl those are both great suggestions, but I don't know how to act on them without creating another PR. Is that the way to do it?

ekohl: There is an Edit file link. For me it points to https://github.com/douglasawh/forklift/edit/patch-1/docs/production.md?pr=%2Ftheforeman%2Fforklift%2Fpull%2F1185. That's probably the easiest way.

docs/production.md
```diff
@@ -85,13 +85,14 @@ After installing a Katello server, you could then spin up a Capsule with the ass
 ### Local Deployment
 
 1. ssh to target machine **as root**
-2. Install Ansible -- `yum install ansible`
+2. Install Ansible -- `yum install ansible` (this will require the EPEL repo)
```
ehelms, line 88: There are many places to get Ansible, (e.g. http://mirror.centos.org/centos/7/configmanagement/x86_64/ or RHEL repos) so perhaps we should leave this up to the user to decide.

ekohl:
```suggestion
2. Install Ansible -- `yum install ansible` (for example from the EPEL repo)
```

douglasawh: oh, I didn't mean to resolve the whole thing, just the "for example". I guess I will Unresolve it until I can verify that "Commit suggestion" option actually did what I wanted.

douglasawh: ...and now I see how the blocks are separated, so all g on this one, I think.

docs/production.md
ekohl:
```suggestion
    ansible-playbook -l <target-host> playbooks/katello.yml -e foreman_repositories_version=WANTED_FOREMAN_VERSION -e katello_repositories_version=WANTED_KATELLO_VERSION
```

docs/production.md
```diff
@@ -85,13 +85,14 @@ After installing a Katello server, you could then spin up a Capsule with the ass
...
-2. Install Ansible -- `yum install ansible`
+2. Install Ansible -- `yum install ansible` (for example from the EPEL repo)
 3. Clone this repository - `git clone https://github.com/theforeman/forklift.git`
 4. Enter the repository - `cd forklift`
+5. Make sure DNS is set up properly. This is a Foreman requirement, not an Ansible requirement.
 
 For a release version in production:
 
-    ansible-playbook -l localhost playbooks/katello.yml -e foreman_repositories_version=1.17 -e katello_repositories_version=3.6
+    ansible-playbook -l localhost playbooks/katello.yml -e foreman_repositories_version=VERSIONYOUWANT -e katello_repositories_version=VERSIONYOUWANT
```
ekohl, line 95:
```suggestion
    ansible-playbook -l localhost playbooks/katello.yml -e foreman_repositories_version=WANTED_FOREMAN_VERSION -e katello_repositories_version=WANTED_KATELLO_VERSION
```

docs/production.md
```diff
@@ -89,6 +90,7 @@ After installing a Katello server, you could then spin up a Capsule with the ass
 3. Clone this repository - `git clone https://github.com/theforeman/forklift.git`
 4. Enter the repository - `cd forklift`
 5. Make sure DNS is set up properly. This is a Foreman requirement, not an Ansible requirement.
+6. Determine the compatible versions of Foreman and Katello you want to install based on the Katello install instructions at https://www.theforeman.org/plugins/katello/3.16/installation/index.html. Replace the 3.16 in the URL with whatever version of Katello you want to install.
```
ekohl, line 93: I'd recommend https://github.com/theforeman/forklift/blob/master/vagrant/config/versions.yaml as a source

docs/production.md
```diff
@@ -52,7 +52,7 @@ Start the installation for CentOS 7:
 
     vagrant up centos7-katello-3.4
 
-This will create a libvirt based virtual machine running the Katello server on CentOS.
+This will create a libvirt based virtual machine running the Katello server on CentOS.
```
ekohl, line 55: Please leave this part out
```suggestion
This will create a libvirt based virtual machine running the Katello server on CentOS.
```

douglasawh: I'm not sure I incorporated this correctly. I think perhaps you meant "Please don't leave this part out". I'm not sure if the "Suggested change" is referring to my change or you are making the suggestion. The two diffs look the same.

douglasawh: @ekohl is there anything else that is needed on this?

docs/production.md
```diff
@@ -52,7 +52,6 @@ Start the installation for CentOS 7:
 
     vagrant up centos7-katello-3.4
 
-This will create a libvirt based virtual machine running the Katello server on CentOS.
```
ekohl, line 55: Why was this removed?

docs/production.md
```diff
@@ -68,10 +67,11 @@ This use case assumes that you are on a box that has SSH access to the target bo
 3. Enter the repository - `cd forklift`
 4. [Create an inventory file](https://docs.ansible.com/ansible/latest/intro_inventory.html) with your hosts in the `inventories` subfolder. You can also pass a different inventory to `ansible-playbook` using `--inventory` or `-i`.
 5. Change the variable remote_user in ./ansible.cfg to the appropriate value (the user that will log into the remote machine)
+6. Determine the compatible versions of Foreman and Katello you want to install based on the Katello install instructions at https://github.com/theforeman/forklift/blob/master/vagrant/config/versions.yaml. Replace the 3.16 in the URL with whatever version of Katello you want to install.
```
ekohl, line 70: Reading this, I don't know what `3.16` refers to since I don't see any URL.

ekohl (commented):
Apologies, it slipped from my radar. If it takes too long for me to review, a ping is the right thing to do.

docs/production.md
douglasawh: @ekohl the old URL was https://www.theforeman.org/plugins/katello/3.16/installation/index.html but that got changed in a previous commit. So, yeah, the reference to 3.16 needs to be removed.

douglasawh: fixed here: https://github.com/theforeman/forklift/pull/1185/commits/2a2728132523822d3e0d78f084ef3b1e999a21de

docs/production.md
douglasawh: fixed in https://github.com/theforeman/forklift/pull/1185/commits/60f2326b96424c4d1faf6c644e389b4d49282844

docs/production.md
douglasawh: fixed in https://github.com/theforeman/forklift/pull/1185/commits/60f2326b96424c4d1faf6c644e389b4d49282844

ekohl (approved):
Thanks!

PR #1187, opened 2020-07-21
Don't run with --upgrade in content proxy upgrade

tbrisker:


tbrisker:
In that case, I'm afraid  don't know how to accomplish it, i'm not familiar enough with how all these roles and playbooks come together in the ci pipelines

ehelms:
Yes, this will need a version check to ensure previous upgrades continue to work. Here is an example of how we did it:

https://github.com/theforeman/forklift/blob/master/roles/foreman_installer/tasks/upgrade.yml#L12

tbrisker:
My ansible knowledge is quite limited, but wouldn't that generic upgrade task already take care of it since it's included in this task?

ekohl:
I had a look a while back gave up on fixing it since the Ansible flow in combination with the Katello upgrade drove me insane. Dropping the need for special casing Katello upgrades will fix it in the long run.

evgeni:
#1197 was merged

PR #1189, opened 2020-07-29
allow to select upgrades with step != 1

evgeni:


roles/forklift_versions/library/forklift_versions.py
```diff
@@ -24,8 +24,12 @@
     description: scenario version to look up
     required: True
   scenario_os:
-    descrption: scenario OS to look up
+    description: scenario OS to look up
     required: True
+  scenario_step:
```
evgeni, line 29: bikeshed: `upgrade_step`?

roles/forklift_versions/library/forklift_versions.py
```diff
@@ -24,8 +24,12 @@
     description: scenario version to look up
     required: True
   scenario_os:
-    descrption: scenario OS to look up
+    description: scenario OS to look up
     required: True
+  step:
```
evgeni, line 29: bikeshed: `upgrade_step`?

roles/forklift_versions/molecule/default/verify.yml
```diff
@@ -38,6 +38,22 @@
...
 
+    - name: "Include forklift_versions for upgrades with step=2"
+      include_role:
+        name: "forklift_versions"
+      vars:
+        scenario: "{{ pipeline_type }}"
+        scenario_os: "{{ pipeline_os }}"
+        scenario_version: "3.16"
+        upgrade: True
+        scenario_step: 2
```
ekohl, line 49: This doesn't match the `step` used everywhere.

roles/forklift_versions/library/forklift_versions.py
ekohl: Yes, I think that would make more sense.

roles/forklift_versions/molecule/default/verify.yml
evgeni: the playbook in the role uses `scenario_step` that gets passed to the `step` param, but yeah, alligning those is nice

PR #1191, opened 2020-08-05
Update centos medium name for 2.1

xprazak2:
There are media for CentOS 7 and 8
which causes the os creation to fail.


xprazak2:
Yes, https://github.com/theforeman/foreman/commit/f41cb5b992d4a6bc21240345f7a91c80c90a5866

PR #1192, opened 2020-08-07
Add Foreman 2.2 config

ehelms:


evgeni:
@ehelms this broke molecule tests as they don't expect this upgrade path -- would you mind either updating the tests to match, or (probably better) modify the tests to check a specific version for upgrades, not nightly?

evgeni:
@ehelms you'll also have to update https://github.com/theforeman/forklift/blob/master/roles/forklift_versions/molecule/default/verify.yml#L11

roles/forklift_versions/molecule/default/verify.yml
```diff
@@ -65,6 +65,6 @@
     - name: Ensure 3.15 was properly skipped
       assert:
         that:
-          - forklift_upgrade_version_start == '3.16'
-          - forklift_upgrade_version_intermediate == '3.16'
-          - forklift_upgrade_version_final == 'nightly'
+          - forklift_upgrade_version_start == '3.15'
+          - forklift_upgrade_version_intermediate == '3.15'
+          - forklift_upgrade_version_final == '3.16'
```
evgeni, line 70: I think the correct upgrade path here is 16, 16, 17, as 15 doesn't have EL8

roles/forklift_versions/molecule/default/verify.yml
```diff
@@ -8,7 +8,7 @@
     forklift_versions_file: "{{ playbook_dir }}/fixtures/versions.yaml"
     pipeline_type: katello
     pipeline_os: centos7
-    pipeline_version: nightly
+    pipeline_version: 3.17
```
evgeni, line 11: This should probably be a string, even though it doesn't break here today ;)

roles/forklift_versions/molecule/default/verify.yml
```diff
@@ -20,8 +20,8 @@
     - name: Ensure versions have been set correctly
       assert:
         that:
-          - foreman_repositories_version == 'nightly'
-          - katello_repositories_version == 'nightly'
+          - foreman_repositories_version == '3.17'
```
evgeni, line 23: 2.2 ;)

ehelms: Oh yea, I just wanted you to see your testing get put through it's paces ;)

PR #1195, opened 2020-08-13
Drops haveged from use in Foreman role, restricts EPEL to EL7

ehelms:


ehelms:
Designed to help fix EL8 upgrade job failures

PR #1197, opened 2020-08-24
Drop --upgrade explicitly on 3.17+ upgrade

ehelms:
Current cause of 3.17 RC1 release failure.

### DIFF
--- a/roles/foreman_proxy_content/tasks/upgrade.yml
+++ b/roles/foreman_proxy_content/tasks/upgrade.yml
@@ -9,7 +9,6 @@
     foreman_installer_scenario: foreman-proxy-content
     foreman_installer_disable_system_checks: True
     foreman_installer_options_internal_use_only:
-      - --upgrade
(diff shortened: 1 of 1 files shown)


ekohl (commented):
Is this an issue on older installs? In upstream we have 3.15 and 3.16 supported, which means we also upgrade to 3.14. AFAIK it was still needed there so any 3.15.z release may break?

evgeni:
Does it? https://www.github.com/theforeman/forklift/tree/master/roles%2Fforeman_installer%2Ftasks%2Fupgrade.yml adds --upgrade on these old versions, so it should be safe to drop it here?

ekohl:
If that's true, then also `--certs-update-all` is redundant.

evgeni:
That's probably correct. Let's drop the offending `--upgrade`, see that it works and clean up afterwards.

evgeni (approved):

PR #1198, opened 2020-08-31
use new pulpcore repos for katello 3.18+

evgeni:


roles/forklift_versions/library/forklift_versions.py
```diff
@@ -87,6 +87,8 @@ def main():
                         'pulp_repositories_version': version['pulp'],
                         'puppet_repositories_version': version['puppet'],
                         }
+                if 'pulpcore' in version:
```
evgeni, line 90: this is separate, so the var is undefined for versions that don't use the split pulpcore

vagrant/config/versions.yaml
```diff
@@ -90,6 +90,7 @@ installers:
   - foreman: 'nightly'
     katello: 'nightly'
     pulp: '2.21'
+    pulpcore: '3.4'
```
evgeni, line 93: same q @jlsherrill  ;)

ekohl (commented):
Does it make sense to define a `pulpcore_repositories` role and include that if needed?

evgeni:
I considered that, but thought it's only useful when one wants to use pulpcore without Katello, which is out of scope for this repo?

roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -53,7 +53,20 @@
...
-    - katello_repositories_version == "nightly" or katello_repositories_version is version('3.15', '>=')
+    - katello_repositories_version != 'nightly'
+    - katello_repositories_version is version('3.15', '>=')
+    - katello_repositories_version is version('3.17', '<=')
+
+- name: 'Add Pulpcore repository'
+  yum_repository:
+    name: pulpcore-repository
+    description: Pulpcore repository for Katello {{ katello_repositories_version }}
+    baseurl: "https://yum.theforeman.org/pulpcore/{{ katello_repositories_pulpcore_version }}/el{{ ansible_distribution_major_version }}/x86_64/"
```
ekohl, line 64: This doesn't look like the staging URL Is that intentional? Should there be some `pulpcore-release` package inside that repo itself?

roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -53,7 +53,20 @@
...
+    gpgcheck: no
```
ekohl, line 65: Any plans to GPG sign pulpcore releases?

vagrant/config/versions.yaml
jlsherrill: 3.6 all the way

roles/katello_repositories/tasks/staging_repos.yml
evgeni: There is no staging as such (well, besides repoclosure). I can add such a package, yeah.

roles/katello_repositories/tasks/staging_repos.yml
evgeni: Yes, it's on my todo for this week.

evgeni:
@ekohl I've posted the alternative with a separate `pulpcore_repositories` role in https://github.com/theforeman/forklift/pull/1199, which one do you like better?

evgeni:
closing in favor of #1199

PR #1199, opened 2020-09-01
use new pulpcore repos for katello 3.18+

evgeni:
this is an alternative to https://github.com/theforeman/forklift/pull/1198


ekohl (approved):
I prefer this version, but I'll let others weigh in.

PR #1204, opened 2020-09-10
only add a katello proxy box for el7

evgeni:
This PR is not a PR of honor... no highly esteemed deed is commemorated here... nothing valued is here.

And yet, it works.

Please revert this disgrace as soon as we can have EL8 proxies again.

### DIFF
--- a/pipelines/vars/forklift_katello.yml
+++ b/pipelines/vars/forklift_katello.yml
@@ -14,7 +14,7 @@ server_box:
         - "--puppet-server-jvm-min-heap-size 1G"
         - "--puppet-server-jvm-max-heap-size 1G"
       bats_tests_additional:
-        - fb-proxy.bats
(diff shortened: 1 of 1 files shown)


evgeni:
@ehelms so that took 30 minutes of time, but I don't know how many puppies will be killed because of it.

ehelms:
@evgeni were you able to run through el7, and el8 install and upgrade tests locally? If not, I can run through a few.

evgeni:
I only made the config generate properly, didn't run anything.

evgeni:
i ran install pipes for el7/el8 and while el8 failed with an weird error that seems unrelated to the change here

evgeni:
okay, the weird error I have been seein' on EL8 is fixed by https://github.com/theforeman/forklift/pull/1205 and unrelated

PR #1205, opened 2020-09-11
disable GPG checking when installing RPMs from an URL

evgeni:
Ansible started checking GPG signatures also for "local" RPMs in 2.9.13
(see https://github.com/ansible/ansible/blob/stable-2.9/changelogs/CHANGELOG-v2.9.rst#security-fixes)
    
However, this poses a chicken & egg problem, as in these cases we
usually don't have the GPG key yet. So explicitly disable the check for
this transaction and expect the installed RPM to contain the key for the
following interactions.

Also drop puppet 3 and 4 repos


roles/epel_repositories/tasks/main.yml
```diff
@@ -2,6 +2,7 @@
 - name: 'Setup Epel Repository'
   yum:
     name: "{{ epel_repositories_name }}"
+    disable_gpg_check: True
```
ekohl, line 5: AFAIK this isn't always needed because on CentOS we typically install `epel-release` from CentOS extras which is GPG signed. Perhaps we need to split this into a RHEL and CentOS task?

roles/puppet_repositories/tasks/puppetlabs-5-redhat.yml
```diff
@@ -2,6 +2,7 @@
 - name: 'Setup Puppet 5 Repository'
   yum:
     name: https://yum.puppet.com/puppet5-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
```
ekohl, line 4: 5 and 6 are now very similar. With Puppet 7 coming soon, I wonder if we can use variables

roles/puppet_repositories/tasks/puppetlabs-nightly-redhat.yml
```diff
@@ -2,6 +2,7 @@
 - name: 'Setup Puppet 6 Nightly Repository'
   yum:
     name: https://yum.puppet.com/puppet6-nightly-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
```
ekohl, line 4: This doesn't appear to exist anymore so we can also drop it.

roles/epel_repositories/tasks/main.yml
evgeni: No, we always install from URL:
https://github.com/theforeman/forklift/blob/8139cc3eaf2323013c0073e331ad69b482425cf1/roles/epel_repositories/defaults/main.yml#L4

roles/puppet_repositories/tasks/puppetlabs-5-redhat.yml
evgeni: Probably. But 6 is also purging 5, so it's more than one simple variable.

I'd defer that to a later point.

roles/epel_repositories/tasks/main.yml
evgeni: (unless the user overrides this to `epel-release`, but ain't nobody does that)

ekohl: hmm ok

roles/puppet_repositories/tasks/puppetlabs-5-redhat.yml
ekohl: Right. We do that because they conflict, but don't upgrade in place. I opened https://tickets.puppetlabs.com/browse/CPR-762

ehelms (approved):

Issue #1206, opened 2020-09-11
katello tests fail on el8 box


evgeni:
I've been poking Katello on EL8 (austin_powers-i_like_to_live_dangerously.jpg) and encountered the following bats errors:

```
    1..1
    ok 1 verify no old SCL packages
    1..3
    ok 1 # skip (FOREMAN_EXPECTED_VERSION is not set, not asserting) assert correct Foreman version is installed
    ok 2 check web app is up
    ok 3 check smart proxy is registered
    1..10
    ok 1 check smart proxy is registered
    ok 2 assert puppet version
    ok 3 wake up puppet agent
    ok 4 check host is registered
    ok 5 install puppet module
    ok 6 import motd puppet class
    ok 7 Assign environment to default taxonomies
    ok 8 # skip (Assignment not needed) Assign proxy to default taxonomies
    ok 9 assign puppet class to host
    ok 10 apply class with puppet agent
    1..2
    ok 1 check hammer ping
    ok 2 check service status
    1..57
    ok 1 Create an Empty Organization
    ok 2 create an Organization
    ok 3 create a product
    ok 4 create package repository
    ok 5 upload package
    ok 6 sync repository
    ok 7 create a file repository
    ok 8 sync file repository
    ok 9 fetch file from file repository
    ok 10 create a container repository
    ok 11 sync container repository
    ok 12 # skip (Puppet content is not available in scenarios without Pulp 2) create puppet repository
    ok 13 # skip (Puppet content is not available in scenarios without Pulp 2) upload puppet module
    ok 14 create lifecycle environment
    ok 15 create content view
    ok 16 add yum repo to content view
    ok 17 publish content view
    ok 18 promote content view
    ok 19 create activation key
    ok 20 disable auto-attach
    ok 21 add subscription to activation key
    ok 22 install subscription manager
    ok 23 disable puppet agent to prevent checkin from registering host to another org
    ok 24 delete host if present
    not ok 25 register subscription manager with username and password
    # (in test file fb-content-katello.bats, line 170)
    #   `subscription-manager register --force --org="${ORGANIZATION_LABEL}" --username=admin --password=changeme --env=Library' failed with status 70
    # rc=1
    # This system is currently not registered.
    # rc=0
    # All local data removed
    #
    # WARNING
    #
[...]

PR #1207, opened 2020-09-11
properly query non-SCL'ed hammer version on EL8

evgeni:
Refs: #1206

### DIFF
--- a/bats/foreman_helper.bash
+++ b/bats/foreman_helper.bash
@@ -29,7 +29,12 @@ tSkipIfNoPulp2() {
 }
 
 tSkipIfHammerBelow018() {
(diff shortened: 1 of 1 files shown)


ehelms (approved):

ekohl:
Can't we add a `Provides: hammer-cli = %{version}` to our spec and query on that? The benefit is that you can also simply run `yum install hammer-cli`. Similarly, we have `foreman-plugin-$plugin` provides. That reminds me to check back if Puppet supports this nowadays.

evgeni:
@ekohl we can (and should), but you can't `rpm -q` on a provides, so in this particular case we'd still be out of luck how to find the version of the installed hammer.

ekohl:
Ah, that's probably also why in the past Puppet didn't support it properly - it used `rpm` to detect the current state.

PR #1208, opened 2020-09-16
Allow use of foreman-installer PR

wbclark:
This is a very rough draft of an idea to use a foreman-installer PR with forklift

Actually currently it allows any fork and branch to be used, but could be updated to specify a PR number instead

The RPM is still installed but we are doing `git checkout -f` over it which is kind of hackish but I need to figure out a strategy to deal with the `modules/` subdirectory.

### DIFF
--- a/roles/foreman_installer/defaults/main.yml
+++ b/roles/foreman_installer/defaults/main.yml
@@ -19,6 +19,9 @@ foreman_installer_module_prs_strategy: "merge"
 # Comma-separated list of "organization/module/branch", e.g. "theforeman/foreman/foo,theforeman/katello/bar"
 foreman_installer_module_branches: []
 
+foreman_installer_from_source: False
+foreman_installer_git_pr_number:
+
 # There are two options, so a user can supply their own, and a playbook can
 # specify theirs. For example, foreman-proxy needs "--foreman-proxy-foreman-url"
 # so we put it in internal_use_only in a role or playbook.
--- a/roles/foreman_installer/tasks/main.yml
+++ b/roles/foreman_installer/tasks/main.yml
@@ -1,8 +1,16 @@
 ---
+- name: "build installer from source when installer PR is specified"
+  set_fact:
+    foreman_installer_from_source: True
+  when: foreman_installer_git_pr_number is defined and foreman_installer_git_pr_number
+
 - include_tasks: packages.yml
 
 - include_tasks: locales.yml
   when: ansible_os_family == 'Debian'
+
+- include_tasks: installer_from_source.yml
+  when: foreman_installer_from_source == True
 
 - include_tasks: module_prs.yml
   when: (foreman_installer_module_prs|length > 0) or (foreman_installer_module_branches|length > 0)
--- a/roles/foreman_installer/tasks/install.yml
+++ b/roles/foreman_installer/tasks/install.yml
@@ -22,6 +22,15 @@
     - { new: '--foreman-initial-admin-password', old: '--foreman-admin-password' }
     - { new: '--foreman-initial-admin-username', old: '--foreman-admin-username' }
 
+- name: 'Replace installer command when using installer source'
+  set_fact:
+    foreman_installer_command: scl enable {{ scl_ruby_version.stdout }} -- bundle exec bin/foreman-installer
+  when: foreman_installer_from_source == True
+
+- name: 'setup puppetserver ca'
+  shell: /opt/puppetlabs/bin/puppetserver ca setup
+  when: foreman_installer_from_source == True
+
 - name: 'Run installer'
   shell: >
       {{ foreman_installer_command }} {{ (foreman_installer_verbose|bool) | ternary("-v", "") }}
@@ -30,5 +39,7 @@
       {{ foreman_installer_scenario_flag }} {{ foreman_installer_scenario }}
       {{ foreman_installer_options_joined }}
   when: not foreman_installer_skip_installer
+  args:
+    chdir: /usr/share/foreman-installer/
   tags:
     - installation
--- a/roles/foreman_installer/tasks/installer_from_source.yml
+++ b/roles/foreman_installer/tasks/installer_from_source.yml
@@ -1,0 +1,53 @@
+---
+- name: "clone foreman-installer repository"
+  git:
+    repo: https://github.com/theforeman/foreman-installer.git
+    dest: "/usr/share/foreman-installer"
+    force: yes
+
+- name: "checkout foreman-installer PR"
+  shell: >
+      git fetch origin pull/{{ foreman_installer_git_pr_number }}/head:pr &&
+      git checkout pr
+  args:
+    chdir: "/usr/share/foreman-installer"
+  when: foreman_installer_git_pr_number is defined and foreman_installer_git_pr_number
+
+- name: "ensure ruby-augeas in Gemfile"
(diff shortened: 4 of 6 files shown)


roles/foreman_installer/tasks/installer_pr.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: "install git"
+  package:
+    name: 'git'
+    state: 'present'
+  tags:
+    - packages
+
+- name: "fetch git PR"
```
evgeni, line 9: we have something very similar in the `module_pr` playbook:

https://github.com/theforeman/forklift/blob/bb09a2c9bf9ba0862706f12b0005504717129cd3/roles/foreman_installer/tasks/module_pr.yml#L32-L54

I wonder if we could slim down the `shell` part a bit here?

roles/foreman_installer/defaults/main.yml
```diff
@@ -19,6 +19,10 @@ foreman_installer_module_prs_strategy: "merge"
 # Comma-separated list of "organization/module/branch", e.g. "theforeman/foreman/foo,theforeman/katello/bar"
 foreman_installer_module_branches: []
 
+foreman_installer_from_source: False
+installer_fork: theforeman
+installer_branch: develop
```
evgeni, line 24:
```suggestion
foreman_installer_git_fork: theforeman/foreman-installer
foreman_installer_git_branch: develop
```
Should we also support PR instead of branch like `module_pr.yml`?

roles/foreman_installer/tasks/installer_pr.yml
```diff
@@ -0,0 +1,17 @@
...
+  shell: >
+      git init &&
+      git remote add {{ installer_fork }}  https://github.com/{{ installer_fork }}/foreman-installer.git &&
+      git fetch &&
+      git branch {{ installer_branch }} origin/{{ installer_branch }} &&
+      git checkout -f {{ installer_branch }}
```
evgeni, line 15:
```suggestion
      git checkout --force --branch {{ installer_branch }} origin/{{ installer_branch }}
```

technically, there is no need for `--branch …` as we can also directly checkout the remote branch and don't care that we don't have a local one in this case

roles/foreman_installer/tasks/installer_pr.yml
```diff
@@ -0,0 +1,17 @@
...
+    name: 'git'
+    state: 'present'
+  tags:
+    - packages
+
+- name: "fetch git PR"
+  shell: >
+      git init &&
+      git remote add {{ installer_fork }}  https://github.com/{{ installer_fork }}/foreman-installer.git &&
+      git fetch &&
```
evgeni, line 13:
```suggestion
      git fetch {{ installer_fork }} &&
```

roles/foreman_installer/tasks/installer_pr.yml
```diff
@@ -0,0 +1,17 @@
...
+  package:
+    name: 'git'
+    state: 'present'
+  tags:
+    - packages
+
+- name: "fetch git PR"
+  shell: >
+      git init &&
+      git remote add {{ installer_fork }}  https://github.com/{{ installer_fork }}/foreman-installer.git &&
```
ekohl, line 12: Should this be simplified to `git clone --branch {{ installer_branch }} https://github.com/{{ installer_fork }}/foreman-installer.git /usr/share/foreman-installer`

evgeni: Given the destination folder already exists, this will fail.

ekohl: Right. I think it may make sense to replace the package altogether and do a pure source installation. Right now it doesn't pull in the modules either. While that's slower, it is the correct thing to do. In a PR you can modify the Puppetfile to point to a branch for complete testing.

wbclark:
@ekohl I updated this to truly skip RPM installation and install foreman-installer and all puppet modules from source using librarian-puppet.

Any idea how I could properly handle the SCL ruby version?

roles/foreman_installer/defaults/main.yml
wbclark: Done.

wbclark:
I also need to figure out what is different w.r.t. answers files vs. RPM installation and ensure the right executable gets called.

wbclark:
With my latest work on this, I have it working to build the installer from source (haven't fully implemented using a PR number yet) and running the installer which fails due to:

```
TASK [foreman_installer : Run installer] *****************************************************************************
fatal: [centos7-foreman-nightly]: FAILED! => changed=true 
  cmd: |-
    scl enable rh-ruby25 -- bundle exec bin/foreman-installer -v --no-colors  --scenario foreman  --foreman-initial-admin-password changeme
  delta: '0:00:12.248301'
  end: '2020-09-17 16:35:46.142440'
  msg: non-zero return code
  rc: 1
  start: '2020-09-17 16:35:33.894139'
  stderr: |-
    Errno::ENOENT: No such file or directory @ rb_sysopen - /etc/foreman-installer/custom-hiera.yaml
```

wbclark:
Now getting through all hooks and reaching puppet apply, where it fails due to:

```
[ERROR 2020-09-17T17:01:29 verbose] Evaluation Error: Error while evaluating a Resource Statement, Unknown resource type: 'augeas' (file: /usr/share/foreman-installer/_build/modules/redis/manifests/ulimit.pp, line: 45, column: 5) on node centos7-foreman-nightly.mercury.example.com
```

wbclark:
Installing puppet-agent fixed the issue: `Unknown resource type: 'augeas'`

wbclark:
Pushed another update. I'm sticking with SCLs but had to apply a few workarounds.

I'm setting up the puppetserver ca before running the installer (because it doesn't work to do it within the SCL)

I had to edit the Gemfile as well to install some additional dependencies.

roles/foreman_installer/defaults/main.yml
```diff
@@ -19,6 +19,9 @@ foreman_installer_module_prs_strategy: "merge"
 # Comma-separated list of "organization/module/branch", e.g. "theforeman/foreman/foo,theforeman/katello/bar"
 foreman_installer_module_branches: []
 
+foreman_installer_from_source: False
+foreman_installer_git_pr_number:
```
ehelms, line 23: Can you use `foreman_installer_git_pr_number` without also using it from source?

wbclark: No. The relevant task only gets included when building the installer from source. I'm planning to add comments which document what these variables do and how to use them.

ehelms: A nice improvement would be that if I set the value of `foreman_installer_git_pr_number` it automatically sets `foreman_installer_from_source` to true.

roles/foreman_installer/tasks/main.yml
```diff
@@ -1,9 +1,17 @@
 ---
+- name: "build installer from source when installer PR is specified"
+  set_fact:
+    foreman_installer_from_source: True
+  when: foreman_installer_git_pr_number is defined and foreman_installer_git_pr_number
```
wbclark, line 5: @ehelms This should provide the functionality you requested.

ehelms:
This can be done via the Packit integration now -- I am opting to close. If you'd still like to perform this using this method please re-open.

PR #1209, opened 2020-09-16
update comment to document we detect AIO Puppet

evgeni:


roles/foreman_installer/tasks/module_prs.yml
```diff
@@ -1,16 +1,10 @@
 ---
-- name: "Determine if Puppet 4 is used"
```
ekohl, line 2: This is really an outdated comment. It should be Puppet AIO, which is the default since Puppet 4. Technically it still works on Debian 10 since it ships Puppet 5.5 and can satisfy the foreman-installer Puppet dependency. I'd prefer to keep this in at least for now, but with an updated comment.

ekohl (approved):

PR #1210, opened 2020-09-16
use pulp 3.6 for katello 3.17

jlsherrill:


jlsherrill:
note, we may want to wait for rc2 to merge this

evgeni:
@jlsherrill is this still needed? I'd assume so.

We have a test that for 3.17 we don't set a pulpcore version:
https://github.com/theforeman/forklift/blob/d55e8ce27144196e365bb7b99067d9c72315a823/roles/forklift_versions/molecule/default/verify.yml#L11-L25

This will need to be adjusted (probably just flipping it to Foreman 2.1/Katello 3.16 is the simplest here)

vagrant/config/versions.yaml
```diff
@@ -79,6 +79,7 @@ installers:
   - foreman: '2.2'
     katello: '3.17'
     pulp: '2.21'
+    pulpcore: '3.6'
```
ekohl, line 82: By now this should be 3.7, right?

ekohl (changes requested):

evgeni:
#1222 should fix that all :)

evgeni:
closing given #1222 is merged.

PR #1216, opened 2020-11-02
Foreman 2.3 updates

pcreech:


vagrant/config/versions.yaml
```diff
@@ -87,6 +87,18 @@ installers:
       - 'debian10'
       - 'ubuntu1804'
 
+  - foreman: '2.3'
+    katello: 'nightly'
```
ekohl, line 91:
```suggestion
    katello: '3.18'
```

vagrant/config/versions.yaml
```diff
@@ -87,6 +87,18 @@ installers:
...
+    pulp: '2.21'
+    pulpcore: '3.6'
```
ekohl, line 93: Not sure, but I think they want 3.7. We can fix that later since nightly is also still on 3.6.

ekohl (changes requested):
Note that our nightly upgrade tests will break after this is merged, until RC1 is available and installable.

vagrant/config/versions.yaml
pcreech: +1 to fix it after nightly.

vagrant/config/versions.yaml
pcreech: should be fixed

ekohl (approved):
Approving but not merging so it can be done in sync with other tasks. Feel free to merge when you think is the right time.

vagrant/config/versions.yaml
ekohl: @jlsherrill FYI

tbrisker:
RC1 is close now, merging

PR #1217, opened 2020-11-03
Add automatic disk resizing when disk_size is specified

ehelms:
If a box is defined with a larger disk size through the use of
disk_size parameter the user then has to manually resize the filesystem.
With the introduction of CentOS 8, which has a base disk size of
10GB within the image this operation will be prevalent for nearly
all Katello production and development deployments through Vagrant.

This adds automatic disk resizing by running a playbook as a
provisioner if the disk_size parameter is defined.

### DIFF
--- a/playbooks/resize_disk.yaml
+++ b/playbooks/resize_disk.yaml
@@ -1,0 +1,37 @@
+---
+- hosts: all
+  become: true
+  tasks:
+    - name: Find root partition
+      shell: df --output=source,fstype / | tail -n 1
+      register: root_partition
+
+    - name: Install cloud-utils-growpart package
+      package:
+        name: cloud-utils-growpart
+      when: ansible_os_family == 'RedHat'
+
+    - name: Install cloud-guest-utils package
+      package:
+        name: cloud-guest-utils
+      when: ansible_os_family == 'Debian'
+
+    - name: Install e2fsprogs package
+      package:
+        name: e2fsprogs
+      when: '"ext4" in root_partition.stdout'
+
+    - name: Extend partition
+      command: growpart /dev/vda {{ root_partition.stdout.split(' ')[0].replace('/dev/vda', '') }}
+      register: growpart_output
+      failed_when:
+        - growpart_output.rc != 0
+        - '"NOCHANGE" not in growpart_output.stdout'
+
+    - name: Grow XFS filesystem
+      command: xfs_growfs -d /
+      when: '"xfs" in root_partition.stdout'
+
+    - name: Grow ext4 filesystem
+      command: resize2fs {{ root_partition.stdout.split(' ')[0] }}
+      when: '"ext4" in root_partition.stdout'
--- a/vagrant/.rubocop.yml
+++ b/vagrant/.rubocop.yml
@@ -12,7 +12,7 @@ Metrics/LineLength:
   Max: 120
 
 Metrics/ClassLength:
-  Max: 300
+  Max: 325
 
 Documentation:
   Enabled: false # don't require documentation
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -82,6 +82,8 @@ module Forklift
                                 "#{box.fetch('name').to_s.tr('.', '-')}.#{domain}"
                               end
 
+        resize_disk(machine) if box.fetch('disk_size', false)
+
         networks = configure_networks(box.fetch('networks', []))
         configure_shell(machine, box)
         configure_ansible(machine, box['ansible'], box['name'])
@@ -96,6 +98,12 @@ module Forklift
         configure_nfs(config, box)
 
         yield machine if block_given?
+      end
+    end
+
+    def resize_disk(machine)
+      machine.vm.provision('disk_resize', type: 'ansible') do |ansible_provisioner|
+        ansible_provisioner.playbook = 'playbooks/resize_disk.yaml'
       end
     end
 


playbooks/resize_disk.yaml
```diff
@@ -0,0 +1,31 @@
...
+      package:
+        name: cloud-utils-growpart
+
+    - name: Install e2fsprogs package
+      package:
+        name: e2fsprogs
+      when: ansible_distribution_major_version == '7'
+
+    - name: Extend partition
+      shell: growpart /dev/vda 1
```
ekohl, line 15: Isn't command safer than shell? I thought ansible-lint complained about that

playbooks/resize_disk.yaml
```diff
@@ -0,0 +1,31 @@
...
+        - '"NOCHANGE" not in output.stdout'
+        - output.rc is not 0
+
+    - name: Grow XFS filesystem
+      shell: xfs_growfs -d /
+      when: ansible_distribution_major_version == '8'
+
+    - name: Grow ext4 filesystem
+      shell: resize2fs /dev/vda1
+      when: ansible_distribution_major_version == '7'
```
ekohl, line 31: Isn't it better to detect the partition type and resize it accordingly?

ekohl (changes requested):
This assumes a CentOS only world.

playbooks/resize_disk.yaml
ehelms: I don't know. I could try that.

playbooks/resize_disk.yaml
ehelms: Generally yes, but depends on the use case. I had to go with shell here due to command erroring out. I'd have to dig up the error.

ehelms:
This is now a bit more dynamic for more use cases and has been tested on CentOS 7 and 8, Ubuntu 1604 and Debian 10 images.

playbooks/foreman_repos.yml
```diff
@@ -0,0 +1,6 @@
+---
```
ekohl, line 1: This looks unrelated

playbooks/resize_disk.yaml
```diff
@@ -0,0 +1,41 @@
+---
+- hosts: all
+  become: true
+  tasks:
+    - name: Find root partition
+      shell: df -hT | grep /$
```
ekohl, line 6:
```suggestion
      shell: df --output=source,fstype / | tail -n 1
```

playbooks/resize_disk.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Extend partition
+      command: growpart /dev/vda {{ root_partition.stdout.split(' ')[0].replace('/dev/vda', '') }}
+      register: output
+      ignore_errors: true
+
+    - name: Fail if growpart fails, do not fail if NOCHANGE
+      fail:
+      when:
+        - '"NOCHANGE" not in output.stdout'
+        - output.rc != 0
```
ekohl, line 33: Does't this make the `vagrant provision` command fail on runs after the first one?

ekohl (changes requested):

playbooks/resize_disk.yaml
ehelms: Nope. This is an `and` condition. So if it reports a non-zero return code but does not contain NOCHANGE in the output, then the error is likely a real error and we should fail. For whatever reason, NOCHANGE causes a return code of `1` and from all my testing the same is true if oyu attempt it with `--dry-run`.

playbooks/resize_disk.yaml
```diff
@@ -0,0 +1,41 @@
...
+        name: cloud-guest-utils
+      when: ansible_os_family == 'Debian'
+
+    - name: Install e2fsprogs package
+      package:
+        name: e2fsprogs
+      when: '"ext4" in root_partition.stdout'
+
+    - name: Extend partition
+      command: growpart /dev/vda {{ root_partition.stdout.split(' ')[0].replace('/dev/vda', '') }}
```
ekohl, line 25: I am wondering if this is always `vda`. `vda` is naming for virtio but on virtualbox it may fail if they use `sda`. Not sure if we have those, but worth considering.

ekohl (commented):
I'd be OK with trying this.

playbooks/resize_disk.yaml
ehelms: Can cross that bridge if someone gets there.

playbooks/resize_disk.yaml
```diff
@@ -0,0 +1,41 @@
...
+      register: output
+      ignore_errors: true
+
+    - name: Fail if growpart fails, do not fail if NOCHANGE
+      fail:
+      when:
+        - '"NOCHANGE" not in output.stdout'
+        - output.rc != 0
```
evgeni, line 33: Instead of using `fail` (without any message) here, wouldn't it be possible to make a more correct [`failed_when` in the `command` step](https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html#defining-failure) above?

something like:

```yaml
    - name: Extend partition
      command: growpart /dev/vda {{ root_partition.stdout.split(' ')[0].replace('/dev/vda', '') }}
      register: growpart_output
      failed_when:
        - '"NOCHANGE" not in growpart_output.stdout'
        - growpart_output.rc != 0
```

BTW, are the extra quotes around the `… not in …` required?

ehelms: Yes, the quotes appear to be needed.

evgeni: :crying_cat_face:

evgeni (approved):

ekohl (approved):

PR #1218, opened 2020-11-03
Add centos8-devel box for Katello devel environments on EL8

ehelms:
Requires https://github.com/theforeman/puppet-katello_devel/pull/245

This also requires https://github.com/theforeman/forklift/pull/1217 as the base CentOS 8 box is only 10GB and the devel install will run out of disk space before completing.

### DIFF
--- a/roles/foreman_installer_devel_scenario/tasks/main.yml
+++ b/roles/foreman_installer_devel_scenario/tasks/main.yml
@@ -9,12 +9,18 @@
     state: directory
     path: /etc/foreman-installer/scenarios.d
 
-- name: 'Copy scenario files'
+- name: 'Copy scenario file'
   copy:
     src: "{{ role_path }}/files/{{ item }}"
     dest: /etc/foreman-installer/scenarios.d
   with_items:
     - katello-devel.yaml
+
+- name: 'Copy answers file'
+  template:
+    src: "{{ role_path }}/templates/{{ item }}"
+    dest: /etc/foreman-installer/scenarios.d/{{ item }}
+  with_items:
     - katello-devel-answers.yaml
 
 - name: 'Remove current puppet-katello_devel'
--- a/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
+++ b/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
@@ -6,17 +6,11 @@ certs:
 katello_devel:
   deployment_dir: /home/vagrant
   user: vagrant
-  db_type: postgres
-  rvm: true
 foreman_proxy_content:
-  pulp_master: true
-  qpid_router_broker_addr: localhost
+  proxy_pulp_yum_to_pulpcore: true
 foreman_proxy:
-  custom_repo: true
   http: true
   ssl_port: '9090'
-  templates: false
-  tftp: false
   ssl_ca: /etc/foreman-proxy/ssl_ca.pem
   ssl_cert: /etc/foreman-proxy/ssl_cert.pem
   ssl_key: /etc/foreman-proxy/ssl_key.pem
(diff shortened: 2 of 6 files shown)


vagrant/boxes.d/99-local.yaml.example
```diff
@@ -10,6 +10,20 @@ centos7-katello-devel:
...
+  primary: true
+  box: centos8
+  disk_size: 40
+  ansible:
+    playbook: 'playbooks/katello_devel.yml'
+    group: 'devel'
+    variables:
+      ssh_forward_agent: true
+      foreman_devel_github_push_ssh: True
+      katello_devel_github_username: ehelms
```
ekohl, line 23: Can we use a more generic example?
```suggestion
      katello_devel_github_username: <REPLACE ME>
```

vagrant/.bundle/config
```diff
@@ -0,0 +1,2 @@
+---
+BUNDLE_PATH: "/home/ehelms/.rubygems/"
```
ekohl, line 2: This looks unrelated

roles/katello_devel/meta/main.yml
```diff
@@ -6,15 +6,19 @@ dependencies:
     foreman_server_repositories_katello: true
     foreman_server_repositories_foreman_client: true
   - role: ruby_scl
+    when: ansible_distribution_major_version == "7"
   - role: nodejs_scl
+    when: ansible_distribution_major_version == "7"
   - role: postgresql_scl
+    when: ansible_distribution_major_version == "7"
```
evgeni, line 13: should this instead become a `devel_scls` meta-role? (not blocking this PR)

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -10,6 +10,20 @@ centos7-katello-devel:
       foreman_devel_github_push_ssh: True
       katello_devel_github_username: <REPLACE ME>
 
+centos8-devel:
```
evgeni, line 13: technically `centos8-katello-devel` right, to match the other box names?

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -10,6 +10,20 @@ centos7-katello-devel:
...
+  primary: true
+  box: centos8
+  disk_size: 40
```
evgeni, line 16: should this be instead set on the `centos8` box directly, as it will also be required for the pipeline boxes?

roles/katello_devel/meta/main.yml
ehelms: Maybe? I do not see the need for that additional layer of indirection presently.

evgeni (approved):

roles/foreman_installer_devel_scenario/tasks/main.yml
```diff
@@ -15,6 +15,12 @@
     dest: /etc/foreman-installer/scenarios.d
   with_items:
     - katello-devel.yaml
```
ekohl, line 17: This is now just a single file. Mind changing the name to `Copy scenario file`?

roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
```diff
@@ -6,11 +6,8 @@ certs:
...
   deployment_dir: /home/vagrant
   user: vagrant
-  db_type: postgres
-  rvm: true
 foreman_proxy_content:
-  pulp_master: true
-  qpid_router_broker_addr: localhost
+  proxy_pulp_yum_to_pulpcore: true
 foreman_proxy:
   custom_repo: true
```
ekohl, line 12: If you want to clean up: this parameter is gone from the module. tftp is also off by default nowadays, same for templates.

ekohl (approved):

ehelms:
Thanks for approvals. I will wait to merge this till https://github.com/theforeman/puppet-katello_devel/pull/245 is merged since it will give a false sense of availability until then.

PR #1221, opened 2020-11-13
Identify content proxy by hardcoded ID

ehelms:
Will need:

 * https://github.com/Katello/katello/pull/9037
 * SELinux fixes
 * https://github.com/theforeman/puppet-foreman_proxy_content/pull/293
 * https://github.com/theforeman/foreman-installer/pull/611

I will update with PRs once they are available.

### DIFF
--- a/bats/fb-katello-proxy.bats
+++ b/bats/fb-katello-proxy.bats
@@ -9,9 +9,9 @@ load fixtures/content
 
 setup() {
   tSetOSVersion
-  PROXY_INFO=$(hammer --output json proxy list --search "feature = \"Pulp Node\"")
-  PROXY_ID=$(echo $PROXY_INFO | ruby -e "require 'json'; puts JSON.load(ARGF.read).first['Id']")
-  PROXY_HOSTNAME=$(echo $PROXY_INFO | ruby -e "require 'json'; puts JSON.load(ARGF.read).first['Name']")
+  PROXY_ID=2
+  PROXY_INFO=$(hammer --output json proxy show --id $PROXY_ID)
+  PROXY_HOSTNAME=$(echo $PROXY_INFO | ruby -e "require 'json'; puts JSON.load(ARGF.read)['Name']")
 }
 
 @test "proxy is registered" {


bats/fb-proxy.bats
```diff
@@ -9,9 +9,9 @@ load fixtures/content
 
 setup() {
   tSetOSVersion
-  PROXY_INFO=$(hammer --output json proxy list --search "feature = \"Pulp Node\"")
-  PROXY_ID=$(echo $PROXY_INFO | ruby -e "require 'json'; puts JSON.load(ARGF.read).first['Id']")
-  PROXY_HOSTNAME=$(echo $PROXY_INFO | ruby -e "require 'json'; puts JSON.load(ARGF.read).first['Name']")
+  PROXY_ID=2
+  PROXY_INFO=$(hammer --output json proxy show --id 2)
```
ekohl, line 13: Shouldn't this at least use `$PROXY_ID`
```suggestion
  PROXY_INFO=$(hammer --output json proxy show --id $PROXY_ID)
```

bats/fb-proxy.bats
```diff
@@ -9,9 +9,9 @@ load fixtures/content
 
 setup() {
   tSetOSVersion
-  PROXY_INFO=$(hammer --output json proxy list --search "feature = \"Pulp Node\"")
-  PROXY_ID=$(echo $PROXY_INFO | ruby -e "require 'json'; puts JSON.load(ARGF.read).first['Id']")
-  PROXY_HOSTNAME=$(echo $PROXY_INFO | ruby -e "require 'json'; puts JSON.load(ARGF.read).first['Name']")
+  PROXY_ID=2
```
ekohl, line 12: Is there a different way to determine which proxies are mirrors? Is this something we should take up with Hammer? Can it even be read from the API? I'm not sure if we implemented capabilities in the proxy API.

ehelms: Exactly. Either Hammer or the API needs to be updated to respond with capabilities. I'll open an issue to get that fixed and make this better in the future. For now, this is the best work around I could come up with.

ekohl: Please open that issue and add it as a TODO in this file.

An alternative approach is to iterate over the proxies with the Pulp 3 feature and exclude the one with the same hostname as Foreman itself. That's not 100% guaranteed to be correct, but in most deployment it's true and avoids hardcoding the ID.

ekohl: Capabilities are present in the API. If we switch to smoker tests and use apypie, it shouldn't be an issue.

ekohl (approved):
This info should be available via the API with capabilities so once we move it to smoker, it should be available. Since we plan on moving away from bats altogether, I'm fine with this workaround.

PR #1223, opened 2020-11-19
clean subscription manager data after content tests are done

evgeni:
otherwise dnf on EL8 chokes on the unavailable repository when tries to install sos in the collect debug step

### DIFF
--- a/bats/fb-content-katello.bats
+++ b/bats/fb-content-katello.bats
@@ -155,12 +155,8 @@ setup() {
     rpm -e `rpm -qf /etc/rhsm/ca/candlepin-local.pem`
   fi
 
-  run subscription-manager unregister
-  echo "rc=${status}"
-  echo "${output}"
-  run subscription-manager clean
-  echo "rc=${status}"
-  echo "${output}"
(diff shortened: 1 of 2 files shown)


evgeni:
cc @ianballou @ehelms

bats/fb-content-katello.bats
```diff
@@ -491,3 +483,7 @@ setup() {
   hammer docker manifest list --content-view-version-id=$cvv_id --fields="schema version,digest,tags" \
     --order='tag' | grep 'sha256:a6ecbb1553353a08936f50c275b010388ed1bd6d9d84743c7e8e7468e2acd82e'
 }
+
+@test "cleanup subscription-manager after content tests" {
```
ekohl, line 487: Sadly we can't rely on it (bats in EPEL7 doesn't support it), but there is a `teardown_file` which would be nicer for this.

ekohl (approved):

bats/fb-content-katello.bats
evgeni: TIL, one day… :)

PR #1225, opened 2020-11-25
use qpid copr also on el7

evgeni:


roles/katello_repositories/tasks/main.yml
```diff
@@ -8,6 +8,5 @@
   args:
     warn: false # dnf module can't handle modules
 
-  # Required for el6 clients
 - include_tasks: qpid.yml
-  when: ansible_distribution_major_version == "6"
+  when: ansible_distribution_major_version in ["6", "7"]
```
ekohl, line 12: That copr is only published for EL7 (which means this code path was broken before)

ekohl (changes requested):

roles/katello_repositories/tasks/main.yml
evgeni: heh, indeed.

evgeni: updated

roles/katello_repositories/tasks/main.yml
```diff
@@ -8,6 +8,5 @@
   args:
     warn: false # dnf module can't handle modules
 
-  # Required for el6 clients
 - include_tasks: qpid.yml
-  when: ansible_distribution_major_version == "6"
+  when: ansible_distribution_major_version == "7"
```
ekohl, line 12: I think this should only be in `staging_repos.yml`. Release repos use the RPM.

ekohl (changes requested):

roles/katello_repositories/tasks/main.yml
evgeni: This is true, but that'd mean that *all* pipelines are red until we've updated `katello-repos` there.

I'd prefer to keep it here for now, and move to staging when updates are pushed out.

ekohl: Won't it break on nightly because there are 2 definitions on nightly? I guess fresh install will work if `yum_repository` is smart enough to look through all files but upgrades may suffer from it.

evgeni: The module isn't that smart, but I'll give it a try what yum thinks about that.

evgeni: yum yields a warning `Repository qpid is listed more than once in the configuration`, but works fine otherwise if `katello-repos` deploys a `qpid` repo in `/etc/yum.repos.d/katello.repo` and this PR one in `/etc/yum.repos.d/qpid.repo`

evgeni: we could do:

```yaml
when:
  - ansible_distribution_major_version == "7"
  - katello_repositories_environment == "staging" or (katello_repositories_version != "nightly" and katello_repositories_version is version('3.18', '<='))
```

and then tighten the version whenever we have a release containing the fixed `katello-repos`?

ekohl: That sounds good.

evgeni: Thanks, updated

ekohl (approved):

PR #1227, opened 2020-11-25
Client tests on smoker

ehelms:
I am only currently applying this change to the install pipeline for feedback. The idea here is to move the client testing bats tests to run on the smoker box. Thereby creating an isolated client from which to test. This requires changes to the client bats tests that are reflected here. After this change, I am planning to explore running these bats tests each from a container running on the smoker box, one for each OS we care about testing.

### DIFF
--- a/roles/hammer_credentials/tasks/main.yml
+++ b/roles/hammer_credentials/tasks/main.yml
@@ -1,4 +1,8 @@
 ---
+- name: Install hammer
+  package:
+    name: tfm-rubygem-hammer_cli_katello
+
 - name: 'Add hammer config directory'
   file:
     path: ~/.hammer
--- a/bats/fb-katello-client.bats
+++ b/bats/fb-katello-client.bats
@@ -30,7 +30,7 @@ load fixtures/content
   run yum erase -y 'katello-ca-consumer-*'
   echo "rc=${status}"
   echo "${output}"
-  run rpm -Uvh http://localhost/pub/katello-ca-consumer-latest.noarch.rpm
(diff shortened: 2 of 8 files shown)


pipelines/install/06-smoker.yaml
```diff
@@ -1,5 +1,6 @@
 ---
 - name: run smoker
+  become: true
```
ekohl, line 3: The `smoker` role doesn't work with `become: true` and there is no `become: false`. That's why you see `become: true` on the roles that should `become: true` at the `roles` level, not at the top level.

roles/hammer_credentials/tasks/main.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- name: Install hammer
+  package:
+    name: tfm-rubygem-hammer_cli_katello
```
ekohl, line 4: I was under the impression we had some good provides to make it easier (like we have `foreman-plugin-$plugin` instead of `tfm-rubygem-foreman_$plugin`) but looks like we don't. Perhaps good to introduce now, even though we can't rely on it yet.

pipelines/install/06-smoker.yaml
evgeni: I think `become: false` exists.

ekohl: When I tried it, it didn't work. Perhaps that's been fixed since but that's the reason it's this way today.

Note that the reason it doesn't run is that chromium doesn't want to run as root. Rather than override it, I let smoker run as the normal user.

PR #1228, opened 2020-11-27
drop support for old (and unsupported) Foreman/Katello versions

evgeni:


roles/foreman_installer/tasks/install.yml
```diff
@@ -10,18 +10,6 @@
   set_fact:
     foreman_installer_options_joined: "{{ foreman_installer_options | join(' ') }} {{ foreman_installer_options_internal_use_only | join(' ') }}"
 
-- name: 'Replace certs-tar and --inital-admin parms with legacy options'
-  set_fact:
-    foreman_installer_options_joined: "{{ foreman_installer_options_joined | replace(item.new, item.old) }}"
-  when: foreman_installer_version_file['content'] | b64decode is version('1.21', '<')
```
evgeni, line 16: this drops support for Satellite 6.5 and older, but given that is EOL too…

ekohl (approved):

PR #1231, opened 2020-12-02
drop old and unsupported releases

evgeni:
1.24+ is kept for upgrade testing


tbrisker (approved):
about time, though i'm not sure what the molecule failure is about

evgeni:
the molecule failure was due to me using an old katello version in an molecule test -- updated the test :)

PR #1238, opened 2020-12-10
Set subuid and subgid to a large range for podman rootless smoker con…

ehelms:
…tainers

### DIFF
--- a/roles/podman/tasks/main.yml
+++ b/roles/podman/tasks/main.yml
@@ -1,0 +1,18 @@
+- name: 'Install podman'
+  become: true
+  package:
+    name: 'podman'
+
+- name: Set max user namespaces
+  become: true
+  sysctl:
+    name: user.max_user_namespaces
+    value: 100000
+
+- name: Set subuid range for user
+  become: true
+  command: usermod --add-subuids 10000-75535 {{ ansible_user }}
+
+- name: Set subgid range for user
+  become: true
+  command: usermod --add-subgids 10000-75535 {{ ansible_user }}
--- a/pipelines/install/06-smoker.yaml
+++ b/pipelines/install/06-smoker.yaml
@@ -13,4 +13,5 @@
       become: true
     - role: epel_repositories
       become: true
+    - podman
     - smoker
--- a/pipelines/upgrade/10-smoker.yaml
+++ b/pipelines/upgrade/10-smoker.yaml
@@ -13,4 +13,5 @@
       become: true
     - role: epel_repositories
       become: true
+    - podman
     - smoker
--- a/playbooks/smoker.yml
+++ b/playbooks/smoker.yml
@@ -4,4 +4,5 @@
       become: true
     - role: etc_hosts
       become: true
+    - podman
     - smoker


roles/smoker/tasks/main.yml
```diff
@@ -4,6 +4,23 @@
...
+  package:
+    name: 'podman'
+
+- name: Set max user namespaces
+  become: true
+  shell: echo 100000 > /proc/sys/user/max_user_namespaces
+
+- name: Set subuid range for user
+  become: true
+  command: usermod --add-subuids 10000-75535 {{ ansible_user }}
```
evgeni, line 18: it's super confusing that this is a range, but `/etc/subuid` contains `start:length` instead

ehelms: That's how the command works (with the `-`)

roles/smoker/tasks/main.yml
```diff
@@ -4,6 +4,23 @@
...
     name: '{{ smoker_browser_packages }}'
 
+- name: 'Install podman'
+  become: true
+  package:
+    name: 'podman'
+
+- name: Set max user namespaces
+  become: true
+  shell: echo 100000 > /proc/sys/user/max_user_namespaces
```
ekohl, line 14: Isn't this just `user.max_user_namespaces` that can be set via sysctl?
```suggestion
  sysctl:
    name: user.max_user_namespaces
    value: 100000
```

roles/smoker/tasks/main.yml
evgeni: Yeah, I realize that, just was commenting for whoever will also look at this and wonder why it does not match their memories ;)

roles/smoker/tasks/main.yml
```diff
@@ -4,6 +4,25 @@
...
+  become: true
+  sysctl:
+    name: user.max_user_namespaces
+    value: 100000
+
+- name: Set subuid range for user
+  become: true
+  command: usermod --add-subuids 10000-75535 {{ ansible_user }}
+
+- name: Set subuid range for user
```
ekohl, line 22: This doesn't match the content
```suggestion
- name: Set subgid range for user
```

ekohl (commented):
Can't you combine both usermod commands into a single command?

I also wonder if this should really be a podman role that should be included instead.

evgeni:
I agree that this is probably better suited in an own role. No hard feelings about one or two usermod calls.

roles/podman/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+  become: true
+  sysctl:
+    name: user.max_user_namespaces
+    value: 100000
+
+- name: Set subuid range for user
+  become: true
+  command: usermod --add-subuids 10000-75535 {{ ansible_user }}
+
+- name: Set subuid range for user
```
evgeni, line 16:
```suggestion
- name: Set subgid range for user
```

evgeni (approved):

ekohl (approved):
Feel free to merge when you want to start deploying it.

ehelms:
I'm going to merge it ahead as there will be a follow on to layer in the requirements to run the client tests (e.g. setting the set of clients and building the images)

PR #1240, opened 2020-12-10
Do not run Puppet tests if Pulp 3 only for Katello 4

ehelms:
We have a period of time where Pulp 2 is around for katello-agent support but not content so we need to skip Puppet tests based n Katello version and not the presence of Pulp 2.

### DIFF
--- a/bats/foreman_helper.bash
+++ b/bats/foreman_helper.bash
@@ -18,13 +18,30 @@ tForemanVersion() {
   ) | cut -d. -f1-2
 }
 
+tKatelloVersion() {
+  (
+    if tPackageExists tfm-rubygem-katello; then
+      tPackageVersion tfm-rubygem-katello
+    elif tPackageExists rubygem-katello; then
+      tPackageVersion rubygem-katello
+    fi
+  ) | cut -d. -f1-2
+}
+
+tSkipIfPulp3Only() {
+  KATELLO_VERSION=$(tKatelloVersion)
+  if [[ $KATELLO_VERSION == 4.* ]]; then
+    skip "${1} is not available in scenarios with only Pulp 3"
+  fi
+}
+
 tIsPulp2() {
   tPackageExists pulp-server
 }
 
 tSkipIfNoPulp2() {
   if ! tIsPulp2; then
-   skip "${1} is not available in scenarios without Pulp 2"
+    skip "${1} is not available in scenarios without Pulp 2"
   fi
 }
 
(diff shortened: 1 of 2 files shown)


bats/foreman_helper.bash
```diff
@@ -18,6 +18,23 @@ tForemanVersion() {
   ) | cut -d. -f1-2
 }
 
+tKatelloVersion() {
+  (
+    if tPackageExists tfm-rubygem-katello; then
+      tPackageVersion tfm-rubygem-katello
+    elif tPackageExists rubygem-katello; then
+      tPackageVersion rubygem-katello
+    fi
```
evgeni, line 27: any reason not to check the `katello` metapackage instead?

ehelms: The rubygem package was more reliable for the truth, brace yourself -- https://fedorapeople.org/groups/katello/releases/yum/nightly/katello/el7/x86_64/katello-3.18.0-0.4.master.el7.noarch.rpm

ekohl: I wouldn't mind if we eventually get rid of the `katello` metapackage or at least make it fully optional. Because of that, :+1: for looking at the gem. Alternatively, I'd be fine with deriving it from the Foreman version and see if it's at least 2.4 to avoid testing for yet another package version.

ekohl (commented):
Just wondering out loud: is it possible the get this info (is Puppet content supported) from the API somehow? If not, we should file an issue to get that supported.

ehelms:
The smart proxy reports that information, the status / ping API for Foreman does not.

evgeni:
Maybe with https://theforeman.org/plugins/katello/3.17/api/apidoc/v2/repositories/repository_types.html

bats/foreman_helper.bash
evgeni: Bleh @ Katello.rpm

Ack to ewouds idea

ehelms: I thought of trying to use the Foreman version, but given this is a Katello feature, and we can assume lockstep, the past has proven it will not always be true. I've written the code, it works, and I'd prefer to stick with the Katello version here to be as specific as possible.

evgeni: wfm

evgeni (approved):

evgeni:
I'm fine with the current state, we can explore API in the future?

ekohl (approved):
Given we're trying to get everything in Smoker, let's do a clean implementation there.

PR #1241, opened 2020-12-11
Shadow-utils package is needed for --add-subuids

ehelms:


roles/podman/tasks/main.yml
```diff
@@ -3,6 +3,11 @@
   package:
     name: 'podman'
```
ekohl, line 4: Does it make sense to make this a list
```suggestion
    name:
      - 'podman'
      - 'shadow-utils'
```

ehelms: I like keeping them separate.

evgeni (approved):

PR #1245, opened 2020-12-14
Add role for setup of kerberos

ezr-ondrej:
Role for setting up of kerberos with an existing domain AD controller.
Documentation for such setup in easiest local scenario.

Comment:
Mostly for my future reference I guess.

I was thinking about putting the documentation in `docs/advanced` and just linking it, but unsure what's better.

I can actually remove the playbook as it is quite simple, move the whole documentation to the role `README.md` and just link there from development.


docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
+This list should be organized by foreman feature.
+
+* [Kerberos](#foreman-with-ad)
+
+### Foreman with AD
+
+#### Prepare AD controller
+
+You need the AD controller with domain controller you want to test this against.
+Easiest is to install Windows server (ideally core to safe resources, but its harder to install AD and setup DNS).
```
ekohl, line 370:
```suggestion
Easiest is to install Windows server (ideally core to save resources, but its harder to install AD and setup DNS).
```

docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
 5. From the smoker box, run tests as the vagrant user using the alias or running pytest commands manually. To change the testing options, please see [the smoker documentation](https://github.com/theforeman/smoker) and modify the alias or manually run pytest commands as necessary.
 
+## Advanced features
+
+These are playbooks, roles and variables provided to use additional funcionalities to test advanced foreman features.
+This list should be organized by foreman feature.
+
+* [Kerberos](#foreman-with-ad)
+
+### Foreman with AD
```
ekohl, line 365:
```suggestion
### Foreman with Active Directory
```

docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
+
+You need the AD controller with domain controller you want to test this against.
+Easiest is to install Windows server (ideally core to safe resources, but its harder to install AD and setup DNS).
+This controller doesn't need to be searchable in your environmnet, we will point libvirt there.
+
+I will be using domain ```example.com``` and expect your AD controller runs on IP ```192.168.1.8```.
+
+#### Setup libvirt
+
+Here we will edit the libvirt network, so its dnsmasq point the domain request to our controller.
```
ekohl, line 377:
```suggestion
Here we will edit the libvirt network, so its dnsmasq points the domain request to our controller.
```

docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
+    <dnsmasq:option value="server=/example.com/192.168.1.8"/>
+  </dnsmasq:options>
+</network>
+
+```
+
+#### Setup forklift machine
+
+Then add the machine and specify the domain, playbook and password for the join user.
+I'm using short name as longer names tend to have issues with NETBIOS names that end up shortened, and hostname != NETBIOS name could cause issues.
```
ekohl, line 396:
```suggestion
I'm using short names as longer names tend to have issues with NETBIOS names that end up shortened, and hostname != NETBIOS name could cause issues.
```

docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
+If you choose longer name you're on your own, contributions with fixes welcome! :)
+
+```yaml
+# boxes.d/99-local.yaml
+
+foreman-ad:
+  box: centos7
+  memory: 4098
```
ekohl, line 404:
```suggestion
  memory: 4096
```

docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
+#### Setup forklift machine
+
+Then add the machine and specify the domain, playbook and password for the join user.
+I'm using short name as longer names tend to have issues with NETBIOS names that end up shortened, and hostname != NETBIOS name could cause issues.
+If you choose longer name you're on your own, contributions with fixes welcome! :)
+
+```yaml
+# boxes.d/99-local.yaml
+
+foreman-ad:
```
ekohl, line 402: Perhaps this should also be in the example boxes file?

roles/foreman_installer/tasks/module_local.yml
```diff
@@ -0,0 +1,35 @@
+---
```
ekohl, line 1: Why is this file needed?

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,116 @@
...
+  loop:
+    - sssd
+    - adcli
+    - realmd
+    - ipa-python-compat
+    - krb5-workstation
+    - samba-common-tools
+    # Irrelevant, missing installer support for gssproxy
+    - gssproxy
+    - nfs-utils
```
ekohl, line 34: I think `name` also accepts a list and that's faster than looping.

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,116 @@
...
+  become: yes
+  package: name=pexpect state=latest
+  when: ansible_distribution != "CentOS" or ansible_distribution_major_version >= '8'
+
+- name: Install pip to get supported pexpect version
+  become: yes
+  package: "name={{ item }} state=present"
+  loop:
+    - epel-release
+    - python-pip
```
ekohl, line 12: I think `name` accepts a list as well and is faster than looping.

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,116 @@
+---
+- name: Ensure pexpect to answer password
+  become: yes
+  package: name=pexpect state=latest
```
ekohl, line 4: Writing it out is generally cleaner (repeated below, not going to repeat the comment)
```suggestion
  package:
    name: pexpect
    state: latest
```

roles/realm_join/files/httpd-gssproxy.service
```diff
@@ -0,0 +1,3 @@
+.include /lib/systemd/system/httpd.service
```
ekohl, line 1: `.include` isn't needed if you create it as a drop in. That means storing it as `/etc/systemd/system/httpd.service.d/gss-use-proxy.conf` (or similar). On newer systemd versions (included in EL8) the `.include` statement generates a deprecation warning.

ekohl (changes requested):
Thanks! Looks like a nice thing to have.

roles/realm_join/files/httpd-gssproxy.service
ezr-ondrej: I'm trying to follow the Satellite manual as strictly as possible - that is why I'm having the `gssproxy` steps even though those are not used.
Noted for the manual improvements tho! Thanks!

roles/foreman_installer/tasks/module_local.yml
ezr-ondrej: :shy:

roles/realm_join/files/httpd-gssproxy.service
ekohl: We should file a docs bug then.

docs/development.md
ezr-ondrej: I don't think so, this is very obscure setup for most developers (you need the AD controller first), so it is hardly something most users would use, so I think it should not be in examples of everyuser (and thus being offered in `vagrant up` autocomplete if user doesn't delete it)

docs/development.md
ezr-ondrej: I believe my version is correct.

docs/development.md
ekohl: https://github.com/theforeman/forklift/blob/master/vagrant/boxes.d/99-local.yaml.example isn't loaded by default so it wouldn't show up in the autocomplete after a fresh clone. However, I'm fine either way since it is indeed not something that works on its own.

docs/development.md
ekohl: I think it's either `a short name` or `short names`, but since we're both not native speakers we may want to ask another reviewer :)

docs/development.md
ezr-ondrej: I know, but first thing most people will do is probably `cp 99-local.yaml.example 99-local.yaml` to try those examples out :shrug:
At least that's what I've done the first time.

roles/realm_join/tasks/realm_join_ad.yml
ezr-ondrej: :+1: yes it indeed does!

docs/development.md
ezr-ondrej: Right! I'm missing the `a`. I was just refering to singular being used correctly there :)

roles/realm_join/files/httpd-gssproxy.service
ezr-ondrej: I'll followup with the improvements of the feature, docs improvement is part of it, I will take this into account :+1:

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,120 @@
...
+
+- name: Create httpd conf dir to put keytab [Hack - remove once using gssproxy]
+  become: yes
+  file:
+    path: /etc/httpd/conf
+    state: directory
+    mode: '0775'
+    recurse: yes
+
+- name: Check wheter the keytab exists
```
ekohl, line 68:
```suggestion
- name: Check whether the keytab exists
```

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,120 @@
+---
+- name: Ensure pexpect to answer password
+  become: yes
+  package:
+    name: pexpect
+    state: latest
+  when: ansible_distribution != "CentOS" or ansible_distribution_major_version >= '8'
```
ekohl, line 7: So on RHEL 7 pexpect is present?

docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
+This list should be organized by foreman feature.
+
+* [Kerberos](#foreman-with-active-directory)
+
+### Foreman with Active Directory
+
+#### Prepare AD controller
+
+You need the AD controller with domain controller you want to test this against.
+Easiest is to install Windows server (ideally core to save resources, but its harder to install AD and setup DNS).
```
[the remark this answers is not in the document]
ezr-ondrej: Well, that would be indead cool, but we could just have freeIPA kerberos :)

roles/realm_join/tasks/realm_join_ad.yml
ezr-ondrej: I'd expect no ;) ... I'll add RHEL :)

docs/development.md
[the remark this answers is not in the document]
ezr-ondrej: But let's first get it in without it... if there I find out people struggle to get some reviews done, I'll try to come up with the REALM installation here as well.

roles/realm_join/files/httpd-gssproxy.service
ekohl: I still don't like implementing this because it's not the way it's supposed to be used. By placing the file in `/etc/systemd/system/httpd.service` it means you need to have a migration in the future. Probably not a huge issue for most forklift deployments, but I did want to mention it.

ekohl:
@evgeni could you take a look at this and see if I've missed anything?

docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
+This controller doesn't need to be searchable in your environmnet, we will point libvirt there.
+
+I will be using the domain ```example.com``` and expect your AD controller runs on IP ```192.168.1.8```.
```
evgeni, line 373:
```suggestion
I will be using the domain ```example.com``` and expect your AD controller runs on IP ```192.0.2.8```.
```

[rfc5735](https://tools.ietf.org/html/rfc5735) ;)

evgeni: And I think you want singe backticks here, not three?

docs/development.md
```diff
@@ -348,9 +349,82 @@ In the vagrant box, the dynflow repository is cloned to `/home/vagrant/dynflow`.
...
+Provision it
+
+```vagrant up foreman-ad```
+
+Test it
+
+```vagrant ssh foreman-ad```
+```kinit test-domain-user``` # initialize Kerberos with an existing user in the AD domain
+```klist``` # verify the above went smooth
+```curl -k -u : --negotiate https://foreman-ad.example.com/users/extlogin``` # substitute for your fqdn
```
evgeni, line 425:
```suggestion
```console
$ vagrant ssh foreman-ad
$ kinit test-domain-user # initialize Kerberos with an existing user in the AD domain
$ klist # verify the above went smooth
$ curl -k -u : --negotiate https://foreman-ad.example.com/users/extlogin # substitute for your fqdn
```

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,121 @@
...
+  package:
+    name:
+      - epel-release
+      - python-pip
+    state: present
+  when: ansible_os_family == "RedHat" and ansible_distribution_major_version < '8'
+
+- name: Ensure pexpect to answer password from pip
+  become: yes
+  pip: name=pexpect
```
evgeni, line 21:
```suggestion
  pip:
    name: pexpect
```

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,121 @@
...
+  when: ansible_os_family == "RedHat" and ansible_distribution_major_version < '8'
```
evgeni, line 22:
```suggestion
  when:
    - ansible_os_family == "RedHat"
    - ansible_distribution_major_version is version('8', '<')
```

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,121 @@
...
+  when: ansible_os_family != "RedHat" or ansible_distribution_major_version >= '8'
+
+- name: Install pip to get supported pexpect version
+  become: yes
+  package:
+    name:
+      - epel-release
+      - python-pip
+    state: present
+  when: ansible_os_family == "RedHat" and ansible_distribution_major_version < '8'
```
evgeni, line 17:
```suggestion
  when:
    - ansible_os_family == "RedHat"
    - ansible_distribution_major_version is version('8', '<')
```

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,121 @@
+---
+- name: Ensure pexpect to answer password
+  become: yes
+  package:
+    name: pexpect
+    state: latest
+  # on Fedora this will be True due to the major version. Not nice, but works.
+  when: ansible_os_family != "RedHat" or ansible_distribution_major_version >= '8'
```
evgeni, line 8:
```suggestion
  when: ansible_os_family != "RedHat" or ansible_distribution_major_version is version('8', '>')
```

But there is no `pexpect` on my Fedora 33, only `python3-pexpect`. And neither is there such a package on Debian ;-)

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,121 @@
...
+
+- name: Get the keytab
+  become: yes
+  environment:
+    KRB5_KTNAME: 'FILE:/etc/httpd/conf/http.keytab'
+  expect:
+    command: "net ads keytab add HTTP -U {{ foreman_realm_directory_admin_name }} -d3 -s /etc/net-keytab.conf"
+    responses:
+      (?i)password: '{{ foreman_realm_directory_admin_password }}'
+  when: not keytab_result.stat.exists
```
evgeni, line 82: instead of `when`, you probably can have something like `creates: /etc/httpd/conf/http.keytab` (and then the step before can be remoted)

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,121 @@
+---
+- name: Ensure pexpect to answer password
+  become: yes
+  package:
+    name: pexpect
+    state: latest
```
evgeni, line 6: Does it really need to be `latest` (aka: trigger upgrade on re-run?)

evgeni (changes requested):

docs/development.md
ezr-ondrej: Yeah, I guess I can, no idea why I used three, it was bit late and probably I went with three is more then one :thinking: :shrug:

roles/realm_join/tasks/realm_join_ad.yml
ezr-ondrej: Probably not, but quite a new version is necessary, so I didn't want to go with just simple `present` in case there is old version, but no idea if that can even happen on 8 :thinking:

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,121 @@
+---
+- name: Ensure pexpect to answer password
+  become: yes
+  package:
+    name: pexpect
```
evgeni, line 5: On EL8:
```console
[root@a8f349df7349 /]# yum install pexpect
Failed to set locale, defaulting to C.UTF-8
Last metadata expiration check: 0:00:08 ago on Tue Jan 12 12:57:58 2021.
No match for argument: pexpect
Error: Unable to find a match: pexpect
```

similar on Fedora 33 and Debian 10, where is this package coming from?

Should this be `python3-pexpect`?

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,121 @@
...
+    state: latest
+  # on Fedora this will be True due to the major version. Not nice, but works.
+  when: ansible_os_family != "RedHat" or ansible_distribution_major_version is version('8', '>')
```
evgeni, line 8: this evaluates to: "on debian" or "on all fedoras" or "on EL8+"
is that what you wanted?

evgeni (changes requested):

roles/realm_join/tasks/realm_join_ad.yml
ezr-ondrej: probably not it I see now that it differs on debian and RHEL.

roles/realm_join/tasks/realm_join_ad.yml
ezr-ondrej: It should .... python3-pexpect on RedHat > 8 and Fedora and python-pexpect on Debian.

ezr-ondrej:
I've extracted the pexpect install to separate file as it was cluttering the main file quite a bit already.

roles/realm_join/tasks/realm_join_ad.yml
```diff
@@ -0,0 +1,96 @@
+---
+- include_tasks: install_pexpect.yml
```
evgeni, line 2: this file is not in the PR? :)

roles/realm_join/tasks/main.yml
```diff
@@ -0,0 +1 @@
+- include_tasks: realm_join_ad.yml
```
evgeni, line 1: what's the reason to have all tasks in another file? prep-work for `realm_join_ipa.yml`?

roles/realm_join/tasks/realm_join_ad.yml
ezr-ondrej: :facepalm:

roles/realm_join/tasks/install_pexpect.yml
```diff
@@ -0,0 +1,37 @@
...
+  when:
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version is version('8', '>')
+
+- name: Ensure pexpect to answer password
+  become: yes
+  package:
+    name: python-pexpect
+    state: latest
+  # on Fedora this will be True due to the major version. Not nice, but works.
```
evgeni, line 16: I hightly doubt that `ansible_os_family: Debian` will ever be true on Fedora ;)

I think this comment can co now.

ezr-ondrej: xD yes, indeed ....

roles/realm_join/tasks/main.yml
ezr-ondrej: yeah, doesn't make much sense xD I think first I was thinking it could run the installer afterwards, but then decided against it.
So now it's tree steps, get the pexpect on the necessary version, join the realm and then prepare the keytabs for apache.

roles/realm_join/tasks/install_pexpect.yml
```diff
@@ -0,0 +1,36 @@
...
+      - epel-release
+      - python-pip
+    state: present
+  when:
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version is version('8', '<')
+
+- name: Ensure pexpect to answer password from pip
+  become: yes
+  pip:
```
evgeni, line 32: pexpect 2.3-11.el7 that is in EL7 is too old I guess?

roles/realm_join/tasks/install_pexpect.yml
```diff
@@ -0,0 +1,36 @@
+- name: Ensure pexpect to answer password
+  become: yes
+  package:
+    name: python3-pexpect
```
evgeni, line 4: instead of having two blocks for different OSes, we could have vars like this:

https://github.com/theforeman/forklift/blob/4090e636e55d83a39d7f7cdaf5ce9b1f3ed41dc8/roles/sos_report/tasks/main.yml#L2-L8

https://github.com/theforeman/forklift/blob/4090e636e55d83a39d7f7cdaf5ce9b1f3ed41dc8/roles/sos_report/vars/Debian.yml#L2
https://github.com/theforeman/forklift/blob/4090e636e55d83a39d7f7cdaf5ce9b1f3ed41dc8/roles/sos_report/vars/RedHat.yml#L2

It's up to you if you want to go that route, for a single package it might be overkill.

roles/realm_join/tasks/install_pexpect.yml
ezr-ondrej: Yes, we need `>= 3.3` for the ansible module to work.

roles/realm_join/tasks/install_pexpect.yml
ezr-ondrej: I :heart: it. Lets do it

evgeni (approved):
ACK from an "the ansible role reads fine" POV, I didn't test it actually works or matches the docs.

ekohl (approved):
Thanks for addressing all concerns!

PR #1246, opened 2020-12-14
Enable testing local puppet modules

ezr-ondrej:
Add posibility to use local modules for installer.
This helps if we need to test change we dont want to push to git just yet.

For various reasons like shame or more seriously security bugs testing.


roles/foreman_installer/tasks/module_prs.yml
```diff
@@ -38,3 +38,10 @@
     loop_var: module
   tags:
     - packages
+
+- include_tasks: module_local.yml
```
ekohl, line 42: I think you pushed this file in the AD PR.

ezr-ondrej: indeed... fixed :+1:

ezr-ondrej:
Do we want this?

PR #1250, opened 2020-12-15
Shorten pipeline hostnames due to reverse proxy size limits

ehelms:
I hit this in about all of my environments, so throwing it out there in hopes to get it accepted so I can stop accidentally committing this change with every PR I open.


evgeni (approved):
This will require updating https://github.com/theforeman/foreman-infra/blob/2a5cbbd3800eaeee8f425ea2345485be7efafe19/puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/pipelines.groovy#L3-L6

ehelms:
Thanks! Opened -- https://github.com/theforeman/foreman-infra/pull/1539

PR #1251, opened 2020-12-15
Drop fb-finish.bats in favor of playbooks/collect_debug.yml

ehelms:


evgeni:
Should delete-org.bats go the same route?

pipelines/vars/forklift_foreman.yml
```diff
@@ -12,7 +12,5 @@ server_box:
         - "fb-verify-packages.bats"
         - "fb-test-foreman.bats"
         - "fb-test-puppet.bats"
-      bats_teardown:
-        - "fb-finish.bats"
```
ekohl, line 16: This means it'll now use the default, which is `fb-destroy-organization.bats`. That doesn't work on Foreman itself. I think you need to keep it in but set it to an empty array instead. Same below.

bats/fb-finish.bats
```diff
@@ -1,12 +0,0 @@
...
-# vim: ft=sh:sw=2:et
-
-set -o pipefail
-
-load os_helper
-load foreman_helper
-
-@test "collect important logs" {
-  tail -n100 /var/log/{apache2,httpd}/*_log /var/log/foreman{-proxy,}/*log /var/log/messages > /root/last_logs || true
-  foreman-debug -q -d /root/foreman-debug || true
```
ekohl, line 11: sosreport on Debian doesn't call foreman-debug so we would lose that info. I don't mind seeing this go in the Ansible role, but that info is useful to have.

ekohl (changes requested):

bats/fb-finish.bats
ehelms: Do we collect this information today? I couldn't find any references to that path in Forklift or foreman-infra.

ehelms:
@evgeni I was not aware of something else that covered the delete organization test case.

evgeni:
Oh, is it a legitimate test case? I always thought that it's just cleanup.

ehelms:
Yup! It dates back to when organization delete would just be broken, often due to having to orchestrate repository and product deletes across backends, etc. So we added a test to say after we do all of these content actions, try and delete the organization to ensure the orchestration works.

ehelms:
Any further thoughts here?

bats/fb-finish.bats
evgeni: I don't think this is collected (as in moved anywhere non-ephemeral) today.

I vote "nuke it here" and let's add a separate "foreman_debug" role that can get loaded on Debian?

ekohl: Sounds good.

roles/bats/defaults/main.yaml
```diff
@@ -20,4 +20,3 @@ bats_tests:
 bats_tests_additional: []
 bats_teardown:
   - "fb-destroy-organization.bats"
```
ekohl, line 22: I now wonder if it should be empty by default and opt-in, but that doesn't have to block this.

ekohl (approved):

PR #1253, opened 2020-12-16
use proxy info instead of proxy show

evgeni:
proxy show seems to work on Foreman 2.1+, but not on 1.24 and older.
proxy info yields the same output, and works everywhere.

this was originally introduced in c2adb0daa031eef6dce089a2be720ed49283cfa8


evgeni:
(this is needed to keep the tests compatible with Satellite 6.6/6.7 ;))

ekohl (approved):

PR #1254, opened 2020-12-17
Omit foreman-proxy-content-parent-fqdn for nightly pipelines

ehelms:


pipelines/install/04-install_proxy.yaml
```diff
@@ -12,7 +12,7 @@
       - "--certs-tar-file {{ foreman_proxy_content_certs_tar }}"
       - "--foreman-proxy-trusted-hosts {{ server_fqdn.stdout }}"
       - "--foreman-proxy-trusted-hosts {{ ansible_nodename }}"
-      - "--foreman-proxy-foreman-base-url https://{{ server_fqdn.stdout }}"
+      - "{{ ('--foreman-proxy-foreman-base-url https://%s' % server_fqdn.stdout) if pipeline_version != 'nightly' else '' }}"
```
ehelms, line 15: This definitely won't scale for future versions, I cannot really think of a great way to do this. I tried using the `version` function for comparison but that function did not exist here.  Any ideas @evgeni ?

ehelms:
The reason for this change is https://github.com/theforeman/puppet-foreman_proxy_content/pull/309

pipelines/install/04-install_proxy.yaml
evgeni: "that function does not exist here"?

Can you show me the invocation and the error msg?

ekohl:
Shouldn't you make `--foreman-proxy-content-parent-fqdn` optional? The Foreman base URL is not going away.

ehelms:
> Shouldn't you make `--foreman-proxy-content-parent-fqdn` optional? The Foreman base URL is not going away.

Oops, duh. Yea thats the parameter I was aiming for.

pipelines/install/04-install_proxy.yaml
ehelms: The trick to get something to work, is to use it the way it's intended. Or remember to plug it in.

evgeni: You're talking in riddles, but I guess that means that works now?

evgeni:
The same change should also happen in:

https://github.com/theforeman/forklift/blob/33592862df05f392c74602f94a694ef3520263ab/pipelines/upgrade/04-install_proxy.yaml#L17
(based on `forklift_upgrade_version_start`)

https://github.com/theforeman/forklift/blob/33592862df05f392c74602f94a694ef3520263ab/playbooks/foreman_proxy_content.yml#L26
https://github.com/theforeman/forklift/blob/33592862df05f392c74602f94a694ef3520263ab/playbooks/foreman_proxy_content_dev.yml#L27
(no idea based on what)

and maybe we should drop `pipelines/pipeline_katello_lbproxy.yml` as noone is using that it seems? Or we should update it.

pipelines/install/04-install_proxy.yaml
```diff
@@ -16,6 +16,7 @@
       - "--foreman-proxy-oauth-consumer-key {{ oauth_consumer_key.stdout }}"
       - "--foreman-proxy-oauth-consumer-secret {{ oauth_consumer_secret.stdout }}"
       - "--foreman-proxy-content-parent-fqdn {{ server_fqdn.stdout }}"
+      - "{{ '' if (pipeline_version == 'nightly' or 2.3 is version(pipeline_version, '>')) else ('--foreman-proxy-content-parent-fqdn %s' % server_fqdn.stdout) }}"
```
ekohl, line 19: Just a question: does the following work as well?
```suggestion
      - "{{ ('--foreman-proxy-content-parent-fqdn %s' % server_fqdn.stdout) if (2.3 is version(pipeline_version, '<=')) else '' }}"
```

pipelines/install/04-install_proxy.yaml
```diff
@@ -16,6 +16,7 @@
       - "--foreman-proxy-oauth-consumer-key {{ oauth_consumer_key.stdout }}"
       - "--foreman-proxy-oauth-consumer-secret {{ oauth_consumer_secret.stdout }}"
       - "--foreman-proxy-content-parent-fqdn {{ server_fqdn.stdout }}"
```
ekohl, line 18: This line needs to be removed
```suggestion
```

ekohl (changes requested):
Could you also update the commit message to reflect the parameter change?

Another thought: after this, the only Katello specific parameter is `--certs-tar-file`. It would be great if we could make it optional in a future version. That would allow us to test a standalone proxy in a Foreman scenario as well.

pipelines/install/04-install_proxy.yaml
evgeni: are you talking about the re-ordering, or droping of the "is nightly" drop?

anyways, no it will not: `version()` can't compare `nightly` to `2.3`, so we never should pass `nightly` to it.

ekohl: It was the latter part.

pipelines/upgrade/04-install_proxy.yaml
```diff
@@ -14,7 +14,7 @@
       - "--foreman-proxy-foreman-base-url https://{{ server_fqdn.stdout }}"
       - "--foreman-proxy-oauth-consumer-key {{ oauth_consumer_key.stdout }}"
       - "--foreman-proxy-oauth-consumer-secret {{ oauth_consumer_secret.stdout }}"
-      - "--foreman-proxy-content-parent-fqdn {{ server_fqdn.stdout }}"
+      - "{{ '' if (pipeline_version == 'nightly' or 2.3 is version(pipeline_version, '>')) else ('--foreman-proxy-content-parent-fqdn %s' % server_fqdn.stdout) }}"
```
evgeni, line 17: Pipeline version is the *target* version of the upgrade pipeline. You probably want `forklift_upgrade_version_start` here

evgeni (changes requested):

pipelines/upgrade/04-install_proxy.yaml
ehelms: Ok will update... though by the time we have reached this line actually needing this check, we should be dropping it all together.

evgeni: The previous version would have omitted the param on 3.17 (to 18 to nightly) upgrade runs, and that would be wrong.

pipelines/upgrade/04-install_proxy.yaml
```diff
@@ -14,7 +14,7 @@
       - "--foreman-proxy-foreman-base-url https://{{ server_fqdn.stdout }}"
       - "--foreman-proxy-oauth-consumer-key {{ oauth_consumer_key.stdout }}"
       - "--foreman-proxy-oauth-consumer-secret {{ oauth_consumer_secret.stdout }}"
-      - "--foreman-proxy-content-parent-fqdn {{ server_fqdn.stdout }}"
+      - "{{ '' if (pipeline_version == 'nightly' or 2.3 is version(forklift_upgrade_version_start, '>')) else ('--foreman-proxy-content-parent-fqdn %s' % server_fqdn.stdout) }}"
```
evgeni, line 17: Actually, this will still omit the param on 17-18-nightly upgrades.

Given the upgrade_start will never be nightly, you can just omit (hah) the first part of the condition.

evgeni (changes requested):

evgeni (approved):

ekohl (approved):

ekohl:
Looks like something did fail:
```
17:25:57  TASK [foreman_installer : Join options] ****************************************
17:25:57  e[0;31mfatal: [pipe-up-katello-proxy-nightly-centos7]: FAILED! => e[0m
17:25:57  e[0;31m  msg: 'An unhandled exception occurred while templating ''[u''--certs-tar-file {{ foreman_proxy_content_certs_tar }}'', u''--foreman-proxy-trusted-hosts {{ server_fqdn }}'', u''--foreman-proxy-trusted-hosts {{ ansible_nodename }}'', u''--foreman-proxy-foreman-base-url https://{{ server_fqdn }}'', u''--foreman-proxy-oauth-consumer-key {{ oauth_consumer_key.stdout }}'', u''--foreman-proxy-oauth-consumer-secret {{ oauth_consumer_secret.stdout }}'', u"{{ '''' if (2.3 is version(forklift_upgrade_version_start, ''>'') else (''--foreman-proxy-content-parent-fqdn %s'' % server_fqdn) }}", u''--puppet-server-foreman-url https://{{ server_fqdn }}'', u''--puppet-server-max-active-instances 1'', u''--puppet-server-jvm-min-heap-size 1G'', u''--puppet-server-jvm-max-heap-size 1G'']''. Error was a <class ''ansible.errors.AnsibleError''>, original message: template error while templating string: expected token '')'', got ''else''. String: {{ '''' if (2.3 is version(forklift_upgrade_version_start, ''>'') elsee[0m
17:25:57  e[0;31m    (''--foreman-proxy-content-parent-fqdn %s'' % server_fqdn) }}'e[0m
```

Issue #1211, opened 2020-10-06
smoker tests fail to connect to non-nightly boxes


evgeni:
```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='pipeline-up-katello-3.16-centos7.n5.example.com', port=443): Max retries exceeded with url: /menu (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fccb76bcac8>: Failed to establish a new connection: [Errno -2] Name or service not known',))
```

the correct name would have been `pipeline-up-katello-3-16-centos7.n5.example.com`, not `pipeline-up-katello-3.16-centos7.n5.example.com` (dot vs hyphen in the katello version).

this is because the tests set:
```
smoker_base_url: "https://{{ forklift_server_name }}.{{ ansible_domain }}"
```

but I think this should be something like:
```
smoker_base_url: "https://{{ hostvars[forklift_server_name].ansible_fqdn }}"
```

PR #1258, opened 2021-01-05
use ansible_nodename of the server as smoker_url

evgeni:
the inventory name does not match the hostname for non-nightly boxes
which have a version with dots in their inventory name

Fixes: #1211

### DIFF
--- a/pipelines/install/06-smoker.yaml
+++ b/pipelines/install/06-smoker.yaml
@@ -5,7 +5,7 @@
   vars_files:
     - ../vars/install_base.yml
   vars:
-    smoker_base_url: "https://{{ forklift_server_name }}.{{ ansible_domain }}"
+    smoker_base_url: "https://{{ hostvars[forklift_server_name].ansible_nodename }}"
     smoker_variables:
       password: "{{ foreman_installer_admin_password|default('changeme') }}"
   roles:
--- a/pipelines/upgrade/10-smoker.yaml
+++ b/pipelines/upgrade/10-smoker.yaml
@@ -5,7 +5,7 @@
   vars_files:
     - ../vars/upgrade_base.yml
   vars:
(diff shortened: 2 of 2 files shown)


ekohl (approved):
This makes me think about https://github.com/theforeman/forklift/blob/2dbf2ebe1a70067b6cc84434a75d31135384291a/roles/foreman_proxy_content/tasks/install.yml#L2-L21 and wonder how that could be improved as well. I'm still wondering if it's needed at all. Separate discussion, but perhaps we should start to ensure DNS just works (we can do so in our vagrant setup).

ekohl:
This broke the pipeline:

```
13:23:14  TASK [pytest_project : Build command] ******************************************
13:23:14  [0;31mfatal: [pipe-foreman-smoker-nightly-centos7]: FAILED! => [0m
13:23:14  [0;31m  msg: |-[0m
13:23:14  [0;31m    The task includes an option with an undefined variable. The error was: {{ smoker_command_args }}: --driver chrome --variables '{{ smoker_variables_path }}' --base-url '{{ smoker_base_url }}' -vv: https://{{ hostvars[forklift_server_name].ansible_nodename }}: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'ansible_nodename'[0m
13:23:14  [0;31m  [0m
13:23:14  [0;31m    The error appears to be in '/root/forklift/roles/pytest_project/tasks/run.yml': line 2, column 3, but may[0m
13:23:14  [0;31m    be elsewhere in the file depending on the exact syntax problem.[0m
13:23:14  [0;31m  [0m
13:23:14  [0;31m    The offending line appears to be:[0m
13:23:14  [0;31m  [0m
13:23:14  [0;31m    ---[0m
13:23:14  [0;31m    - name: 'Build command'[0m
13:23:14  [0;31m      ^ here[0m
```

PR #1260, opened 2021-01-05
don't use hostname -f to get the server hostname

evgeni:
also don't set /etc/hosts, in CI we already do, and other users should just have working DNS, really

inspired by @ekohl's comment in #1258


evgeni:
Note: this will require changes to downstream playbooks too, which I can take care of!

roles/foreman_proxy_content/tasks/install.yml
```diff
@@ -1,24 +1,12 @@
...
-- name: 'Add /etc/hosts record for foreman_proxy_content'
-  delegate_to: "{{ foreman_proxy_content_server }}"
-  lineinfile:
-    dest: /etc/hosts
-    regexp: ".*{{ ansible_nodename }}$"
-    line: "{{ foreman_proxy_content_ip }} {{ ansible_nodename }}"
-    state: present
+- name: 'Get Server Hostname'
+  set_fact:
+    server_fqdn: "{{ hostvars[foreman_proxy_content_server].ansible_nodename }}"
```
ekohl, line 9: Perhaps we can classify this as well?
```suggestion
    foreman_server_fqdn: "{{ hostvars[foreman_proxy_content_server].ansible_nodename }}"
```

Perhaps even expand it (no idea if it's valid, but you get the idea):
```suggestion
    foreman_server_fqdn: "{{ hostvars[foreman_proxy_content_server].ansible_nodename }}"
    foreman_server_url: "https://{{ foreman_server_fqdn }}"
```

Alternatively we could read the config. `foreman-rake config` has a few settings we need (`foreman_url`, `oauth_consumer_key` and `oauth_consumer_secret`). I don't think it currently can easily output JSON, but that would be ideal. Perhaps a good enhancement.

ekohl (approved):
Some notes: the oauth and foreman_url + name reading is really foreman-proxy, not a content proxy. I'd like to extract it at some point so we can use it in vanilla Foreman as well.

Most comments here would be good follow up improvements and I'm fine with merging it as is.

PR #1265, opened 2021-01-07
Use pytest-xdist in smoker to launch 2 processes

ekohl:
In my experience, there is a speedup if you use multiple processes since there are times you wait. The number 2 was chosen because a large number of our tests runs a browser and that can take a bit of memory. Even on my laptop I saw diminishing returns after a while. 2 is a fairly conservative number, but can already help.


ehelms (approved):

evgeni:
in my env, this made things worse :(

ehelms:
Touche, this jumped my environment up to 1 hour. Somehow this seems to exacerbate whatever the issue caused by the chrome updated is. As I used to could run with `-n 4` locally running the pages and client tests and be done in 3-6 minutes.

PR #1267, opened 2021-01-08
Revert "Use pytest-xdist in smoker to launch 2 processes"

evgeni:
Reverts theforeman/forklift#1265

This made things worse and CI now takes 4h for smoker, instead of 1.


ekohl (commented):
Should we hold back on this until we've tested https://github.com/theforeman/forklift/pull/1268? I think it should still be a good thing with a working browser.

evgeni:
we can, sure

evgeni:
closing for now, seems to be relatively stable with the pinned chrome

PR #1268, opened 2021-01-08
pin chromium to 86.x which works fine

evgeni:
87.x randomly waits for 10 minutes before doing *anything*, essentially breaking our nightly pipelines.


evgeni:
https://bugzilla.redhat.com/show_bug.cgi?id=1914166 is the issue

roles/smoker/defaults/main.yml
```diff
@@ -8,5 +8,6 @@ smoker_base_url:
 smoker_command_args: "--driver chrome --variables '{{ smoker_variables_path }}' --base-url '{{ smoker_base_url }}' --html 'report/htmlreport.html' -vv --numprocesses 2"
 smoker_markers:
 smoker_browser_packages:
```
ekohl, line 10:
```suggestion
# https://bugzilla.redhat.com/show_bug.cgi?id=1914166
smoker_browser_packages:
```

ekohl (approved):
Might make sense to link to the BZ? :+1: for pinning.

PR #1270, opened 2021-01-13
disable vagrant-scp by default

evgeni:
we don't use it in CI and installing it is rather expensive (>1 minute)


ekohl (approved):
I had the same thought a while back but didn't act on it yet.

PR #1271, opened 2021-01-13
add pulp_installer requirements

evgeni:


evgeni:
@ekohl it helps to run the pipelines if they actually install all the bits :see_no_evil:

ekohl (approved):

PR #1279, opened 2021-01-21
Addresses 1036 - Add support for a katello-hammer-devel all-in-one box.

jeffmcutter:
https://github.com/theforeman/forklift/issues/1036

### DIFF
--- a/playbooks/katello_hammer_devel.yml
+++ b/playbooks/katello_hammer_devel.yml
@@ -1,0 +1,10 @@
+---
+- hosts: all
+  roles:
+    - role: katello_devel
+      become: yes
+    - etc_hosts
+    - git
+    - role: ruby_scl
+      become: yes
+    - hammer_devel
--- a/vagrant/boxes.d/99-local.yaml.example
+++ b/vagrant/boxes.d/99-local.yaml.example
@@ -48,6 +48,21 @@ centos7-hammer-devel:
       hammer_devel_github_username: <REPLACE ME>
       #hammer_devel_server: centos7-katello-devel.custom-domain
 
+# Useful for content-view version import/export where hammer must be run locally.
+centos7-katello-hammer-devel:
+  primary: true
+  box: centos7
+  memory: 8192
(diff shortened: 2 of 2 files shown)


ehelms:
Sorry for the long delay reviewing. Is this intended to be a hammer development box? Or a katello development box that also has hammer setup via git for development as well? If it's the latter, I think we could opt to add a variable and conditional inclusion of `hammer_devel` in the normal devel box.

jeffmcutter:
The use case was to develop hammer content-view version export/import which needs to be run on the katello server to work.  That being said, it seems that content-view version import/export is deprecated with Pulp 2.  I'm not sure what the situation is for Pulp 3.  Perhaps this isn't useful at all anymore, would need to research how the import/export works with Pulp 3.

https://github.com/Katello/hammer-cli-katello/pull/718

ekohl:
Hammer is supposed to be 100% remote and that content view part was IMHO always a bug. With Pulp 3 we're making steps forward. I'm not sure how far it is and what the exact current design is, but maybe @jjeffers or @jturel can point to the correct issue.

jturel:
> The use case was to develop hammer content-view version export/import which needs to be run on the katello server to work. That being said, it seems that content-view version import/export is deprecated with Pulp 2. I'm not sure what the situation is for Pulp 3. Perhaps this isn't useful at all anymore, would need to research how the import/export works with Pulp 3.
> 
> [Katello/hammer-cli-katello#718](https://github.com/Katello/hammer-cli-katello/pull/718)

Hey @jeffmcutter ! We just released Katello 4.0 which has all of the new pulp3-based import/export functionality. Upstream docs should reflect the new hammer commands `hammer content-export` and `hammer content-import`. Even better news is that both of those commands should be able to be executed via hammer either on the Katello server itself or external.

jeffmcutter:
Thanks for the confirmation.  Seems this PR isn't very useful anymore.  I'll go ahead and close it.  Thanks!

PR #1281, opened 2021-01-22
Test bootstrap script

swadeley:
Hello

For this, my first, BATS test I need environment variables on the client to set Satellite and Capsule name.

I could not get Ansible "environment:" method to work., so tired creating them in an "BATS session. Then I found that each BATS @test block starts a new BASH shell, so any environment variables created in an earlier @test session are not present in the next @test session.

 So then I put the variables in the ".bash_profile", fixed the FQDNs using Ansible,  and then sourced the ".bash_profile" in each @test 

Is that an acceptable method?


ehelms:
We are looking to move client tests to our smoker library, starting with https://github.com/theforeman/smoker/pull/11 -- give that a look and see if it might be a better place for the tests you want to introduce?

PR #1283, opened 2021-01-27
Fix bug in swapfile role

wbclark:
This fixes a bug that occurs when running bats tests:

```
TASK [activate /swapfile] ******************************************************
fatal: [centos7-katello-bats-ci]: FAILED! => 
  msg: |-
    The conditional check 'swapfile_active | failed' failed. The error was: template error while templating string: no filter named 'failed'. String: {% if swapfile_active | failed %} True {% else %} False {% endif %}
```

### DIFF
--- a/roles/swapfile/tasks/main.yml
+++ b/roles/swapfile/tasks/main.yml
@@ -33,7 +33,7 @@
 
 - name: "activate {{ swapfile_path }}"
   command: "swapon {{ swapfile_path }}"
-  when: swapfile_active | failed
+  when: swapfile_active is failed
 
 - name: set vm.swappiness to 60
   sysctl:


evgeni:
This doesn't sound right. Especially not the bool part.

Can yiu try "is failed" as documented in https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#playbooks-conditionals instead?

wbclark:
I originally added:

```
- debug:
    msg: "{{ swapfile_active }}"
```

And what I got was:

```
TASK [swapfile : debug] ********************************************************
ok: [centos7-katello-bats-ci] => 
  msg:
    changed: true
    cmd: swapon --summary | grep '^/swapfile'
    delta: '0:00:00.007606'
    end: '2021-01-28 13:52:11.569738'
    failed: false
    rc: 0
    start: '2021-01-28 13:52:11.562132'
    stderr: ''
    stderr_lines: []
    stdout: "/swapfile                              \tfile\t2097148\t0\t-2"
    stdout_lines:
    - "/swapfile                              \tfile\t2097148\t0\t-2"
 [started TASK: activate {{ swapfile_path }} on centos7-katello-bats-ci]
 [started TASK: swapfile : set vm.swappiness to 60 on centos7-katello-bats-ci]
```

So `swapfile_active.failed` selects the failed field seen above and `| bool` is probably not necessary but I usually like to ensure types so that I don't find myself asking later "what is that thing again?".

Is the intention for this check something different, @evgeni ?

roles/swapfile/tasks/main.yml
```diff
@@ -33,7 +33,7 @@
 
 - name: "activate {{ swapfile_path }}"
   command: "swapon {{ swapfile_path }}"
-  when: swapfile_active | failed
+  when: swapfile_active.failed | bool
```
ekohl, line 36: I think what @evgeni suggested was:
```suggestion
  when: swapfile_active is failed
```

evgeni: Correct

evgeni:
So the logic here is:
Check if the the swapfile is already enabled
If not (command is failed) enable it

The generic way to do this in Ansible is to register the result of a task and then have a `when: registered_name is failed` on the second one.

The .failed attribute exists, but is not supposed to be used directly.

Now why the old |failed notation bugged out for you, I don't know. Which Ansible version is that?

wbclark:
Thanks -- my Ansible version is Ansible 2.9.14

wbclark:
Updated based on https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#playbooks-conditionals

Thanks for the reference!

evgeni (approved):

PR #1285, opened 2021-01-29
Enable content proxy testing on EL8 Katello pipelines

ehelms:


ehelms:
Blocked on https://community.theforeman.org/t/pulp-3-selinux-issues-on-centos-8-content-proxy/22153

pipelines/vars/install_base.yml
```diff
@@ -1,7 +1,7 @@
 forklift_name: "pipe-{{ pipeline_type }}-{{ pipeline_version }}-{{ pipeline_os }}"
 forklift_server_name: "pipe-{{ pipeline_type }}-server-{{ pipeline_version }}-{{ pipeline_os }}"
 forklift_proxy_name: "pipe-{{ pipeline_type }}-proxy-{{ pipeline_version }}-{{ pipeline_os }}"
-forklift_smoker_name: "pipe-{{ pipeline_type }}-smoker-{{ pipeline_version }}-{{ pipeline_os }}"
+forklift_smoker_name: "pipe-{{ pipeline_type }}-smoker-{{ pipeline_version }}-centos7"
```
ekohl, line 4: The reason I kept in `pipeline_os` previously is that it won't conflict if you would run multiple pipelines, even if it doesn't match the actual OS. This way you can spin up a pipeline for CentOS 7, keep the boxes and spin up another one on CentOS 8 without having them conflict.

jlsherrill (approved):

PR #1286, opened 2021-02-01
add foreman 2.4 to versions

evgeni:


ekohl (approved):

pcreech (approved):
ACK

evgeni:
thanks for the ACKs -- please don't merge before we actually are doing the RC, otherwise things will explode

PR #1287, opened 2021-02-05
Always run katello-agent tests

ehelms:
This will enable EL7 to keep testing with Pulp 2 code dropping away, but also enables it for EL8 as part of https://github.com/theforeman/foreman-packaging/pull/6282)


jturel (approved):
LGTM

PR #1289, opened 2021-02-09
Use puppet_repositories from theforeman.operations collection

ehelms:


ekohl (changes requested):
This should document how to get the ansible collection.

Another thing I dislike is that the requirements states pulp_installer is needed, but I won't need that for most operations. It means I'll pull in a lot more than I really need.

Now I also need to get stuff from galaxy. As someone who dislikes Ansible, this feels like a step back. I'm leaning to holding off on using any collections until there's a real benefit. I can see it in complex roles like where we now use hammer, but here it feels complex. It is literally more lines of code even if you don't take the added dependency into account.

evgeni:
The idea of the operations collection was that we also dogfood it, and forklift seems like a good place for dogfooding :)

We could split the requirements files into Foreman and Pulp easily, would that help? Eric and I expected this to be a problem at some point due to different usecases.

ekohl:
> We could split the requirements files into Foreman and Pulp easily, would that help? Eric and I expected this to be a problem at some point due to different usecases.

I think that would be a good idea.

ansible.cfg
```diff
@@ -4,8 +4,8 @@ display_skipped_hosts = false
 show_per_host_start = true
 nocows = 1
 inventory = $PWD/inventories/
-roles_path = ./playbooks/galaxy_roles:./user_playbooks/roles:./roles
-collections_paths = ./playbooks/galaxy_collections
+roles_path = .cache/galaxy_collections:./user_playbooks/roles:./roles
```
evgeni, line 7: I don't think that setting `roles_path` to something …collections is the right thing. There are no roles inside that folder, only collections (that might contain roles, but that is not relevant for the "roles_path".

playbooks/luna_demo_environment.yml
```diff
@@ -11,10 +11,8 @@
     - etc_hosts
     - epel_repositories
     - ostree_repositories
-    - puppet_repositories
-    - foreman_repositories
-    - katello_repositories
-    - foreman_installer
+    - theforeman.operations.puppet_repositories
+    - theforeman.operations.foreman_repositories
```
evgeni, line 15: this drops the call to the installer role?

roles/foreman_staging_repositories/tasks/redhat.yml
```diff
@@ -0,0 +1,98 @@
...
+
+    - name: 'Add Pulp {{ foreman_staging_repositories_pulp_version }} Stable repository'
+      yum_repository:
+        name: pulp-repository
+        description: Pulp release repository
+        baseurl: "https://repos.fedorapeople.org/repos/pulp/pulp/{{ foreman_staging_repositories_pulp_release }}/{{ foreman_staging_repositories_pulp_version }}/$releasever/$basearch"
+        gpgcheck: no
+        enabled: yes
+      when:
+        - foreman_staging_repositories_katello_version is version('4.0', '<=')
```
evgeni, line 65: This will fail if `foreman_staging_repositories_katello_version` is `nightly`

```suggestion
        - foreman_staging_repositories_katello_version != 'nightly'
        - foreman_staging_repositories_katello_version is version('4.0', '<=')
```

roles/foreman_staging_repositories/tasks/redhat.yml
```diff
@@ -0,0 +1,98 @@
...
+
+    - name: Add Pulp Nightly repository
+      yum_repository:
+        name: pulp-repository
+        description: Pulp nightly repository
+        baseurl: "https://repos.fedorapeople.org/repos/pulp/pulp/testing/automation/2-master/stage/$releasever/$basearch"
+        gpgcheck: no
+        enabled: yes
+      when:
+        - foreman_staging_repositories_katello_version is version('4.0', '<=')
```
evgeni, line 77: same as above

roles/foreman_staging_repositories/tasks/redhat.yml
evgeni: addendum: fail on python3, not on 2 :woman_shrugging:

roles/foreman_server_repositories/meta/main.yml
```diff
@@ -6,12 +6,12 @@ dependencies:
       - ansible_distribution_major_version == '7'
   - role: ansible_repositories
     when: foreman_server_repositories_ansible|bool
-  - role: puppet_repositories
+  - role: theforeman.operations.puppet_repositories
```
evgeni, line 9: the forklift `puppet_repositories` role would erase `puppet5-release` before installing `puppet6-release`, which the operations one doesn't.

we had that for p5 to p6 upgrades, as their RPM would not conflict but not obsolete the old one…

the p7 files seem to have the same issue, given we don't support p7 yet it's not an issue today, but will be in future

ehelms: Sounds like when we add p7 over there we should add that ability.

evgeni: Yeah, I was thinking something along the lines of:

```yaml
supported_versions: [6,7]
removed_versions: "{{ supported_versions | difference([puppet_version]) }}"
```

(and tasks doing so)

ehelms: Sounds like we should file an issue over on the collection to track this?

evgeni: https://github.com/theforeman/foreman-operations-collection/issues/20

ehelms: Thanks for filing it!

ehelms:
> We could split the requirements files into Foreman and Pulp easily, would that help? Eric and I expected this to be a problem at some point due to different usecases.

I am not seeing the value in splitting. Having one requirements.yaml means users running a single command, means our CI not having to track job-to-job which file to install from and when I checked locally this totaled 6 MB. If I extrapolate, and we started using FAM in some of the seeding roles, would that be another requirements.yaml file?

ekohl (changes requested):
I still think this is a bad execution.

First of all, we lose the ability to upgrade Puppet between releases. https://github.com/theforeman/foreman-operations-collection/issues/20 is IMHO a blocker to merging it.

Then it unconditionally enables Katello repositories. That's also a big regression IMHO. Foreman must be able to install without Katello and the best way to test it is to not have the repository present. That it slows down tests is maybe a smaller issue but it also means that Foreman's pipeline can be red if the Katello repositories are offline.

Lastly I really dislike adding a dependency on ansible-galaxy, especially if it also pulls in the whole pulp installer with all its dependencies.

ehelms:
> Then it unconditionally enables Katello repositories. That's also a big regression IMHO. Foreman must be able to install without Katello and the best way to test it is to not have the repository present. That it slows down tests is maybe a smaller issue but it also means that Foreman's pipeline can be red if the Katello repositories are offline.

I am missing something here. Katello repositories are only enabled when a variable is defined for that purpose: https://github.com/theforeman/foreman-operations-collection/blob/develop/roles/foreman_repositories/tasks/redhat.yml#L26

> Lastly I really dislike adding a dependency on ansible-galaxy, especially if it also pulls in the whole pulp installer with all its dependencies.

You aren't really saying why you dislike having the dependency, or why it is bad for Forklift. Can you expand in a way that I can either make changes to address or explain more about it?

requirements.yml
```diff
@@ -1,2 +1,4 @@
 collections:
   - pulp.pulp_installer
+  - name: theforeman.operations
+    version: '>=0.2.0'
```
evgeni, line 4:
```suggestion
    version: '>=0.3.0'
```

to get the new variable names

roles/foreman_installer/tasks/install.yml
```diff
@@ -6,17 +6,19 @@
...
-      {{ foreman_installer_options_joined }}
+- name: Run installer
+  import_role:
+    name: theforeman.operations.installer
+  vars:
+    installer_scenario: "{{ foreman_installer_scenario }}"
+    installer_command: "{{ foreman_installer_command }}"
+    installer_verbose: "{{ foreman_installer_verbose|bool }}"
+    installer_no_colors: "{{ foreman_installer_no_colors|bool }}"
+    installer_options: "{{ foreman_installer_options + foreman_installer_options_internal_use_only + disable_system_checks }}"
```
evgeni, line 21: needs foreman prefix now, or rather dropping them here, as they are already defined in `foreman_installer/defaults`

(options one will be funny)

roles/foreman_server_repositories/meta/main.yml
evgeni: Thanks for fixing it!

ehelms:
I'll get around to updating this one once https://github.com/theforeman/forklift/pull/1313 is in as I feel like I should get that set of changes in first and then make this one.

roles/forklift_versions/library/forklift_versions.py
```diff
@@ -81,9 +81,11 @@ def main():
...
                         'foreman_repositories_version': version['foreman'],
+                        'foreman_staging_repositories_version': version['foreman'],
                         'foreman_client_repositories_version': version['foreman'],
-                        'katello_repositories_version': version['katello'],
-                        'katello_repositories_pulp_version': version['pulp'],
+                        'foreman_repositories_katello_version': version['katello'],
+                        'foreman_staging_repositories_katello_version': version['katello'],
+                        'foreman_staging_repositories_pulp_version': version['pulp'],
                         'pulp_repositories_version': version['pulp'],
                         'puppet_repositories_version': version['puppet'],
```
evgeni, line 90: this needs translating (well, adding) to the new name used by FOC.

roles/forklift_versions/library/forklift_versions.py
```diff
@@ -81,9 +81,11 @@ def main():
...
                 forklift_vars = {
                         'foreman_repositories_version': version['foreman'],
+                        'foreman_staging_repositories_version': version['foreman'],
                         'foreman_client_repositories_version': version['foreman'],
-                        'katello_repositories_version': version['katello'],
-                        'katello_repositories_pulp_version': version['pulp'],
+                        'foreman_repositories_katello_version': version['katello'],
+                        'foreman_staging_repositories_katello_version': version['katello'],
+                        'foreman_staging_repositories_pulp_version': version['pulp'],
                         'pulp_repositories_version': version['pulp'],
```
evgeni, line 89: is anything still using the standalone `pulp_repositories` role?

README.md
```diff
@@ -38,6 +38,7 @@ This will walk through the simplest path of spinning up a production test enviro
 ```
 git clone https://github.com/theforeman/forklift.git
 cd forklift
+ansible-galaxy collection install -r requirements.yml
```
ekohl, line 41: Does this work with Ansible 2.7 or should the Requirements section be updated?

ekohl (commented):
In general I'm not a fan of Ansible and having to rely on more Ansible tools is something that I'm hesitant about.

For example, if I want to make a change to a role that's used and test it out, I know how to do it and easily send a PR. Instead of `ansible-galaxy`, can I point it to a git checkout to develop on?

README.md
evgeni: It's 2.9+, yeah

ehelms:
> For example, if I want to make a change to a role that's used and test it out, I know how to do it and easily send a PR. Instead of `ansible-galaxy`, can I point it to a git checkout to develop on?

You can do:

```
collections:
  - name: git+file:///home/user/path/to/repo/.git
    type: git
```

I could add this to the docs to give some lead in for developers to know how to do it.

ehelms:
See https://github.com/theforeman/forklift/pull/1313 for updated documentation regarding questions here. That PR is the first use of the collection that I want to get in and then come back and update this PR.

pipelines/pipeline_katello_lbproxy.yml
```diff
@@ -42,7 +42,6 @@
   become: yes
   vars:
     foreman_server_repositories_katello: true
-    puppet_repositories_version: 4
```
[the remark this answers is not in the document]
evgeni: this pipeline is unused, so it should be either removed, or updated to latest.

evgeni:
Thanks @ehelms!

I think there is one thing left: the old role used `puppet_repositories_version` to indicate which version to install, while the new one uses `foreman_puppet_repositories_version`, but nothing sets this new variable right now. We don't deploy anything but 6 *right now*, so it won't break today, but I guess soon we'll want to try 7.

We can either add a translation step that does `foreman_puppet_repositories_version: "{{ puppet_repositories_version }}"` or teach the `forklift_versions` role to set the right variable directly.

I prefer the later.

ehelms:
Done (I hope)

playbooks/bats_pipeline_katello_nightly.yml
```diff
@@ -5,7 +5,7 @@
     katello_repositories_version: nightly
     foreman_repositories_version: nightly
     foreman_repositories_environment: staging
-    puppet_repositories_version: 5
+    foreman_puppet_repositories_version: 5
```
evgeni, line 8: I do wonder if this file can't go to. katello nightly with puppet5?! ;)

evgeni:
Ha!

`git grep` says you also want to edit:

https://github.com/theforeman/forklift/blob/46e782927c2b20b96f6f4f6346dcff88cf9c04e4/vagrant/lib/forklift/box_factory.rb#L102

evgeni:
@ehelms would you mind finishing this up? the recent puppet.com key rotation broke nightly pipelines, and I expect this to fix it once and forever

ehelms:
Fixed up that one spot and dropped that playbook

evgeni (approved):

playbooks/bats_pipeline_katello_nightly.yml
```diff
@@ -1,27 +0,0 @@
-- hosts: all
```
ekohl, line 1: Should we drop https://github.com/theforeman/forklift/blob/master/playbooks/bats_pipeline_foreman_nightly.yml at the same time for consistency?

ekohl (approved):

playbooks/bats_pipeline_katello_nightly.yml
evgeni: probably a good idea, yeah, even if unrelated to the original change here.

ehelms:
There's probably a few things we ought to clean up, I am happy to do that separately from the intent of this change.

evgeni:
I dropped the one file Ewoud pointed out in https://github.com/theforeman/forklift/pull/1383

If you see more, please do the same :)

PR #1295, opened 2021-02-12
Support ansible-lint 5.0 and drop custom rules/

ehelms:


evgeni (approved):
I've opened https://github.com/theforeman/forklift/issues/1296 to track the reduction of the skip list, but ACK for now.

PR #1298, opened 2021-02-12
Fix some ansible lint rules

ehelms:


roles/fips/tasks/main.yml
```diff
@@ -21,9 +21,10 @@
...
-                host={{ ansible_ssh_host }}
-                state=started
+  wait_for:
+    state: started
+    host: "{{ ansible_ssh_host }}"
+  delegate_to: localhost
 
 - name: 'Verify FIPS enabled - If failed means the machine is not FIPS enabled'
-  shell: cat /proc/sys/crypto/fips_enabled | grep 1
+  shell: set -o pipefail && cat /proc/sys/crypto/fips_enabled | grep 1
```
evgeni, line 30: This can be just `command: grep -q 1 /proc/…`

roles/katello_client/tasks/main.yml
```diff
@@ -19,7 +19,7 @@
     server_insecure: "{{ katello_client_insecure }}"
   ignore_errors: "{{ katello_client_ignore_registration_errors }}"
   when:
-    - katello_client_activationkey == ""
+    - not katello_client_activationkey
```
evgeni, line 22: I'd drop that from defaults and check for `is defined` instead

roles/puppet_agent/tasks/main.yml
```diff
@@ -27,7 +27,7 @@
     - name: add server option
       set_fact:
         puppet_agent_bootstrap_command: "{{ puppet_agent_bootstrap_command }} --server {{ puppet_agent_bootstrap_server }}"
-      when: puppet_agent_bootstrap_server != ""
+      when: puppet_agent_bootstrap_server
```
evgeni, line 30: Same "is defined" sounds nicer here?

evgeni (changes requested):

roles/katello_client/tasks/main.yml
```diff
@@ -29,7 +29,7 @@
     server_insecure: "{{ katello_client_insecure }}"
   ignore_errors: "{{ katello_client_ignore_registration_errors }}"
   when:
-    - katello_client_activationkey != ""
+    - katello_client_activationkey
```
evgeni, line 32:
```suggestion
    - katello_client_activationkey is defined
```

otherwise it will error in the case it's not defined as undef doesn't cast to bool

roles/libvirt/tasks/tftp.yml
```diff
@@ -33,10 +33,15 @@
   shell: cp /usr/share/syslinux/{pxelinux.0,menu.c32,chain.c32} /var/lib/tftpboot
 
 - name: 'chown /var/lib/tftpboot for foreman-proxy user'
-  shell: chown -R foreman-proxy:nobody /var/lib/tftpboot
+  file:
+    path: /var/lib/tftpboot
+    state: directory
+    owner: foreman-proxy
+    group: nobody
+    recursive: true
```
evgeni, line 41:
```suggestion
    recurse: true
```

at least if I trust https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html#parameter-recurse

ehelms: I think I read the description and got recursively stuck in my head.

evgeni (approved):

PR #1299, opened 2021-02-15
Enable fips for nightly pipeline testing

jlsherrill:


jlsherrill:
Opening this for discussion, it was expressed by the pulp team that it would be highly desirable to run a regular pipeline test with fips enabled.  It feels like there's no reason not to enable on our night (assuming everything is passing).  There were some problems with the existing fips role, including:

1) /boot isn't on a 2nd partition within my centos 7 box, so this command failed. I originally had it check both /boot and then fallback to /  but...
2) passing boot= option to grubby seemed to cause it to fail to boot completely.  I'm not sure why, or why it was working before, but passing just fips=1 caused the reboot to work flawlessly

When testing with this it fails pretty hard right now with:
* uploading an rpm
* syncing a repo
* capsule syncs

But the pulp team is working on some fixes which we can integrate into our rpm builds.  A couple of questions come up:

1) Did i enable this 'correctly'?
2) do we want to do enable this for both katello and foreman pipelines (which is what it does right now).  I haven't tested a foreman only pipeline yet

evgeni:
Shall we retry this?

I had made some changes to the fips role (https://github.com/theforeman/forklift/commits/master/roles/fips/), also one for boot: https://github.com/theforeman/forklift/commit/69cfb0fdd6a85f5c1b8bff8e774e2f44c101ad99

bmbouter:
+1 to retrying this. Pulp cannot test its FIPS related features due to GHA not supporting it.

PR #1300, opened 2021-02-16
Remove enable pki-core workaround

ekohl:
The installer started to take care of this. It can be removed safely because there has been no Katello release on EL8 and is only needed for Candlepin.


ehelms (approved):
Merging so that nightly can run without this work around.

PR #1301, opened 2021-02-17
test for selinux denials

evgeni:


bats/fb-verify-selinux.bats
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bats
+# vim: ft=sh:sw=2:et
+
+set -o pipefail
+
+load os_helper
+
+@test "ensure no SELinux denials" {
+  if tIsRedHatCompatible; then
```
ekohl, line 9: Should this also check for `selinuxenabled`?

evgeni: it could, I guess, but we're not disabling selinux in our ci?

ehelms (approved):
We will find that this fails nightly and at least 4.0, and would not surprise me to see other releases fail. So we may find we need to come back and selectively enable this to unblock pipelines.

Looks good, I tested it locally. Thanks for the quick action on it.

bats/fb-verify-selinux.bats
```diff
@@ -0,0 +1,14 @@
...
+    run ausearch --message AVC
+    echo "$output"
+    [ "${status}" -eq 1 ]
```
ekohl, line 12: Why did you choose exit code 1? Accoding to `man ausearch` this is:
> if nothing found, or argument errors or minor file acces/read errors

evgeni: correct, I want it not to find anything, because if it finds something, there were denials.

you could also `-ne 0` ;)

ekohl (approved):

PR #1303, opened 2021-02-18
forklift_versions: allow specifying an offset to get an older version

evgeni:
this can be used to install a version n server with an n-1 proxy

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -82,35 +82,6 @@ When you are finished with the test, you can tear down the associated infrastruc
     ansible-playbook pipelines/install_pipeline.yml -e forklift_state=up -e pipeline_os=centos7 -e pipeline_type=katello -e pipeline_version=nightly -e pipeline_proxy_delay=1
     ansible-playbook pipelines/upgrade_pipeline.yml -e forklift_state=up -e pipeline_os=centos7 -e pipeline_type=katello -e pipeline_version=3.10
 
-## Creating Pipelines
-
-If you wish to add a new version of an existing pipeline (e.g. a new Katello release), you only have to add the corresponding vars files to `pipelines/vars/`.
-
-For Katello 3.11, you'd be adding the following two files:
-
-`pipelines/vars/katello_3.11.yml`:
-```yaml
-forklift_name: pipeline-katello-3.11
-forklift_boxes:
-  pipeline-katello-3.11-centos7:
-    box: centos7
-    memory: 8192
-  pipeline-proxy-3.11-centos7:
-    box: centos7
-    memory: 3072
-katello_repositories_version: '3.11'
-katello_repositories_pulp_version: '2.19'
-foreman_repositories_version: '1.21'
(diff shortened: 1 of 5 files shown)


ehelms:
I think I may need to see this in action to understand how it will be used.

evgeni:
@ehelms like this?

docs/testing.md
```diff
@@ -78,37 +79,9 @@ When you are finished with the test, you can tear down the associated infrastruc
 #### Examples
 
     ansible-playbook pipelines/install_pipeline.yml -e forklift_state=up -e pipeline_os=debian10 -e pipeline_type=foreman -e pipeline_version=nightly
+    ansible-playbook pipelines/install_pipeline.yml -e forklift_state=up -e pipeline_os=centos7 -e pipeline_type=katello -e pipeline_version=nightly -e pipeline_proxy_delay=1
```
ehelms, line 82: What do you think about: 

 * pipeline_proxy_version_offset=1
 * pipeline_proxy_versions_back=1

evgeni: I like the version offset one, used it now.

ehelms:
This probably complicates things, can we modify the box name on spin up if the offset is set? That way I have clearer understanding of the version of the proxy deployed? e.g.

```
changed: [pipe-katello-server-nightly-centos7]
changed: [pipe-katello-proxy-3.18-centos7]
```

evgeni:
No, the box name generation happens before the role is executed.

The only way to achieve that if we *manually* pass the proxy version to the playbook like:

```console
$ ansible-playbook pipelines/install_pipeline.yml -e forklift_state=up -e pipeline_os=centos7 -e pipeline_type=katello -e pipeline_server_version=nightly -e pipeline_proxy_version=3.18
```

ehelms:
What if we put the offset in the box name? `pipe-katello-proxy-nightly-1-centos7`  that kinda looks like nightly minus 1 :) or pipe-katello-proxy-offset-1-centos7

evgeni:
and regular boxes would then be `nightly-0`?

ehelms:
> and regular boxes would then be `nightly-0`?

I guess, I realize it is a big ugly. I am just worried when making use of this I will get myself confused debugging something about what is what version.

evgeni:
Would a shiny `/etc/motd` help?

ehelms:
> Would a shiny `/etc/motd` help?

Can it just say "ERIC THE VERSION OF THE PROXY IS X.Y - Evgeni" ?

evgeni:
Yeah, something like that ;)

evgeni:
The hostnames we generate are already quite long and I don't think adding a single digit would make the difference obvious - but adding a longer string will bring issues with too long names.

ehelms:
Since there is not a rather good solution for this, let's give this a try as is and see how often we bump into said wall.

evgeni:
going to close this and replace it with something else.

PR #1305, opened 2021-02-18
Remove empty strings from locale list

ehelms:


evgeni:
I think the first list isn't required anymore, but doesn't hurt either.

Otherwise clever abuse of '' being False-y, and you know what I think about clever code ;)

`select("ne", "")` might be more obvious ;) (untested)

Why does your env have those vars set to empty anyways?

ekohl:
You know I'm a fan of context and should complain about the lack of it in this PR. The context is https://community.theforeman.org/t/foreman-nightly-deb-pipeline-580-failed/22454.

evgeni:
aha, yeah, that makes sense!

ehelms:
I don't know what you want me to do here.

roles/foreman_installer/tasks/locales.yml
```diff
@@ -11,7 +11,7 @@
 - name: 'Ensure ENV locales are available'
   locale_gen:
     name: "{{ item }}"
-  loop: "{{ query('env', 'LANG', 'LC_ADDRESS', 'LC_ALL', 'LC_COLLATE', 'LC_CTYPE', 'LC_IDENTIFICATION', 'LC_MEASUREMENT', 'LC_MESSAGES', 'LC_MONETARY', 'LC_NAME', 'LC_NUMERIC', 'LC_PAPER', 'LC_TELEPHONE', 'LC_TIME') | list | unique }}"
+  loop: "{{ query('env', 'LANG', 'LC_ADDRESS', 'LC_ALL', 'LC_COLLATE', 'LC_CTYPE', 'LC_IDENTIFICATION', 'LC_MEASUREMENT', 'LC_MESSAGES', 'LC_MONETARY', 'LC_NAME', 'LC_NUMERIC', 'LC_PAPER', 'LC_TELEPHONE', 'LC_TIME') | list | unique | select() | list }}"
```
evgeni, line 14:
```suggestion
  loop: "{{ query('env', 'LANG', 'LC_ADDRESS', 'LC_ALL', 'LC_COLLATE', 'LC_CTYPE', 'LC_IDENTIFICATION', 'LC_MEASUREMENT', 'LC_MESSAGES', 'LC_MONETARY', 'LC_NAME', 'LC_NUMERIC', 'LC_PAPER', 'LC_TELEPHONE', 'LC_TIME') | unique | select("ne", "") | list }}"
```

evgeni: @ehelms that ;)

ehelms: That lacks a certain.... panache of cleverness. But Ok I did it :) Thanks for the snippet

evgeni (approved):
you can still drop the first `list` as in my suggestion, but I don't care too much ;)

PR #1308, opened 2021-02-24
place template export in /run/foreman

evgeni:
/tmp and friends are hidden behind PrivateTmp and others might get
SELinux denials

### DIFF
--- a/bats/fb-test-foreman-templates.bats
+++ b/bats/fb-test-foreman-templates.bats
@@ -4,7 +4,6 @@
 set -o pipefail
 
 @test "export templates to a temp folder" {
-  export_dir=$(mktemp --directory)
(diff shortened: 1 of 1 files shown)


ekohl (approved):
So hammer export-templates lets the server write to some directory? What kind of weird API? If you use a load balanced server you have no idea where it's going to end up. I also wonder if there's some hidden exploit there.

Anyway, at least this tests the functionality now.

evgeni:
> So hammer export-templates lets the server write to some directory? What kind of weird API? If you use a load balanced server you have no idea where it's going to end up. I also wonder if there's some hidden exploit there.

the default modus operandi is to configure a remote git repo and let the plugin push there, but that seemed like a rather complicated setup for a simple test.

PR #1311, opened 2021-03-03
remove option --ostree enable

rabajaj0509:
`"--katello-enable-ostree true"` this option does not exists.

I get error as:
```
TASK [foreman_installer : Run installer] ***************************************
fatal: [centos7-luna-devel]: FAILED! => changed=true 
  cmd: |-
    foreman-installer -v --no-colors --disable-system-checks --scenario katello-devel --katello-devel-github-username 'rabajaj0509' --ka
tello-devel-enable-ostree=true --katello-devel-extra-plugins theforeman/foreman_remote_execution --katello-devel-extra-plugins theforema
n/foreman_discovery --katello-devel-extra-plugins theforeman/foreman_ansible --katello-devel-extra-plugins theforeman/foreman-tasks --ka
tello-devel-extra-plugins theforeman/foreman_bootdisk --katello-devel-extra-plugins theforeman/foreman_openscap --katello-devel-extra-pl
ugins theforeman/foreman_templates --katello-devel-extra-plugins theforeman/foreman_hooks --katello-devel-scl-ruby=rh-ruby25 --katello-d
evel-admin-password changeme --katello-devel-extra-plugins theforeman/foreman_remote_execution
  delta: '0:00:08.423099'
  end: '2021-03-03 13:48:12.373727'
  msg: non-zero return code
  rc: 1
  start: '2021-03-03 13:48:03.950628'
  stderr: |-
    ERROR: Unrecognised option '--katello-devel-enable-ostree'
  

```

### DIFF
--- a/pipelines/vars/forklift_katello.yml
+++ b/pipelines/vars/forklift_katello.yml
@@ -4,7 +4,7 @@ server_box:
   ansible:
     variables:
       foreman_server_repositories_katello: true
-      foreman_server_repositories_ostree: true
+      foreman_server_repositories_ostree: false
       foreman_installer_disable_system_checks: true
       foreman_installer_scenario: katello
       foreman_installer_additional_packages:
@@ -21,7 +21,7 @@ proxy_box:
   ansible:
     variables:
       foreman_server_repositories_katello: true
-      foreman_server_repositories_ostree: true
+      foreman_server_repositories_ostree: false
       foreman_installer_disable_system_checks: true
       foreman_installer_scenario: foreman-proxy-content
       foreman_installer_additional_packages:
--- a/playbooks/bats_pipeline_katello_nightly.yml
+++ b/playbooks/bats_pipeline_katello_nightly.yml
@@ -12,7 +12,6 @@
     foreman_installer_disable_system_checks: true
     foreman_installer_options_internal_use_only:
       - "--foreman-initial-admin-password {{ foreman_installer_admin_password }}"
-      - "--katello-enable-ostree=true"
       - "--puppet-server-max-active-instances 1"
       - "--puppet-server-jvm-min-heap-size 1G"
       - "--puppet-server-jvm-max-heap-size 1G"
--- a/playbooks/luna-devel.yml
+++ b/playbooks/luna-devel.yml
@@ -3,7 +3,6 @@
   vars:
     foreman_installer_options:
       - "--katello-devel-github-username '{{ katello_devel_github_username }}'"
-      - "--katello-devel-enable-ostree=true"
       - "--katello-devel-extra-plugins theforeman/foreman_remote_execution"
       - "--katello-devel-extra-plugins theforeman/foreman_discovery"
       - "--katello-devel-extra-plugins theforeman/foreman_ansible"
--- a/pipelines/vars/forklift_luna.yml
+++ b/pipelines/vars/forklift_luna.yml
@@ -4,7 +4,7 @@ server_box:
   ansible:
     variables:
       foreman_server_repositories_katello: true
-      foreman_server_repositories_ostree: true
+      foreman_server_repositories_ostree: false
       foreman_installer_disable_system_checks: true
       foreman_installer_scenario: katello
       foreman_installer_additional_packages:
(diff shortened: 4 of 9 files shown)


playbooks/luna-devel.yml
```diff
@@ -3,12 +3,11 @@
...
-      - "--katello-devel-enable-ostree=true"
-      - "--katello-devel-extra-plugins theforeman/foreman_remote_execution"
-      - "--katello-devel-extra-plugins theforeman/foreman_discovery"
-      - "--katello-devel-extra-plugins theforeman/foreman_ansible"
-      - "--katello-devel-extra-plugins theforeman/foreman-tasks"
-      - "--katello-devel-extra-plugins theforeman/foreman_bootdisk"
-      - "--katello-devel-extra-plugins theforeman/foreman_openscap"
-      - "--katello-devel-extra-plugins theforeman/foreman_templates"
-      - "--katello-devel-extra-plugins theforeman/foreman_hooks"
+      - '--katello-devel-extra-plugins theforeman/foreman_remote_execution'
```
ekohl, line 6: Why are the different quotes needed?

rabajaj0509: blame my editor! it works fine with single quotes too, so i did not mind it. If you insist i can change it back :)

rabajaj0509: I see in github action, it says syntax error

rabajaj0509:
Would you prefer to remove this option from other files too?

ekohl (commented):
Yes, please also remove it from `playbooks/luna-devel.yml`.

For the others I'm not sure how much we want to keep it enabled on Katello 3.x releases.

Also, we can set `foreman_server_repositories_ostree` to false in most places (or stop including `ostree_repositories`).

pipelines/vars/forklift_katello.yml
```diff
@@ -4,7 +4,7 @@ server_box:
   ansible:
     variables:
       foreman_server_repositories_katello: true
-      foreman_server_repositories_ostree: true
+      foreman_server_repositories_ostree: false
```
ekohl, line 7: I am worried this will break older pipelines

playbooks/bats_pipeline_katello_nightly.yml
```diff
@@ -12,7 +12,7 @@
     foreman_installer_disable_system_checks: true
     foreman_installer_options_internal_use_only:
       - "--foreman-initial-admin-password {{ foreman_installer_admin_password }}"
-      - "--katello-enable-ostree=true"
+      - "--katello-enable-ostree=false"
```
ekohl, line 15: This parameter no longer exists. I wonder if this playbook is still used.

ekohl (commented):
@evgeni could you also have a look?

pipelines/vars/forklift_katello.yml
evgeni: We don't run any ostree *tests*, so probably not. But also enabling those repos on current nightly has no harm, so I'd leave that `true` for now?

playbooks/bats_pipeline_katello_nightly.yml
evgeni: *probably* not.

`centos7-bats` box refers to `playbooks/bats_pipeline_nightly.yml` which doesn't exist, and `centos7-katello-bats-ci` (in ` 99-local.yaml.example`) refers to this one, but I don't think anyone uses those boxes today?

rabajaj0509: Let keep this PR to remove the ostree parameter and I will create another PR to remove that playbook if required.

rabajaj0509:
@evgeni @ekohl any further thoughts on this PR?

evgeni:
there are more references, should these be killed/false'd too?
```
% git grep 'ostree.*true'
pipelines/vars/forklift_katello.yml:      foreman_server_repositories_ostree: true
pipelines/vars/forklift_katello.yml:      foreman_server_repositories_ostree: true
playbooks/foreman_proxy_content.yml:        - '--foreman-proxy-content-enable-ostree true'
playbooks/foreman_proxy_content_dev.yml:    foreman_server_repositories_ostree: true
playbooks/foreman_proxy_content_dev.yml:        - '--foreman-proxy-content-enable-ostree true'
```

ehelms (approved):

PR #1313, opened 2021-03-08
Use postgresql_upgrade role from foreman-operations-collection for Fo…

ehelms:
…reman upgrades

```
PLAY RECAP *******************************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=5    changed=2    unreachable=0    failed=0    skipped=2    rescued=0    ignored=0   
pipe-up-foreman-nightly-centos8 : ok=134  changed=37   unreachable=0    failed=0    skipped=87   rescued=0    ignored=2   
pipe-up-foreman-smoker-nightly-centos8 : ok=18   changed=13   unreachable=0    failed=0    skipped=2    rescued=0    ignored=0   

ehelms@wareagle forklift (use-puppet-upgrade-foc)$ vagrant ssh pipe-up-foreman-nightly-centos8
sudo su
rpm -q postgresql*******************************************************************************
This is a private virtual machine that is the property of the Foreman project.
It is for authorized use only. Users (authorized or unauthorized) have no
explicit or implicit expectation of privacy.

Any or all uses of this system and all files on this system may be monitored,
recorded, copied, audited, or inspected.

LOG OFF IMMEDIATELY if you do not agree to the conditions stated in this
warning.
*******************************************************************************

CentOS 8.3.2011 x86_64

FQDN:      pipe-up-foreman-nightly-centos8.wareagle.example.com (192.168.122.235)
Processor: 2x Intel Core Processor (Skylake, IBRS)
Kernel:    4.18.0-240.1.1.el8_3.x86_64
Memory:    4.40 GiB


Last login: Mon Mar  8 18:43:52 2021 from 192.168.122.1
sudo su
rpm -q postgresql-[vagrant@pipe-up-foreman-nightly-centos8 ~]$ sudo su
s[root@pipe-up-foreman-nightly-centos8 vagrant]# rpm -q postgresql-server
postgresql-server-12.5-1.module_el8.3.0+620+f1abd701.x86_64
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Forklift provides tools to create Foreman/Katello environments for development, 
    - [Adding Custom Boxes](#adding-custom-boxes)
    - [Customize Deployment Settings](#customize-deployment-settings)
    - [Post Install Playbooks](#post-install-playbooks)
+   - [Using Local Ansible Collection](#using-local-ansible-collection)
  * [Production Environments](docs/production.md)
  * [Development Environments](docs/development.md)
  * [Stable Boxes](docs/stable_boxes.md)
@@ -25,7 +26,7 @@ Forklift provides tools to create Foreman/Katello environments for development, 
 ### Requirements
 
 * Vagrant - 2.2+ - Both the VirtualBox and Libvirt providers are tested
-* Ansible - 2.7+
+* Ansible - 2.9+
 * [Vagrant Libvirt provider plugin](https://github.com/vagrant-libvirt/vagrant-libvirt) (if using Libvirt)
 * Virtualization enabled in BIOS
 
@@ -38,6 +39,7 @@ This will walk through the simplest path of spinning up a production test enviro
 ```
 git clone https://github.com/theforeman/forklift.git
 cd forklift
+ansible-galaxy collection install -r requirements.yml
 vagrant up centos7-foreman-nightly
 ```
 
@@ -280,4 +282,20 @@ ansible:
     playbook:
       - 'user_playbooks/vim.yml'
       - 'user_playbooks/zsh.yml'
+```
+
+### Using Local Ansible Collection
+
+If needing to use a local copy of an Ansible collection used by Forklift, such as developing updates to theforeman.operations collection, you can temporarily update the `requirements.yml` to point at your local checkout:
+
+```
+collections:
+  - name: git+file:///home/user/path/to/repo/.git
+    type: git
+```
+
+Then run `ansible-galaxy` install:
+
+```
+ansible-galaxy collection install -r requirements.yml --force
 ```
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,0 +1,2 @@
+collections:
+  - theforeman.operations
--- a/roles/foreman_installer/tasks/upgrade.yml
+++ b/roles/foreman_installer/tasks/upgrade.yml
@@ -7,6 +7,13 @@
   import_role:
     name: update_os_packages
 
+- name: 'Upgrade postgresql'
+  import_role:
+    name: theforeman.operations.postgresql_upgrade
+  when:
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version == '8'
+
 - include_tasks: installer_version.yml
 
 - name: 'Set internal installer options'
--- a/playbooks/setup_forklift.yml
+++ b/playbooks/setup_forklift.yml
@@ -14,6 +14,7 @@
     forklift_url: "https://github.com/theforeman/forklift"
     forklift_dest: "{{ ansible_env.HOME }}/forklift"
     forklift_version: master
+    forklift_install_from_galaxy: True
     forklift_install_pulp_from_galaxy: False
     forklift_config:
       libvirt_options:
@@ -44,6 +45,12 @@
         dest: "{{ forklift_dest }}/vagrant/settings.yaml"
 
     - name: 'Install Forklift collection dependencies'
+      command:
+        cmd: ansible-galaxy collection install -r requirements.yml
+        chdir: "{{ forklift_dest }}"
+      when: forklift_install_from_galaxy
+
+    - name: 'Install Forklift Pulp collection dependencies'
       command:
         cmd: ansible-galaxy collection install -r requirements-pulp.yml
         chdir: "{{ forklift_dest }}"


requirements.yml
```diff
@@ -1,2 +1,5 @@
 collections:
   - pulp.pulp_installer
+  - name: https://github.com/ehelms/foreman-operations-collection
+    type: git
+    version: postgres-upgrade
```
ehelms, line 5: Needs update before merging, after the role has been merged to the collection (and released).  See https://github.com/theforeman/foreman-operations-collection/pull/21

.github/workflows/main.yml
```diff
@@ -28,6 +28,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: Set up Python
         uses: actions/setup-python@v2
+      - name: Install Ansible
+        run: pip install --upgrade git+https://github.com/ansible/ansible.git@stable-2.10
       - name: Install molecule
         run: pip install molecule
```
ekohl, line 34: Please make this a single pip install. Running pip multiple times gives no guarantees and is always the wrong thing to do. For example, `pip install molecule` can install some other version of ansible.

Also, a note why we need a git checkout is nice.

ekohl (changes requested):
Also needs to upgrade instructions. Also update CI. I still really dislike adding a dependency on Ansible collections.

ehelms:
Here I have broken out to make the Pulp piece stand-alone: https://github.com/theforeman/forklift/pull/1314

.github/workflows/main.yml
```diff
@@ -28,8 +28,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: Set up Python
         uses: actions/setup-python@v2
-      - name: Install molecule
-        run: pip install molecule
+      - name: Install molecule and Ansible
+        run: pip install molecule git+https://github.com/ansible/ansible.git@stable-2.10
```
ehelms, line 32: Here I am just picking the current highest stable version. And letting FAM take the lead on how we craft Ansible github workflows (i.e. https://github.com/theforeman/foreman-ansible-modules/blob/develop/.github/workflows/main.yml#L68)

ekohl: Can't you use this:
```suggestion
        run: pip install molecule ansible~=2.10.0
```

My reading of https://www.python.org/dev/peps/pep-0440/#compatible-release is that you'll get the latest 2.10 release from pypi. This avoids resorting to a git checkout.

pipelines/upgrade/07-server_to_final.yaml
```diff
@@ -19,4 +19,6 @@
       scenario_os: "{{ pipeline_os }}"
       scenario_version: "{{ forklift_upgrade_version_final }}"
     - foreman_server_repositories
+    - update_os_packages
```
ekohl, line 22: I think this implies we end up running OS updates twice. It's confusing to me that our `foreman_installer` role has upgrade support but this is implemented only in a pipeline.

ehelms: I can put this in the foreman_installer role. That did not occur to me previously.

.github/workflows/main.yml
ehelms: That works for me. @evgeni any reason to go one route or the other from your experience?

evgeni: Well, `ansible` on PyPI is not `ansible` in git, starting with 2.10.
When you install from git, you really get `ansible-base` (or `ansible-core`, whatever), whereas from PyPI you also get all the collections.

Given we don't *actually* care for the ansible version here, you can just fetch `ansible` for PyPI, which will give you 3.0 (with ansible-base 2.10) and everything should work?

requirements-pulp.yml
```diff
@@ -1,2 +1,4 @@
 collections:
-  - pulp.pulp_installer
+  - name: https://github.com/ehelms/foreman-operations-collection
```
evgeni, line 2: this should go to `requirements.yml` or something I guess?

ehelms: Dang me. Reading is hard on fridays.

requirements.yml
ehelms: Also I will work on a release of the collection.

requirements.yml
```diff
@@ -0,0 +1,3 @@
+collections:
+  - name: https://github.com/theforeman/foreman-operations-collection
+    type: git
```
evgeni, line 3: I think you can now point to galaxy, >= 0.3.0

roles/foreman_installer/tasks/upgrade.yml
```diff
@@ -7,6 +7,10 @@
   import_role:
     name: update_os_packages
 
+- name: 'Upgrade postgresql'
+  import_role:
+    name: theforeman.operations.postgresql_upgrade
```
evgeni, line 12:
```suggestion
  import_role:
    name: theforeman.operations.postgresql_upgrade
  when:
      - ansible_os_family == "RedHat"
      - ansible_distribution_major_version == '8'
```

Otherwise it will bomb out on Debian and EL7

evgeni:
Left a few blocking comments, but overall this is the direction we should be aiming.

I do wonder if the installation should default to `True` tho, as it's something we will need in our CI all the time (minus the pulpcore pipes, which can set it to `False`)?

ehelms:
> I do wonder if the installation should default to `True` tho, as it's something we will need in our CI all the time (minus the pulpcore pipes, which can set it to `False`)?

Works for me to make it straight forward for the main use case. I updated the jenkins job PR as well.

evgeni (approved):
ACK

@ekohl any objections besides "it uses/requires collections"?

README.md
```diff
@@ -281,3 +283,19 @@ ansible:
       - 'user_playbooks/vim.yml'
       - 'user_playbooks/zsh.yml'
 ```
+
+### Using Local Ansible Collection
```
ehelms, line 287: @ekohl Is this what you were looking for?

ehelms:
@ekohl Looking for your final input here

ekohl:
> @ekohl any objections besides "it uses/requires collections"?

No, that's still my main issue. IMHO we should still strive to use less Ansible, not more. The risk I see is that we're building a foreman-maintain 2.0. However, I'm willing to step aside and let you deal with it.

ehelms:
> No, that's still my main issue. IMHO we should still strive to use less Ansible, not more. The risk I see is that we're building a foreman-maintain 2.0. However, I'm willing to step aside and let you deal with it.

What I always find I am missing with these sort of comments is what the alternative is. If you are proposing we move away from Ansible, where should we head?

evgeni:
I think we discussed this (let installer do the pg10→12 upgrade) and ended up saying that this is a corner case (only needed upstream, only on el8, and at that time only Foreman) which we handle via docs/ansible.

Looking at https://github.com/theforeman/foreman-installer/pull/650, we wanted to pick it back to 2.4 -- to make Katello start out with 12 on el8, but never did.

PR #1314, opened 2021-03-09
Switch Pulp collection to requirements specific file

ehelms:


ehelms:
Here is breaking the Pulp collection stuff out to it's own requirements file and then updating the job to a parameter specific to that: https://github.com/theforeman/jenkins-jobs/pull/17

This is done as a stand-alone change given https://github.com/theforeman/forklift/pull/1313 and https://github.com/theforeman/forklift/pull/1289 were requested not to install Pulp collection bits alongside Foreman collection bits.

ehelms:
@evgeni thoughts on this stage as part of staging up to use FOC?

evgeni (approved):

ekohl (approved):

PR #1316, opened 2021-03-11
Pipeline playbook for running pulpcore integration tests in live mode

jjeffers:
This PR creates a pipeline playbook that will run katello pulpcore integration tests in "live" mode against the latest gems and packages,.

Test output is in the jenkins reporter format. As many possible actions are driven via ansible to minimize the complexity of associated jenkins-job definitions.


pipelines/katello_devel/03-tests.yml
```diff
@@ -0,0 +1,8 @@
+---
+- hosts:
+  - pipeline-katello-devel-centos7
+  become: yes
+  vars_files:
+  - ../vars/repos_staging.yml
+  roles:
+  - role: katello_test
```
zjhuntin, line 8: Just add the other roles before this one here.

zjhuntin (commented):
Good start!

jjeffers:
@ehelms @zjhuntin Anything else you see that needs to be changed here?

roles/katello_test/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+    - "pulp_2to3_migration_client"
+    - "pulp_certguard_client"
+
+- name: 'env'
+  ansible.builtin.shell: |
+    id
+  args:
+    executable: /bin/bash
+    chdir: "{{ foreman_development_dir }}"
+  become_user: "{{ test_user }}"
```
ehelms, line 23: What is this step doing? Looks like its just printing out user information?

roles/katello_test/vars/main.yml
```diff
@@ -0,0 +1,3 @@
+foreman_development_dir: /home/vagrant/foreman/
+test_user: vagrant
+test_vcr_mode: all
```
ehelms, line 3: Are any of these intended to be configurable by someone running this role based n their environment? If yes, I would move these to `defaults/main.yml` and then scope them on the role by prefixing `katello_test` on each.

roles/katello_test/tasks/main.yml
jjeffers: That's correct - not needed and was just for debugging purposes.

zjhuntin (approved):
I think this is good to go for an initial pipeline. Right now it fails, but that is as expected.

zjhuntin:
@ehelms did you have anything else you wanted to ask for this?

roles/katello_test/vars/main.yml
```diff
@@ -0,0 +1 @@
+
```
ehelms, line 1: Given this file is empty `vars/` directory can be dropped.

roles/katello_test/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+katello_test_foreman_development_dir: /home/vagrant/foreman/
+katello_test_test_user: vagrant
```
ehelms, line 2: This could be just `katello_test_user`

roles/katello_test/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
...
+katello_test_test_vcr_mode: all
```
ehelms, line 3: Similar, you could make this `katello_test_vcr_mode`.

pipelines/katello_devel/03-tests.yml
```diff
@@ -0,0 +1,11 @@
+---
+- hosts:
+  - pipeline-katello-devel-centos7
+  become: yes
+  vars_files:
+  - ../vars/repos_staging.yml
+  roles:
+  - ruby_scl
+  - nodejs_scl
+  - postgresql_scl
```
ehelms, line 10: In the `02-install.yaml` step, the SCL packages are already being handled by the inclusion of `katello_devel` role which additionally ensures things are handled properly on EL8 (https://github.com/theforeman/forklift/blob/master/roles/katello_devel/meta/main.yml#L2). If that works, I'd drop these call outs here and let that step handle things. Otherwise, I would go ahead and additionally scope these similar to `katello_devel` to only include these roles on El7. That way, given devel installs can be on El8 and will be there primarily in a few months this is already ready for it.

ehelms (approved):

PR #1317, opened 2021-03-12
Improvements for dev environments

cmeissner:
This PR provides some small improvements for local development environments.
  
Maybe it needs to be discussed if inline comments for ignoring issues are okay for that project.
The extension of ignore should be okay I think. I would propose to add `.ruby-version` also to ignores.
Than you should remove the existing file from repo.

### DIFF
--- a/inventories/vagrant.py
+++ b/inventories/vagrant.py
@@ -8,9 +8,9 @@ import subprocess
 import sys
 
 try:
-    from StringIO import StringIO
+    from StringIO import StringIO  # pyright: reportMissingImports=false
 except ImportError:
-    from io import StringIO
+    from io import StringIO  # pyright: reportMissingImports=false
 
 import paramiko
 
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,11 @@ packer/packer_cache/
 packer/output*
 packer/*.box
 packer/build.*
+
+# VSCode
+.vscode
+
+# local development environment
+.env
+.venv
+.python-version


ehelms:
I'm OK with the .gitignores since I know that helps users of different code editors and environments. The addition of the `pyright` ignores gives me pause, since that is not something that this project runs as part of it's automated tests.

cmeissner:
> `pyright` ignores gives me pause, since that is not something that this project runs as part of it's automated tests.

I know what you mean. The problem which it should solve is that the code in [vagrant.py](https://github.com/theforeman/forklift/blob/master/inventories/vagrant.py#L18-L21)

```
try:
    DEVNULL = subprocess.DEVNULL
except AttributeError:
    DEVNULL = open(os.devnull, 'w')
```
ever leads to import errors of pylance (e.g. in vscode). I agree that such comments for ignoring errors of linters or so always are a compromise.

ehelms:
Let's see what @evgeni or @ekohl think. Their python-fu is much, much better than mine.

ehelms:
Let's see what @evgeni or @ekohl think. Their python-fu is much, much better than mine.

inventories/vagrant.py
```diff
@@ -8,9 +8,9 @@
 import sys
 
 try:
-    from StringIO import StringIO
+    from StringIO import StringIO  # pyright: reportMissingImports=false
 except ImportError:
-    from io import StringIO
+    from io import StringIO  # pyright: reportMissingImports=false
```
ekohl, line 13: I've never used pyright so I'm just wondering. This is Python 2/3 compatibility code. If we write it the other way around and first try Python 3 (from io ..), do you still need the comment?

cmeissner: The order doesn't matter. Depending of the python version I use with the project one import will ever lead to a warning.
Tthe change is only an convenient thing. I guess it is too environment specific so I would remove this from PR. 

pyright is part of pylance which is used by vscode to check such things.

ekohl: My reasoning was that most environments will now be Python 3 so it makes sense to me to first try that. If that would also solve the import error, then I'd be happier with that. In general I'm not against linters or other static analysis but would like to have the minimum amount of control statements.

cmeissner: > If that would also solve the import error, then I'd be happier with that.

There are no import errors, only warnings. 

without comments:

image

with comments:

image

As I said, this is only a convenient change.

ekohl (approved):
I'm ok with this. If as a human you need to know which warnings in your IDE to ignore, you've lost. It's properly documented in the commit message and we know we can get rid of it once we drop Python 2 (which is dropping EL7 support, whenever that may be). Thanks!

PR #1318, opened 2021-03-12
Add docker as a provider

mikedep333:
Pulp's forklift usage has a need for this to run a Pulp 2 docker
container on top of Pulp 3. And the docker provider supports docker
boxes nested on top of VM boxes.

See https://github.com/pulp/pulp_installer/pull/543 for our example
usage.

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -88,11 +88,7 @@ module Forklift
         networks = configure_networks(box.fetch('networks', []))
         configure_shell(machine, box)
         configure_ansible(machine, box['ansible'], box['name'])
-        configure_libvirt(machine, box, networks)
-        configure_virtualbox(machine, box, networks)
-        configure_rackspace(machine, box)
-        configure_openstack_provider(machine, box)
-        configure_google_provider(machine, box)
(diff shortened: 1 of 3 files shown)


evgeni:
@mikedep333 this lookgs good to me.

Could you please fix the rubocop issue (by either bumping the max block length or ignoring that cop for that block) and then it's good to go.

vagrant/lib/forklift/box_distributor.rb
```diff
@@ -93,6 +93,7 @@ def define_vm(config, box = {})
         configure_rackspace(machine, box)
         configure_openstack_provider(machine, box)
         configure_google_provider(machine, box)
+        configure_docker_provider(machine, box)
```
ekohl, line 96: Given the whole bunch, perhaps it gets to a poin where we want to introduce `configure_providers` and then call all of these providers? Probably makes sense to still keep all the separate methods, but this method is getting very long.

evgeni: Good idea!

mikedep333: Done!

(Correctly I think, I have minimal ruby knowledge.)

ehelms (approved):

PR #1319, opened 2021-03-16
Switch ruby_scl default to rh-ruby27

ehelms:


ehelms:
This can go in now that we have finished building onto rh-ruby27.

ekohl (approved):

PR #1321, opened 2021-03-24
Fixes #32171 - consume Katello from yum.theforeman.org

evgeni:
this currently checks for nightly or >= 4.1, as older versions are not
published on yum.tfm yet.

### DIFF
--- a/roles/katello_repositories/tasks/release_repos.yml
+++ b/roles/katello_repositories/tasks/release_repos.yml
@@ -1,6 +1,6 @@
 ---
 - name: 'Setup Katello {{ katello_repositories_version }} Repository'
   yum:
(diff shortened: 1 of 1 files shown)


ekohl (approved):
There is now a 3.18. Should we manually sync in all content from 3.18 and 4.0 so there is a complete archive?

Still, :+1: for a version check so that the code can be used for any version.

evgeni:
> There is now a 3.18. Should we manually sync in all content from 3.18 and 4.0 so there is a complete archive?

I was thinking of having a "fresh cut" with 4.0 and having 3.17 (in progress) and 3.18 (done) being there just because they received z streams.

ekohl:
I can see arguments for either. All other repos have all RCs as well so it would be consistent. It would also allow you to proceed right away without having to run the pipeline.

PR #1322, opened 2021-03-24
use pulpcore repo for katello 3.17

evgeni:
this was configured in 5d9f7f2f, but the playbook was never adjusted,
thus breaking 3.17 fresh installs from the staging repo


chris1984 (approved):
ACK

PR #1324, opened 2021-03-26
Deploy smoker on EL8 to support client testing

ehelms:


roles/smoker/defaults/main.yml
```diff
@@ -9,6 +9,5 @@ smoker_command_args: "--driver chrome --variables '{{ smoker_variables_path }}'
 smoker_markers:
 # pinned to 86 as 87 hangs randomly: https://bugzilla.redhat.com/show_bug.cgi?id=1914166
 smoker_browser_packages:
-  - https://people.redhat.com/evgeni/chrome86/chromedriver-86.0.4240.193-1.el7.x86_64.rpm
-  - https://people.redhat.com/evgeni/chrome86/chromium-86.0.4240.193-1.el7.x86_64.rpm
-  - https://people.redhat.com/evgeni/chrome86/chromium-common-86.0.4240.193-1.el7.x86_64.rpm
+  - chromedriver
+  - chromium
```
ehelms, line 13: @evgeni Is there EL8 versions of the above you be up for hosting in the same way? I think I am seeing the same timeout issue using these from the repositories.

ekohl: I think we pulled the old version from Fedora's koji so I'd search there. However, https://koji.fedoraproject.org/koji/buildinfo?buildID=1641050 was deleted.

evgeni: Nope. Best thing I can offer is to pull the source from Fedora's dist git and throw it at a friendly koji that knows EL8 ;)

ekohl:
Can we revisit this now that we're dropping EL7?

evgeni:
> Can we revisit this now that we're dropping EL7?

Sure. I've rebased, fixed the conflicts and moved to c8s instead of c8, but didn't actually test any of this yet.

ekohl (approved):
I suppose we have one way to find it.

evgeni:
Well, this passed locally, so going to merge!

PR #1326, opened 2021-03-30
Run remove-pulp2 command after upgrading to 4.0

ehelms:
Given users are expected to run this at some point, and as we progress through the current Katello versions and those coming after,

### DIFF
--- a/pipelines/upgrade/05-server_to_intermediate.yaml
+++ b/pipelines/upgrade/05-server_to_intermediate.yaml
@@ -21,3 +21,11 @@
     - foreman_server_repositories
     - role: foreman_installer
       foreman_installer_skip_installer: "{{ true if forklift_upgrade_version_start == forklift_upgrade_version_intermediate else false }}"
+  post_tasks:
+    - name: Run remove Pulp 2
+      shell: yes | foreman-maintain content remove-pulp2
+      when:
+        - forklift_upgrade_version_intermediate == "4.0"
+        - pipeline_remove_pulp2
+        - ansible_os_family == "RedHat"
+        - ansible_distribution_major_version == "7"
--- a/pipelines/upgrade/07-server_to_final.yaml
+++ b/pipelines/upgrade/07-server_to_final.yaml
@@ -20,3 +20,11 @@
       scenario_version: "{{ forklift_upgrade_version_final }}"
     - foreman_server_repositories
     - foreman_installer
+  post_tasks:
+    - name: Run remove Pulp 2
+      shell: yes | foreman-maintain content remove-pulp2
+      when:
+        - forklift_upgrade_version_final == "4.0"
+        - pipeline_remove_pulp2
+        - ansible_os_family == "RedHat"
+        - ansible_distribution_major_version == "7"
--- a/pipelines/vars/forklift_luna.yml
+++ b/pipelines/vars/forklift_luna.yml
@@ -34,6 +34,7 @@ server_box:
         - "fb-test-foreman-ansible.bats"
         - "fb-test-foreman-templates.bats"
         - "fb-katello-proxy.bats"
+      pipeline_remove_pulp2: true
 proxy_box:
   box: "{{ pipeline_os }}"
   memory: 3072
--- a/pipelines/upgrade/06-proxy_to_intermediate.yaml
+++ b/pipelines/upgrade/06-proxy_to_intermediate.yaml
@@ -22,3 +22,11 @@
     - foreman_server_repositories
     - role: foreman_proxy_content
       when: forklift_upgrade_version_start != forklift_upgrade_version_intermediate
+  post_tasks:
+    - name: Run remove Pulp 2
+      shell: yes | foreman-maintain content remove-pulp2
+      when:
+        - forklift_upgrade_version_intermediate == "4.0"
+        - pipeline_remove_pulp2
+        - ansible_os_family == "RedHat"
+        - ansible_distribution_major_version == "7"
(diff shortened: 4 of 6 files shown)


evgeni:
Should this also run for luna? IMHO yes.

And currently this will run remove after the upgrade to 3.18, which means it will purge katello (as it depends on pulp-selinux in 3.18). I guess it should check whether the intermediate/final version is 4.0 instead?

pipelines/upgrade/05-server_to_intermediate.yaml
```diff
@@ -21,3 +21,11 @@
     - foreman_server_repositories
     - role: foreman_installer
       foreman_installer_skip_installer: "{{ true if forklift_upgrade_version_start == forklift_upgrade_version_intermediate else false }}"
+  post_tasks:
+    - name: Run remove Pulp 2
+      command: foreman-maintain content remove-pulp2
+      when:
+        - pipeline_version == "4.0"
```
ekohl, line 28: Shouldn't it check the version that's upgraded to instead?
```suggestion
        -  forklift_upgrade_version_intermediate== "4.0"
```

pipelines/upgrade/07-server_to_final.yaml
```diff
@@ -20,3 +20,11 @@
       scenario_version: "{{ forklift_upgrade_version_final }}"
     - foreman_server_repositories
     - foreman_installer
+  post_tasks:
+    - name: Run remove Pulp 2
+      command: foreman-maintain content remove-pulp2
+      when:
+        - pipeline_version == "4.0"
```
ekohl, line 27:
```suggestion
        - forklift_upgrade_version_final == "4.0"
```

ehelms:
> Should this also run for luna? IMHO yes.

Done

pipelines/upgrade/05-server_to_intermediate.yaml
```diff
@@ -21,3 +21,11 @@
     - foreman_server_repositories
     - role: foreman_installer
       foreman_installer_skip_installer: "{{ true if forklift_upgrade_version_start == forklift_upgrade_version_intermediate else false }}"
+  post_tasks:
+    - name: Run remove Pulp 2
+      command: foreman-maintain content remove-pulp2
+      when:
+        - forklift_upgrade_version_intermediate == "4.0"
+        - pipeline_type == "katello" or pipeline_type == "luna"
```
ekohl, line 29: Generally not a huge fan of hardcoding pipeline type names. Would it be better to set a variable in a vars file about this?

pipelines/upgrade/07-server_to_final.yaml
```diff
@@ -20,3 +20,11 @@
       scenario_version: "{{ forklift_upgrade_version_final }}"
     - foreman_server_repositories
     - foreman_installer
+  post_tasks:
+    - name: Run remove Pulp 2
+      command: foreman-maintain content remove-pulp2
```
jlsherrill, line 25: i think this will stop and ask if you want to proceed, in retrospect we should have added a --force option

Maybe you force some input into it?

ehelms: Ahh yea, I see that here -- https://github.com/theforeman/foreman_maintain/pull/442/files#diff-69a7be004ceb14fd5a9c0c117906b77e5dc900604a03df595ae737d2484b9bfbR42

I will test a way around this to run it hands off.

ekohl: Should it be `shell: yes | foreman-maintain content remove-pulp2` while we file an issue for a `--yes` or `--force`? (I feel like `--force` may be too strong for just a "do not ask for confirmation").

pipelines/vars/forklift_luna.yml
```diff
@@ -34,6 +34,7 @@ server_box:
         - "fb-test-foreman-ansible.bats"
         - "fb-test-foreman-templates.bats"
         - "fb-katello-proxy.bats"
+      pipeline_remove_pulp2: true
```
ekohl, line 37: Should we also run removal on proxy boxes?

ehelms: I debated it, @jlsherrill thoughts?

pipelines/upgrade/07-server_to_final.yaml
evgeni: Yeah, I too hate it when `--force` means `--assume-yes` (or similar).

jlsherrill: yes, i'd prefer --assume-yes, more meant a general option

pipelines/vars/forklift_luna.yml
jlsherrill: probably makes sense to or else foreman-maintain will fail there too, without https://github.com/theforeman/foreman_maintain/pull/464/files

ehelms: I don't think our pipelines run foreman-maintain on the content proxies at the moment. I can run a test with it and see.

pipelines/upgrade/05-server_to_intermediate.yaml
```diff
@@ -21,3 +21,11 @@
     - foreman_server_repositories
     - role: foreman_installer
       foreman_installer_skip_installer: "{{ true if forklift_upgrade_version_start == forklift_upgrade_version_intermediate else false }}"
+  post_tasks:
+    - name: Run remove Pulp 2
+      shell: yes | foreman-maintain content remove-pulp2
```
ekohl, line 26: I've known about this pattern for a long time but I think this is the first time I see it in action :)

ekohl (approved):
We may need to revisit this if we start to add content migration tests but if this gets our pipelines unstuck then I'm all for it.

PR #1329, opened 2021-03-31
properly use the pulp2 removal variable

evgeni:


pipelines/upgrade/06-proxy_to_intermediate.yaml
```diff
@@ -25,4 +26,5 @@
     - role: remove_pulp2
       when:
         - forklift_upgrade_version_intermediate == "4.0"
+        - pipeline_remove_pulp2 is defined
```
ehelms, line 29: Are these evaluated in order every time? or is it safer to combine this with boolean check?

evgeni: in order.

ehelms (approved):

PR #1330, opened 2021-04-05
Ensure no rh-ruby25 packages in Foreman 2.5

ehelms:
Once Ruby 2.7 building is complete this can be merged

### DIFF
--- a/bats/fb-verify-packages.bats
+++ b/bats/fb-verify-packages.bats
@@ -6,15 +6,34 @@ set -o pipefail
 load os_helper
 load foreman_helper
 
-@test "verify no old SCL packages" {
+@test "verify no rh-ruby24, tfm-ror51 or tfm-ror52 SCL packages" {
   tSetOSVersion
-  FOREMAN_VERSION=$(tForemanVersion)
+  if ! tIsRHEL 7; then
+    skip "SCL package checks only applicable on EL 7 systems"
+  fi
 
-  if tIsRedHatCompatible; then
-    if [[ $FOREMAN_VERSION == 2.* ]]; then
-      if rpm --query --all | grep --quiet --extended-regexp 'rh-ruby24|tfm-ror51|tfm-ror52'; then
-        exit 1
-      fi
-    fi
+  FOREMAN_VERSION=$(tForemanVersion)
+  if [[ $FOREMAN_VERSION != 2.* ]]; then
+    skip "Verification of no rh-ruby24, tfm-ror51 or tfm-ror52 packages only applies to Foreman 2.*"
+  fi
+
+  if rpm --query --all | grep --quiet --extended-regexp 'rh-ruby24|tfm-ror51|tfm-ror52'; then
+    exit 1
+  fi
+}
+
+@test "verify no rh-ruby25 SCL packages in Foreman 2.5" {
+  tSetOSVersion
+  if ! tIsRHEL 7; then
+    skip "SCL package checks only applicable on EL 7 systems"
+  fi
+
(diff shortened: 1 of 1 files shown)


bats/fb-verify-packages.bats
```diff
@@ -16,5 +16,10 @@ load foreman_helper
         exit 1
       fi
     fi
+    if [[ $FOREMAN_VERSION == 2.5* ]]; then
+      if rpm --query --all | grep --quiet --extended-regexp 'rh-ruby25'; then
```
ekohl, line 20: Why is `--extended-regexp` needed if it's not a regexp at all?

ekohl (commented):
With 2 ways it could fail, I wonder if we should somehow split this or make it possible to know where it failed. Maybe a different exit code would be sufficient?

bats/fb-verify-packages.bats
ehelms: Copy and paste from above since I know that works and does not hurt anything.

ehelms:
> With 2 ways it could fail, I wonder if we should somehow split this or make it possible to know where it failed. Maybe a different exit code would be sufficient?

How about this version where there are two separate tests and skips implemented to give more output as to when a particular check should have run or that it skipped due to a condition not being met.

evgeni:
Technically this will skip the test again when we bump to 2.6 or later. Comparing versions in bash/bats is a pain, so I am totally fine with taking it as-is and adding a backlog item, but we shouldn't forget (just for the sake of not regressing in the future).

ekohl:
> How about this version where there are two separate tests and skips implemented to give more output as to when a particular check should have run or that it skipped due to a condition not being met.

I also thought about that. I'm fine with either solution but to me the important part is that when the test fails you can figure out why it failed.

bats/fb-verify-packages.bats
```diff
@@ -6,15 +6,34 @@ set -o pipefail
...
+}
+
+@test "verify no rh-ruby25 SCL packages in Foreman 2.5" {
+  tSetOSVersion
+  if ! tIsRedHatCompatible; then
+    skip "SCL package checks only applicable on EL systems"
+  fi
+
   FOREMAN_VERSION=$(tForemanVersion)
+  if [[ $FOREMAN_VERSION != 2.5* ]]; then
```
ekohl, line 32: I think @evgeni's comment can be addressed by such a check. It can check Foreman 2.5+
```suggestion
  if [[ $FOREMAN_VERSION != 2.[5-9]* ]]; then
```

bats/fb-verify-packages.bats
```diff
@@ -6,15 +6,34 @@ set -o pipefail
 load os_helper
 load foreman_helper
 
-@test "verify no old SCL packages" {
+@test "verify no rh-ruby24, tfm-ror51 or tfm-ror52 SCL packages" {
   tSetOSVersion
+  if ! tIsRedHatCompatible; then
+    skip "SCL package checks only applicable on EL systems"
```
ekohl, line 12: Should this also check if it's EL7? It's not really relevant on EL8.

ehelms: We didn't have that before, but since I am re-factoring I can make that additional update.

bats/fb-verify-packages.bats
ehelms: I can try a few conditions here.

ehelms:
I added a check to skip if not EL7 now.

ekohl (approved):

PR #1331, opened 2021-04-06
Run installer with enable katello-agent flags for Katello 4.1+

ehelms:


bats/fb-katello-client.bats
```diff
@@ -78,6 +78,21 @@ load fixtures/content
...
+@test "try fetching container content" {
+  tPackageExists podman || tPackageInstall podman
+  podman login $HOSTNAME -u admin -p changeme
+  CONTAINER_PULL_LABEL=`echo "${ORGANIZATION_LABEL}-${PRODUCT_LABEL}-${CONTAINER_REPOSITORY_LABEL}"| tr '[:upper:]' '[:lower:]'`
+  podman pull "${HOSTNAME}/${CONTAINER_PULL_LABEL}"
+}
+
+@test "enable katello-agent" {
+  KATELLO_VERSION=$(tKatelloVersion)
+  if [[ $KATELLO_VERSION >= 4.1 ]]; then
```
evgeni, line 90: I don't think that's a valid test expression.

bats/fb-katello-client.bats
```diff
@@ -78,6 +78,21 @@ load fixtures/content
   done
 }
 
+@test "try fetching container content" {
+  tPackageExists podman || tPackageInstall podman
+  podman login $HOSTNAME -u admin -p changeme
+  CONTAINER_PULL_LABEL=`echo "${ORGANIZATION_LABEL}-${PRODUCT_LABEL}-${CONTAINER_REPOSITORY_LABEL}"| tr '[:upper:]' '[:lower:]'`
```
evgeni, line 84: unrelated to this PR, but it feels like this should live in https://github.com/theforeman/forklift/blob/master/bats/fixtures/content.bash?

evgeni (approved):

bats/fb-katello-client.bats
```diff
@@ -78,6 +78,21 @@ load fixtures/content
...
+  podman pull "${HOSTNAME}/${CONTAINER_PULL_LABEL}"
+}
+
+@test "enable katello-agent" {
+  KATELLO_VERSION=$(tKatelloVersion)
+  if [[ $KATELLO_VERSION != 4.[1-9]* ]]; then
+    skip "Enabling katello-agent explicitly is only available with Katello 4.1+"
+  fi
+  foreman-installer --katello-enable-katello-agent true --foreman-proxy-content-enable-katello-agent true
```
ekohl, line 93: Just a thought about how we expose it (https://github.com/theforeman/puppet-katello/pull/387): we can not expose it in init.pp but rather in `katello::params`. Then in the Hiera config we set:
```yaml
---
katello::params::enable_katello_agent: "%{lookup('foreman_proxy_content::enable_katello_agent')}"
```

The result would be that there's a single parameter that works both for all scenarios: `--foreman-proxy-content-enable-katello-agent true`. Now that's my theory and I haven't tested it, but it's a path I'd like to investigate.

evgeni: Semi related: can we do the same for all plugins? :)

ehelms: I can certainly give it a try.

ehelms: I gave this a try and am running into the following:

```
Error: Evaluation Error: Error while evaluating a Function Call, Class[Katello::Params]: parameter 'enable_katello_agent' expects a Boolean value, got String (line: 120, column: 346) on node pipe-katello-server-nightly-centos7.wareagle.example.com
```

With this defined:

```
katello::params::enable_katello_agent: lookup('foreman_proxy_content::enable_katello_agent')
```

ehelms: I read that this issue might be related to how yaml format for Hiera handle boolean types (or lack thereof).

ekohl: https://puppet.com/docs/puppet/5.5/hiera_automatic.html#arguments suggests a second argument can be passed to lookup. That determines the data type. Haven't tried it myself.

PR #1332, opened 2021-04-06
Enable Ruby 2.7 for nightly and Foreman 2.5+ before installing installer

ehelms:


roles/foreman_installer/tasks/packages.yml
```diff
@@ -1,4 +1,13 @@
 ---
+- name: Enable ruby:2.7 module
+  dnf:
+    name: '@ruby:2.7'
+    state: present
+  when:
+    - foreman_repositories_version == "nightly" or 2.5 is version(foreman_repositories_version, '>')
```
evgeni, line 7: This tests if 2.5 > f_r_v, which is not what you want ;)
`foreman_repositories_version is version ('2.5', '>=')` is probably more what's wanted here?

evgeni (changes requested):

evgeni:
Won't this break updates as those already have `ruby:2.5` enabled?

ehelms:
> Won't this break updates as those already have `ruby:2.5` enabled?

My testing of upgrades passed locally. That is because as far as I can tell, the way we are using the dnf Ansible module it performs a reset on the ruby module and then an enable on the 2.7 Ruby profile.

evgeni (approved):

PR #1333, opened 2021-04-07
Only run fb-verify-packages on install

ekohl:
In upgrades old packages can still be present. That's why we only test it on fresh installs. Additionally, it now also runs it on the plugins pipeline.


evgeni:
#1334 was the better solution

PR #1334, opened 2021-04-07
Cleanup rh-ruby25 RPMs on upgrade for nightly and Foreman 2.5+

ehelms:


roles/cleanup_rh_ruby25/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Remove rh-ruby25 packages
+  package:
+    state: absent
+    name: rh-ruby25*
```
ekohl, line 5: You can optimize a bit by adding `when EL7` to it.

ekohl (commented):
I'm wondering if it makes sense to have some `post_upgrade` role with multiple tasks. Does that make sense?

ehelms:
> I'm wondering if it makes sense to have some `post_upgrade` role with multiple tasks. Does that make sense?

I started down this approach but then realized that the tasks are often version specific, so you have to start burying that version info into the module, or through multiple parameters per item and it could spiral out of hand.

roles/cleanup_rh_ruby25/tasks/main.yml
ehelms: Done.

evgeni:
I think nightly will never be intermediate ;)

ehelms:
> I think nightly will never be intermediate ;)

Touche sir, touche.

ekohl (approved):

PR #1336, opened 2021-04-12
add ubuntu 20.04 box

evgeni:


vagrant/boxes.d/00-ubuntu.yaml
```diff
@@ -17,3 +17,20 @@ boxes:
     ansible:
       variables:
         ansible_python_interpreter: /usr/bin/python3
+  ubuntu2004:
+    box_name: 'generic/ubuntu2004'
+    pty: true
+    scenarios:
+      - foreman
+    shell: |
+      # Work around external nameservers in generic/ubuntu2004
```
ekohl, line 26: If only Ubuntu built libvirt boxes, then we wouldn't need this workaround :(

ekohl (approved):

vagrant/boxes.d/00-ubuntu.yaml
evgeni: exactly my thoughts earlier today.

PR #1338, opened 2021-04-15
properly parse centos-stream version

evgeni:


ekohl (commented):
Makes me wonder if we should source `/etc/os-release` and use that:
```bash
tSetOSVersion() {
  if [[ -z "$OS_VERSION" ]]; then
    . /etc/os-release
    OS_VERSION=$VERSION_ID
  fi
}
```

evgeni:
Maybe? You'd still need the `grep -o '^[0-9]*'` part, as on RHEL VERSION_ID is actually of the form X.Y, not X.

evgeni:
Let's get the known working fix in and then see how we refactor it?

ekohl (approved):
Fair enough.

PR #1342, opened 2021-04-21
Group plugins in Luna scenario

ekohl:
This groups all options by plugin rather than just sorting alphabetically. This makes it easier to see which plugins are enabled.

### DIFF
--- a/pipelines/vars/forklift_luna.yml
+++ b/pipelines/vars/forklift_luna.yml
@@ -13,26 +13,36 @@ server_box:
         - "--puppet-server-max-active-instances 1"
         - "--puppet-server-jvm-min-heap-size 1G"
         - "--puppet-server-jvm-max-heap-size 1G"
+
         - "--enable-foreman-cli-ansible"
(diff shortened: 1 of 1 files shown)


evgeni (approved):

pipelines/vars/forklift_luna.yml
```diff
@@ -13,26 +13,36 @@ server_box:
...
-        - "--enable-foreman-proxy-plugin-discovery"
         - "--enable-foreman-proxy-plugin-openscap"
+
+        - "--enable-foreman-cli-remote-execution"
+        - "--enable-foreman-plugin-remote-execution"
         - "--enable-foreman-proxy-plugin-remote-execution-ssh"
         - "--foreman-proxy-plugin-remote-execution-ssh-install-key=true"
+
+        - "--enable-foreman-cli-tasks"
+        - "--enable-foreman-plugin-tasks"
```
ekohl, line 39: I added this line. Previously it was enabled implicitly.

PR #1343, opened 2021-04-22
do not enable the ansible repo on el7, use epel

jlsherrill:


ekohl (approved):
Result of discussion in https://github.com/theforeman/pulpcore-packaging/issues/138

PR #1344, opened 2021-04-22
Checkout puppet-katello_devel from versioned branches

jlsherrill:


jlsherrill:
the idea is that we will branch puppet-katello_devel when we branch katello itself, and thus the devel box can still be installed weeks later when GA is released.

jturel (approved):
looks like it does what we  discussed

ekohl:
I'm not strictly opposed to this, but I think it's incomplete.

For starters, how do you plan to deal with plugin versions? You can't just checkout REX master on an older Katello and expect it to work.

We'll also need to deal with module dependencies and test fixtures. Currently it checks out all development versions of all its dependencies.

In the (very) distant past we used to include the devel module in the installer itself. I feel like that may not have been the worst idea ever.

jlsherrill:
@ekohl it looks like the dev install does a checkout of puppet-katello_devel  and then it installs foreman-installer from rpm which should pull in the correct dependencies (although its possible i'm missing something)? 

For plugins, these get installed via the 'bundle install' as a gem.  I do imagine that this change won't solve all problems and enable a dev install to be done ~6months after release, but as part of the release process we are building stable boxes of each release, and this ideally helps us make it between branch time and release time so that we can build a snapshotted devel box at release time

ekohl:
> @ekohl it looks like the dev install does a checkout of puppet-katello_devel and then it installs foreman-installer from rpm which should pull in the correct dependencies (although its possible i'm missing something)?

I'm talking about CI in https://github.com/theforeman/puppet-katello_devel. https://github.com/theforeman/puppet-katello_devel/blob/master/.fixtures.yml just pulls in all the latest versions. If you do need to make a change to a stable branch, it's nice to have the tests work.

> For plugins, these get installed via the 'bundle install' as a gem. I do imagine that this change won't solve all problems and enable a dev install to be done ~6months after release, but as part of the release process we are building stable boxes of each release, and this ideally helps us make it between branch time and release time so that we can build a snapshotted devel box at release time

I was thinking about [katello_devel::plugin](https://github.com/theforeman/puppet-katello_devel/blob/master/manifests/plugin.pp). I think today we don't use the `extra_plugins` parameter, but you would need to support the branch name there too. A simple name won't suffice anymore.

jlsherrill:
Thanks @ekohl for the explanation, its good to know the limitations of this.  I think i'm okay merging as is?  If we run into a situation where we need a backport to fix something, we could always backport the versions https://github.com/theforeman/foreman-installer/blob/2.4-stable/Puppetfile.lock to the .fixtures.yml directory.  

For the specific case i mentioned (needing to build a snapshot of a devel box ~4 weeks after branching), i think this will likely solve our issues most of the time.

ehelms (approved):

PR #1346, opened 2021-04-28
Create compute-profile without taxonomies

stejskalleos:


ekohl (commented):
What is the reasoning behind this?

stejskalleos:
`hammer compute-profile create` accept only name:

```
hammer compute-profile create -h
Usage:
    hammer compute-profile create [OPTIONS]

Options:
 --name NAME                    
 -h, --help                    Print help
```

ekohl (changes requested):
That sounds like a Hammer or Foreman bug since https://theforeman.org/api/2.4/apidoc/v2/compute_profiles/create.html documents that it should accept the taxonomies.

stejskalleos:
> That sounds like a Hammer or Foreman bug since https://theforeman.org/api/2.4/apidoc/v2/compute_profiles/create.html documents that it should accept the taxonomies.

Yes, it have been removed from Hammer in this [PR](https://github.com/theforeman/hammer-cli-foreman/pull/551/files).
It seems that `Compute profile` resource is not scoped to taxonomies, at least when you create CP in UI taxonomies have no effect at all. So it is probably bug in the APIpie documentation.

ekohl:
>  It seems that `Compute profile` resource is not scoped to taxonomies, at least when you create CP in UI taxonomies have no effect at all. So it is probably bug in the APIpie documentation.

That's surprising because the model does include Taxonomix: https://github.com/theforeman/foreman/blob/2b5ff89b21f797c37ca2bd73c941ce3dc677de03/app/models/compute_resource.rb#L3. It clearly was intended to be taxable and this appears to be a Hammer regression. I'm inclined to not merge this but rather address the root cause.

ekohl (approved):
I clearly needed a bit more caffeine when I made that comment. Compute resource != compute profile. Profile is indeed untaxable:
https://github.com/theforeman/foreman/blob/2b5ff89b21f797c37ca2bd73c941ce3dc677de03/app/models/compute_profile.rb

Per https://github.com/theforeman/hammer-cli-foreman/pull/551#issuecomment-831850979 the API should get fixed and this PR is correct.

ekohl:
(just letting the CI tasks run, otherwise :+1:)

PR #1352, opened 2021-05-06
Unlock packages after enabling katello-agent

ehelms:
If the installer scenario is using the packages lock feature the
subsequent tests will fail as package installs are blocked.


pcreech (approved):
ACK

PR #1353, opened 2021-05-06
Explicitly pass kwargs to support Ruby 3

jeremylenz:
Due to [breaking changes in Ruby 3](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/), keyword args must now be passed explicitly with `**` and can't just be passed as a hash in the last arg.

This will be needed to run forklift on Fedora 34, since system ruby is Ruby 3.

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -169,6 +169,7 @@ module Forklift
       end
     end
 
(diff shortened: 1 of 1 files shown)


evgeni (approved):

ekohl:
Doesn't this break on EL7's Ruby 2.0?

evgeni:
We'll see? ;)

PR #1354, opened 2021-05-07
test on more rubies

evgeni:


ekohl (approved):
Technically we should also test on EL7's Ruby 2.0 since that's we use in CI but this is a good start.

PR #1360, opened 2021-05-15
use hammer puppet-environment

evgeni:
hammer environment was deprecated since hammer 0.18 (Foreman 1.23) and
is now removed in nightly


wbclark (approved):
nice-to-have change would be, specify in the commit message that this change is for bats test

overall it looks very straightforward. LGTM

PR #1361, opened 2021-05-17
Enable katello-agent at the start of client tests

ehelms:
In some test environments, the packages needed to enable katello-agent
may not be available after re-registering the host to Foreman.


ehelms:
@evgeni @zjhuntin I think this is an easy change to unblock certain test environments. And we can consider a different long term solution later.

evgeni:
I'm happy with whatever unblocks the disconnected test setup.

zjhuntin (approved):
ACK

PR #1364, opened 2021-05-25
[TEMP] pull boxes directly from cloud.centos.org

evgeni:
vagrantcloud.com is not redirecting properly and thus downloads fail


ekohl (approved):

evgeni:
(replaced URLs with HTTPS, no need to be silly)

tbrisker (approved):

PR #1365, opened 2021-05-25
Fetch old /pulp/repos/ path in bats

jlsherrill:
the /pulp/repos path is around to make sure that old clients
still work, this adds a bats test to ensure that.  It also
verifies that /pulp/content/ works on smart proxies


jlsherrill:
draft until i run a test

ehelms (approved):

PR #1367, opened 2021-05-26
katello_devel_github_username is not required

jturel:
Recent change broke the build for our stable devel image build since it doesn't specify the github username.

### DIFF
--- a/roles/katello_devel/meta/main.yml
+++ b/roles/katello_devel/meta/main.yml
@@ -20,6 +20,6 @@ dependencies:
     foreman_installer_options_internal_use_only:
       - "{{ '--katello-devel-scl-ruby=' +  ruby_scl_version if ansible_distribution_major_version == '7' else '' }}"
       - "--katello-devel-admin-password {{ foreman_installer_admin_password }}"
-      - "--katello-devel-github-username '{{ katello_devel_github_username }}'"
+      - "{{ '--katello-devel-github-username=' + katello_devel_github_username if katello_devel_github_username is defined }}"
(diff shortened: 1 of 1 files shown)


roles/katello_devel/meta/main.yml
```diff
@@ -20,6 +20,6 @@ dependencies:
     foreman_installer_options_internal_use_only:
       - "{{ '--katello-devel-scl-ruby=' +  ruby_scl_version if ansible_distribution_major_version == '7' else '' }}"
       - "--katello-devel-admin-password {{ foreman_installer_admin_password }}"
-      - "--katello-devel-github-username '{{ katello_devel_github_username }}'"
+      - "--katello-devel-github-username '{{ katello_devel_github_username | default('') }}'"
```
ekohl, line 23: There is an option to omit things. Perhaps the whole option should not be given if it's empty?

jturel: @ekohl the `default(omit)` option doesn't work here at least in terms of omitting the entire entry from the internal_use_only array. do you have a clean approach to do that? if not, I'm personally OK with passing an empty string - it doesn't cause any problems

ekohl: A few lines above we have a similar pattern for `ruby_scl_version`. Perhaps that's a cleaner approach?

jturel: got it. updated!

jturel:
@ekohl updated to omit the installer flag if the var is undefined

jturel:
@ekohl please merge this if you've no further feedback

ekohl (approved):

PR #1368, opened 2021-05-26
Enable smart proxy dynflow + rex ssh in katello devel

jturel:
This enables the smart proxy pieces to use REX in katello devel with an important byproduct being generated SSH keys that can be placed on client hosts to enable the connection.

Future work might include updating the katello client role to copy the pubkey into authorized_keys automatically


chris1984 (approved):

chris1984:
@ehelms is this one good for a merge?

ehelms (approved):

PR #1372, opened 2021-06-18
Update default Pulpcore version to 3.11

ehelms:


ekohl (approved):
If we wait a bit longer we can go straight to 3.13 ;)

PR #1373, opened 2021-06-21
skip /pulp/content tests on Pulp2 installs

evgeni:
Pulp2 doesn't know about /pulp/content, and thus fails tests on such
installs

### DIFF
--- a/bats/foreman_helper.bash
+++ b/bats/foreman_helper.bash
@@ -45,6 +45,12 @@ tSkipIfNoPulp2() {
   fi
 }
 
+tSkipIfPulp2() {
+  if tIsPulp2; then
+    skip "${1} is not available in scenarios with Pulp 2"
+  fi
+}
+
 tSkipIfHammerBelow018() {
   if tPackageExists tfm-rubygem-hammer_cli; then
     RPM_PACKAGE=tfm-rubygem-hammer_cli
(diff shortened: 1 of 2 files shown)


evgeni:
@jlsherrill mind having a look, this breaks satellite pipelines :(

bats/foreman_helper.bash
```diff
@@ -45,6 +45,12 @@ tSkipIfNoPulp2() {
   fi
 }
 
+tSkipIfPulp2() {
```
ekohl, line 48: Wouldn't it be more consistent to use `tSkipIfNoPulp3`

evgeni: For the name of this function? No, as Pulp2 and 3 *can* be co-hosted.
In theory this should be `tSkipIfAnyContentTypesAreWithPulp2` but I didn't want to go and actually implement that ;)

ekohl: Well yes, really it should be `tSkipUnlessPulp3ServesRPM` but that's tricky to determine.

jlsherrill (approved):

bats/foreman_helper.bash
evgeni: Yeah. I'll merge now to unblock @odilhao, and give it a better name tomorrow

PR #1374, opened 2021-06-22
bat cleanup about skipping tests

evgeni:


bats/fb-katello-content.bats
```diff
@@ -159,7 +159,6 @@ setup() {
 
 # TODO: Move modules-rpms to a more permanent repo https://pulp.plan.io/issues/7333
 @test "create and sync modules-rpms repo" {
-  tSkipIfHammerBelow018
```
ekohl, line 162: This empty line could also be dropped

ekohl (approved):

bats/fb-katello-content.bats
evgeni: I *knew* you would say that :P

ekohl (approved):

PR #1375, opened 2021-06-22
rename tSkipIfPulp2 to tSkipIfNotPulp3Only

evgeni:
this is the result of the discussion in https://github.com/theforeman/forklift/pull/1373#discussion_r655197960

but I am still not *really* happy about the naming.

@ekohl originally suggested `tSkipIfNoPulp3` (and `tSkipUnlessPulp3ServesRPM`).

I think the later is the most correct one, but I don't see a good way to determine that, and if we go by the presence of the `pulp-server` RPM (or the Katello version), we really check "is that a pulp3-only install, or not".

### DIFF
--- a/bats/fb-katello-proxy.bats
+++ b/bats/fb-katello-proxy.bats
@@ -35,7 +35,7 @@ setup() {
 }
 
 @test "content is available from proxy using /pulp/content" {
-  tSkipIfPulp2 "/pulp/content"
+  tSkipIfNotPulp3Only "/pulp/content"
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I kind of agree that `tSkipUnlessPulp3ServesRPM` is hard to determine. I do wonder if we can query some API for it. Perhaps it can be as simple as checking if `pulpcore-content.service` is running? I'm not sure if we run tests where the service is running but isn't serving content.

ehelms (approved):
We are like 1 or 2 releases away from no needing these skips anymore, so the naming works good for me.

PR #1376, opened 2021-06-28
Add upgrade step to work around rubygem-foreman_remote_execution_core…

ehelms:
… breaking nightly

### DIFF
--- a/pipelines/upgrade/07-server_to_final.yaml
+++ b/pipelines/upgrade/07-server_to_final.yaml
@@ -20,6 +20,11 @@
       scenario_os: "{{ pipeline_os }}"
       scenario_version: "{{ forklift_upgrade_version_final }}"
     - foreman_server_repositories
+    # Drop when https://projects.theforeman.org/issues/32897 is resolved
+    - role: workaround_foreman_remote_execution_core
+      when:
+        - forklift_upgrade_version_final == "nightly"
+        - ansible_os_family == "RedHat"
     - foreman_installer
     - role: remove_pulp2
       when:
--- a/roles/workaround_foreman_remote_execution_core/tasks/main.yml
+++ b/roles/workaround_foreman_remote_execution_core/tasks/main.yml
@@ -1,0 +1,20 @@
+---
+- name: Update all packages on the system
+  include_role:
+    name: update_os_packages
+
+- name: Remove rubygem-foreman_remote_execution_core
+  package:
+    state: absent
+    name: rubygem-foreman_remote_execution_core
+  when:
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version == '8'
+
+- name: Remove rubygem-foreman_remote_execution_core
+  package:
+    state: absent
+    name: tfm-rubygem-foreman_remote_execution_core
+  when:
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version == '7'


ehelms:
This fixes the nightly failure we are currently seeing. I believe @adamruzicka is working on fixes in this area to obsolete and ensure all the right packages are removed but this can fix nightly pipelines until the proper fixes are available.

adamruzicka:
Correct, there are last two PRs which need to go in until we'll be able to pull the plug on foreman_remote_execution_core. For the time being, this should be fine if we remember to drop it once rex core goes away

roles/workaround_foreman_remote_execution_core/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Update all packages on the system
+  include_role:
+    name: update_os_packages
+
+- name: Remove rubygem-foreman_remote_execution_core
+  command: yum -y remove rubygem-foreman_remote_execution_core
```
ekohl, line 7: Isn't there a package resource for this?

ekohl (commented):
Can you add some more context in the commit? Either in the commit message (git blame will find it) or as comments?

pipelines/upgrade/07-server_to_final.yaml
```diff
@@ -20,6 +20,9 @@
       scenario_os: "{{ pipeline_os }}"
       scenario_version: "{{ forklift_upgrade_version_final }}"
     - foreman_server_repositories
+    - role: workaround_foreman_remote_execution_core
+      when:
+        - forklift_upgrade_version_final == "nightly"
```
ekohl, line 25: Should this only be if the OS family is Red Hat?
```suggestion
        - forklift_upgrade_version_final == "nightly"
        - ansible_os_family == "RedHat"
```

roles/workaround_foreman_remote_execution_core/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version == '8'
+
+- name: Remove rubygem-foreman_remote_execution_core
+  package:
+    state: absent
+    name: tfm-rubygem-foreman_remote_execution_core
+  when:
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version == '7'
```
ekohl, line 20: Can we capture this in a single statement?
```suggestion
- name: Remove (tfm-)rubygem-foreman_remote_execution_core
  package:
    state: absent
    name:
      - rubygem-foreman_remote_execution_core
      - tfm-rubygem-foreman_remote_execution_core
  when:
    - ansible_os_family == "RedHat"
```

evgeni: This should be possible, yes

ehelms:
Updated with an issue and more data

roles/workaround_foreman_remote_execution_core/tasks/main.yml
ehelms: We could, I'd like to stick with my explicit version as this work around will be dropped in a week or so.

roles/workaround_foreman_remote_execution_core/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+- name: Update all packages on the system
+  include_role:
+    name: update_os_packages
+
+- name: Remove rubygem-foreman_remote_execution_core
+  package:
+    state: absent
+    name: rubygem-foreman_remote_execution_core
+  when:
+    - ansible_os_family == "RedHat"
```
ekohl, line 11: Since the whole task is now guarded by this, I don't think there's a need to repeat it.

ehelms:
The reason this is still needed are the following two PRs:

 * https://github.com/ATIX-AG/smart_proxy_acd/pull/1
 * https://github.com/theforeman/smart_proxy_salt/pull/64

Once those are in and in builds, we can obsolete foreman_remote_execution_core.

evgeni (approved):
I am happy with this as-is given it's a workaround. We should track somewhere that this needs to be removed when we start releasing Foreman 3.0/Katello 4.2, as that's the moment where this issue would hit users and *that* we don't want.

ekohl:
I left a review on smart_proxy_salt since I have permissions on that. Personally I don't mind breaking smart_proxy_acd since we're 2 weeks from stabilization and the PR has been open since May.

ehelms:
@adamruzicka We discussed today, and are OK with putting in the Obsoletes now to avoid this workarond being present and then letting those plugins get fixed after the fact given you've had PRs open for 2 months.

adamruzicka:
ACK, packaging PRs are ready:
- https://github.com/theforeman/foreman-packaging/pull/6855
- https://github.com/theforeman/foreman-packaging/pull/6856

ekohl:
Given packaging is being done (RPM already merged), I'm going to close this workaround PR.

PR #1377, opened 2021-07-01
Drop puppet bats

ezr-ondrej:
The proxy assignment to taxonomies is not puppet related, so I'm moving
it to the Foreman bats.

We are planning to drop puppet from Foreman, so Foreman and Katello
pipelines should not test puppet anymore.

We could use this to add Puppet plugin pipeline or general plugin
pipeline in the future, so I'm leaving the bats itself.


bats/fb-test-foreman.bats
```diff
@@ -16,3 +18,11 @@ set -o pipefail
 @test "check smart proxy is registered" {
   hammer proxy info --name=$(hostname -f)
 }
+
+@test "Assign proxy to default taxonomies" {
```
ekohl, line 22: This looks like it doesn't matter anymore and we should actually drop this workaround on 1.23+

ekohl (changes requested):
This isn't really complete because that's not the nightly pipeline. That playbook is a leftover from old stuff and we should probably delete it.

Here is the Foreman pipeline: https://github.com/theforeman/forklift/blob/daa523e6d5c1a3ad2696fede8fc02b6052c03234/pipelines/vars/forklift_foreman.yml#L14

For our plugin pipeline it's here: https://github.com/theforeman/forklift/blob/daa523e6d5c1a3ad2696fede8fc02b6052c03234/pipelines/vars/forklift_plugins.yml#L32

Katello doesn't contain it: https://github.com/theforeman/forklift/blob/master/pipelines/vars/forklift_katello.yml

That means it uses the default (which you did modify). Removing it there means we also don't test it on stable releases anymore.

Luna is the same as Katello and located here: https://github.com/theforeman/forklift/blob/master/pipelines/vars/forklift_luna.yml

Here is an example of how to make something version dependent:
https://github.com/theforeman/forklift/blob/daa523e6d5c1a3ad2696fede8fc02b6052c03234/pipelines/vars/forklift_katello.yml#L11

Note that Foreman and Plugins use Foreman version numbering while Katello and Luna use Katello version numbering (2.5 vs 4.0).

ezr-ondrej:
Thanks @ekohl for the guidance :) updated the vars and added the bats to plugins pipeline.
Once nightlies are fixed, I'll scrach build and test the pipelines :)

pipelines/vars/forklift_plugins.yml
```diff
@@ -17,6 +17,7 @@ server_box:
         - "--enable-foreman-plugin-bootdisk"
         - "--enable-foreman-plugin-discovery"
         - "--enable-foreman-plugin-hooks"
+        - "--enable-foreman-plugin-puppet"
```
ekohl, line 20: This should also be conditional so we don't break pipes like https://ci.theforeman.org/job/foreman-plugins-2.5-rpm-test-pipeline/ and https://ci.theforeman.org/job/foreman-plugins-2.5-deb-test-pipeline/ (and older).

bats/fb-test-foreman.bats
```diff
@@ -3,6 +3,8 @@
 
 set -o pipefail
 
+load foreman_helper
+
```
ekohl, line 7: Looks like a leftover

ekohl (changes requested):

ekohl (commented):
I still think the katello and luna pipelines need to enable it on run the Puppet tests on Katello <= 4.1

pipelines/vars/forklift_plugins.yml
```diff
@@ -17,6 +17,7 @@ server_box:
         - "--enable-foreman-plugin-bootdisk"
         - "--enable-foreman-plugin-discovery"
         - "--enable-foreman-plugin-hooks"
+        - "{{ '--enable-foreman-plugin-puppet' if pipeline_version == 'nightly' or pipeline_version is version('3.0', '>=') else '' }}"
```
evgeni, line 20:
```suggestion
        - "{{ '--enable-foreman-plugin-puppet' if pipeline_version == 'nightly' or pipeline_version is version('3.1', '>=') else '' }}"
```

Or does it really matter? The plugin is already installed on 3.0, and doesn't need manual installation.

playbooks/bats_pipeline_foreman_nightly.yml
```diff
@@ -11,7 +11,6 @@
     bats_tests:
       - "fb-verify-packages.bats"
       - "fb-test-foreman.bats"
-      - "fb-test-puppet.bats"
```
evgeni, line 14: this file is gone.

evgeni:
> I still think the katello and luna pipelines need to enable it on run the Puppet tests on Katello <= 4.1

this can be done here:
https://github.com/theforeman/forklift/blob/9563ec314184e7cb8f392aa1c6c94e90a96d3664/pipelines/vars/forklift_katello.yml#L16-L18
and here:
https://github.com/theforeman/forklift/blob/9563ec314184e7cb8f392aa1c6c94e90a96d3664/pipelines/vars/forklift_luna.yml#L46-L50

yes, making the array entry an empty string works these days (since https://github.com/theforeman/forklift/pull/1415)

pipelines/vars/forklift_plugins.yml
ezr-ondrej: It was for not installing it prior 3.0 release... now it does not matter :)

pipelines/vars/forklift_foreman.yml
```diff
@@ -11,7 +11,6 @@ server_box:
       bats_tests:
         - "fb-verify-packages.bats"
         - "fb-test-foreman.bats"
-        - "fb-test-puppet.bats"
```
ezr-ondrej, line 14: I guess we can just ignore that this is kinda relevant to current releases?

pipelines/vars/forklift_luna.yml
```diff
@@ -30,6 +30,9 @@ server_box:
         - "--enable-foreman-plugin-openscap"
         - "--enable-foreman-proxy-plugin-openscap"
 
+        - "--enable-foreman-cli-puppet"
+        - "--enable-foreman-plugin-puppet"
```
ezr-ondrej, line 34: Do I need to care that these are not available on 2.5 release? Are we running this pipeline there?

pipelines/vars/forklift_plugins.yml
ezr-ondrej: or maybe not, lets be safe :)

pipelines/vars/forklift_luna.yml
ezr-ondrej: I did to be on the safe side

evgeni: we only *run* luna for nightly, but there is nothing that forbids a user to run this with `pipeline_version` parameter, so thanks!

pipelines/vars/forklift_foreman.yml
evgeni: No?

Just do:
```yaml
- "{{ 'fb-test-puppet.bats' if pipeline_version != 'nightly' and pipeline_version is version('3.1', '<') else '' }}"

File: bats/fb-test-puppet.bats [hunk 8]
```diff
@@ -70,14 +70,6 @@ fi
   hammer puppet-environment update --name=production --locations "Default Location" --organizations "Default Organization"
 }
 
-@test "Assign proxy to default taxonomies" {
```
ekohl, line 73: Let's separate this out to its own PR. I think that can already be merged and reduces the scope of this PR.

ekohl (commented):
For completeness: at least now the installer enables the plugin by default. While this work is needed for https://github.com/theforeman/foreman-installer/pull/717, we're some way away from completing that.

ekohl:
> This will make Katello not run Puppet tests either, which I think is wrong for pre 4.2?

Today it must continue testing it in all branches, including nightly. It's still enabled by default.

bats/fb-test-puppet.bats
ezr-ondrej: extracted to #1419

PR #1378, opened 2021-07-01
Update default Pulp repositories and nightly to 3.14

ehelms:
The current state of nightly expects this due to changes in puppet-pulpcore. We can hold this for 3.14 or merge now and then update again. I was hitting this testing nightly and so might others.

### DIFF
--- a/roles/pulpcore_repositories/defaults/main.yml
+++ b/roles/pulpcore_repositories/defaults/main.yml
@@ -1,2 +1,2 @@
 ---
-pulpcore_repositories_version: '3.11'
+pulpcore_repositories_version: '3.14'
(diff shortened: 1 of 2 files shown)


evgeni:
Won't the pipelines rather use https://github.com/theforeman/forklift/blob/daa523e6d5c1a3ad2696fede8fc02b6052c03234/vagrant/config/versions.yaml#L86?

Plus I guess we should make katello-repos match, and also ensure the katello client bindings match?

ehelms:
I'll just stage up the 3.14 changes needed

ekohl (approved):

evgeni (approved):

PR #1381, opened 2021-07-09
module hotfixes fixes

evgeni:


evgeni:
this makes the katello install pipeline on el8 pass now

ekohl (commented):
If it didn't work for EL8 (and it really is only for EL8, right?) then why is it really needed for Foreman's repos?

evgeni:
to match what we deploy in `foreman-release` and just to be on the safe side if we ever decide to override a modular gem in the future

ekohl (approved):

PR #1382, opened 2021-07-09
drop pulp2 from Katello 4.1 and nightly

evgeni:


roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -41,6 +43,8 @@
     gpgcheck: no
     enabled: yes
   when:
+    - katello_repositories_version != "nightly"
+    - katello_repositories_version is version('4.0', '<=')
     - katello_repositories_pulp_version == "nightly"
```
ekohl, line 48: I don't think we use pulp nightly anywhere now. Should we drop it?

ekohl (approved):

roles/katello_repositories/tasks/staging_repos.yml
evgeni: yeah, good idea

evgeni: actually, pipelines/vars/katello_devel.yml still refers to it. I'll follow up with katello about this.

PR #1385, opened 2021-07-14
Cleanup import organziation in bats

ehelms:
@jturel @jeremylenz This makes it so you can more easily re-run the bats tests locally when testing or investigating issues


jeremylenz:
LGTM, I don't think it would hurt anything..

evgeni (approved):

jturel (approved):

PR #1390, opened 2021-07-19
run hammer ping on Foreman too

evgeni:
this exists since 1.24

### DIFF
--- a/bats/fb-test-foreman.bats
+++ b/bats/fb-test-foreman.bats
@@ -26,7 +26,7 @@ load foreman_helper
 
   local next_wait_time=0
   until [ "${status:-1}" -eq 0 -o $next_wait_time -eq 12 ]; do
-    run hammer ping
+    run hammer --reload-cache ping
     [[ $status -eq 0 ]] || sleep $(( next_wait_time++ ))
   done
 
--- a/bats/fb-test-foreman.bats
+++ b/bats/fb-test-foreman.bats
@@ -3,6 +3,9 @@
 
 set -o pipefail
 
+load os_helper
+load foreman_helper
+
 @test "assert correct Foreman version is installed" {
   [ -n "$FOREMAN_EXPECTED_VERSION" ] || skip "FOREMAN_EXPECTED_VERSION is not set, not asserting"
   # prints and fails if any VERSION files don't match expected version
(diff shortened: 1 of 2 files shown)


bats/fb-test-foreman.bats
```diff
@@ -13,6 +16,29 @@ set -o pipefail
   [ `curl -sk "https://localhost$URL_PREFIX/status" | ruby -e "require 'json'; puts JSON.load(ARGF.read)['status']"` == 'ok' ]
 }
 
+@test "check hammer ping" {
+  FOREMAN_VERSION=$(tForemanVersion)
+  if [ $FOREMAN_VERSION == '1.24' ]; then
+    # https://github.com/theforeman/hammer-cli-foreman/pull/394
+    # https://github.com/Katello/hammer-cli-katello/pull/596
+    skip "Functionality is broken in 1.24"
```
evgeni, line 24: I kept this skip, as Satellite 6.7 is still supported, and is based on 1.24.

bats/fb-test-foreman.bats
```diff
@@ -13,6 +16,29 @@ set -o pipefail
...
+  fi
+
+  local next_wait_time=0
+  until [ "${status:-1}" -eq 0 -o $next_wait_time -eq 12 ]; do
+    run hammer --reload-cache ping
```
evgeni, line 29: and I've added `--reload-cache` here, so that the apidoc is always fresh

bats/fb-test-katello.bats
```diff
@@ -6,29 +6,6 @@ set -o pipefail
...
-  echo "${output}"
-
-  [ $status -eq 0 ]
-
-  # Hammer exits with 0 on failures
-  # https://projects.theforeman.org/issues/30496
-  [[ $output != *"FAIL"* ]]
-}
-
 @test "check service status" {
```
ekohl, line 9: Should this be `fb-test-foreman-maintain.bats` now?

ekohl (approved):
I thought I submitted a PR about this but clearly not. However, I had the exact same thought in the past.

bats/fb-test-katello.bats
evgeni: In theory, yes. Or maybe even foreman.bats with a `if package installed` guard. But I'd like to leave that as an exercise to the reader.

PR #1391, opened 2021-07-21
Add global registration bats tests for Katello clients

ehelms:


bats/fb-katello-client-global-registration.bats
```diff
@@ -0,0 +1,74 @@
...
+@test "delete host if present" {
+  hammer host delete --name=$HOSTNAME || echo "Could not delete host"
+}
+
+@test "register with global registration with activation key" {
+  run yum erase -y 'katello-ca-consumer-*'
+  echo "rc=${status}"
+  echo "${output}"
+
+  registration_command=$(curl https://admin:changeme@$HOSTNAME/api/registration_commands -X POST -H 'Content-Type: application/json' -d '"activation_key":"${ACTIVATION_KEY}","organization":"${ORGANIZATION_LABEL}"' | ruby -e "require 'json'; puts JSON.load(ARGF.read)['registration_command']")
```
ehelms, line 33: This appears to not work as it generates the registration command with `Default_Organization` instead of `Test_Organization` not respecting passing the organization label? Is this a bug or am I setting this wrong? @ares @stejskalleos 

Additionally, I noticed that if I pass a parameter that registeration does not support there is not an error, the parameter is just ignored. That led to some debugging confusion

ekohl: For what it's worth, you shouldn't have to post using JSON. You can also shorten the Ruby line a bit.
```suggestion
  registration_command=$(curl https://admin:changeme@$HOSTNAME/api/registration_commands -X POST -d "activation_key=${ACTIVATION_KEY}" -d "organization=${ORGANIZATION_LABEL}" | ruby -rjson -e "puts JSON.load(ARGF.read)['registration_command']")
```

evgeni: I think you need to pass IDs, not names/labels for org?

https://github.com/theforeman/foreman/blob/5a66daa44921e63522bb90d01f8924b6cd4abcd5/app/controllers/concerns/foreman/controller/registration.rb#L17-L20

evgeni: > Additionally, I noticed that if I pass a parameter that registeration does not support there is not an error, the parameter is just ignored. That led to some debugging confusion

What if I told you, ALL of our API behaves like this, and I might have gray hair because of that?

ehelms: Yea... @ares was saddening me with this. Just makes debugging that much harder in this case to know what my client is doing.



> I think you need to pass IDs, not names/labels for org?

Yea... for this to work right now I will have to do that. I left it like to to first help the conversation about adding name/label support for organizations.

roles/bats/defaults/main.yaml
```diff
@@ -17,6 +17,7 @@ bats_tests:
   - "fb-test-katello.bats"
   - "fb-katello-content.bats"
   - "fb-katello-client.bats"
+  - "fb-katello-client-global-registration.bats"
```
ehelms, line 20: I probably need to not include this for some versions of Foreman and Katello we currently support. What those would be, I am not presently sure of.

bats/fb-katello-client-global-registration.bats
ares: I wonder if the user's default org works here. But I guess you want to make sure, user can use a specific org anyway... I can imagine adding another top-level parameter to specify org using the label, but that's not related to the registration.

ehelms: > I wonder if the user's default org works here. But I guess you want to make sure, user can use a specific org anyway... I can imagine adding another top-level parameter to specify org using the label, but that's not related to the registration.

What is the course of action here? I should file a bug?

ares: I would call it an RFE :-) I guess we shouldn't block the test on it. Note that if you insisht on org label and not the name, it must be added from Katello, since label is a katello thing. Perhaps it's time to move this extension to Foreman core where it belongs to?

ares (commented):
Great to see we'll have test for this, do we only keep happy paths here or can we add more variants? It would be awesome to have a test that also creates PAT (personal access token) for a user (e.g. admin) and then we get the registration command using that instead of `changeme` password. OTOH if we already have test for PAT elsewhere, ignore me :-)

bats/fb-katello-client-global-registration.bats
ehelms: I can file an RFE for the label for Katello. Should *name* work though?

ehelms: > I can file an RFE for the label for Katello. Should _name_ work though?

Oh I guess not from what @evgeni linked. I'll file an issue for that too.

ehelms: > I would call it an RFE :-) I guess we shouldn't block the test on it. Note that if you insisht on org label and not the name, it must be added from Katello, since label is a katello thing. Perhaps it's time to move this extension to Foreman core where it belongs to?

By extension, you mean the concept of a label on organization to Foreman core?

ehelms: I filed the following for consideration:

 * https://projects.theforeman.org/issues/33227
 * https://projects.theforeman.org/issues/33228

ehelms: > For what it's worth, you shouldn't have to post using JSON. You can also shorten the Ruby line a bit.

The API only accepts JSON when using POST.

ehelms:
This has been updated to use `organization_id` for now while we work out what to do about name and label. This only adds tests for Katello at the moment and aims to mimic the basics of the existing client bats tests. I opened an issue to track adding additional global registration tests: https://github.com/theforeman/forklift/issues/1396

My goal is to get basic coverage for the Katello scenario to help ensure the new workflow works as we aim to replace the old one and then in subsequent PRs layer in new test suites.

ehelms:
@jturel @stejskalleos Could you both take a look to ensure this looks OK and that it captures a workflow you'd expect?

ares:
Makes sense to me, IMHO good to go in (didn't perform the syntax check, just the semantics) :+1:

roles/bats/defaults/main.yaml
jturel: 4.0+ I believe

jturel:
Looks good to me :)

ehelms:
@evgeni Mind giving this a look?

bats/fb-katello-client-global-registration.bats
```diff
@@ -0,0 +1,78 @@
...
+load foreman_helper
+load fixtures/content
+
+@test "remove subscription manager" {
+  if [ -e "/etc/rhsm/ca/candlepin-local.pem" ]; then
+    rpm -e `rpm -qf /etc/rhsm/ca/candlepin-local.pem`
+  fi
+
+  cleanSubscriptionManager
+  tPackageRemove subscription-manager
```
evgeni, line 16: Is that so that the GR command can install it again?

Doesn't this break pulp3 on EL8 where this might kill python3-rhsm (or whatever the right pkg name is)?

bats/fb-katello-client-global-registration.bats
```diff
@@ -0,0 +1,78 @@
...
+@test "enable content view repo" {
+  subscription-manager repos --enable="${ORGANIZATION_LABEL}_${PRODUCT_LABEL}_${YUM_REPOSITORY_LABEL}" | grep -q "is enabled for this system"
+}
+
+@test "install package locally" {
+  run yum -y remove walrus
+  tPackageInstall walrus-0.71 && tPackageExists walrus-0.71
+}
+
+@test "check available errata" {
```
evgeni, line 58: Feels like this duplicates the original content tests -- might want to move it to a helper later?

evgeni (approved):
Two comments, but willing to give it a try as-is and cleanup the broken pieces later.

(Assuming you have run this against nightly and it passes 😅)

bats/fb-katello-client-global-registration.bats
ehelms: I hope not. I did not see any breaks with my testing.

bats/fb-katello-client-global-registration.bats
ehelms: Yea, can do, as I have another test suite in mind that would do the same thing.

bats/fb-katello-client-global-registration.bats
evgeni: If you ran it on EL8 and it did not make pulp-certguard go belly up, we're good.

ehelms: Looks all good:

```
==========================================================================================================================================================================================================================================================================================
 Package                                                                              Architecture                                                Version                                                              Repository                                                    Size
==========================================================================================================================================================================================================================================================================================
Removing:
 subscription-manager                                                                 x86_64                                                      1.28.13-3.el8_4                                                      @baseos                                                      4.3 M
Removing unused dependencies:
 dnf-plugin-subscription-manager                                                      x86_64                                                      1.28.13-3.el8_4                                                      @baseos                                                       87 k
 python3-dmidecode                                                                    x86_64                                                      3.12.2-15.el8                                                        @baseos                                                      308 k
 python3-ethtool                                                                      x86_64                                                      0.14-3.el8                                                           @baseos                                                       92 k
 python3-inotify                                                                      noarch                                                      0.9.6-13.el8                                                         @baseos                                                      243 k
 python3-librepo                                                                      x86_64                                                      1.12.0-3.el8                                                         @baseos                                                      172 k
 usermode                                                                             x86_64                                                      1.113-1.el8                                                          @baseos                                                      837 k

Transaction Summary
==========================================================================================================================================================================================================================================================================================
Remove  7 Packages
```

PR #1392, opened 2021-07-21
Disable Puppet by default in the Katello devel scenario

ekohl:
After Puppet was extracted from Foreman itself, it either needs the plugin installed or it should be disabled. Disabling saves resources and allows testing of the scenario where it is off.

Untested, but it came up in https://github.com/theforeman/puppet-katello_devel/pull/264#issuecomment-884414228.

### DIFF
--- a/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
+++ b/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
@@ -9,6 +9,7 @@ katello_devel:
 foreman_proxy_content:
   enable_katello_agent: true
   proxy_pulp_yum_to_pulpcore: true
+  puppet: false
 foreman_proxy:
   http: true
   ssl_port: '9090'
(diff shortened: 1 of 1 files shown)


ekohl:
Note we don't clean up existing Puppet installations. If you want to do so:
```
yum -y remove puppetserver
systemctl disable --now puppet
```

roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
```diff
@@ -18,16 +19,9 @@ foreman_proxy:
...
-  server_environments_owner: apache
-  server_foreman_ssl_ca: /etc/pki/katello/puppet/puppet_client_ca.crt
-  server_foreman_ssl_cert: /etc/pki/katello/puppet/puppet_client.crt
-  server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
-  # Tune the server for a minimal deployment
-  puppet_server_jvm_max_heap_size: 1G
-  puppet_server_jvm_min_heap_size: 1G
-  puppet_server_max_active_instances: 1
+  puppet: false
+  puppetca: false
```
ehelms, line 23: I think this needs `manage_puppet_group: false` here.

```
    2021-07-21 19:29:15 [ERROR ] [configure] Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: File[/etc/foreman-proxy/ssl_key.pem] is already declared at (file: /usr/share/foreman-installer/modules/foreman_proxy/manifests/config.pp, line: 135); cannot redeclare (file: /usr/share/foreman-installer/modules/certs/manifests/keypair.pp, line: 36) (file: /usr/share/foreman-installer/modules/certs/manifests/keypair.pp, line: 36, column: 5) (file: /usr/share/foreman-installer/modules/certs/manifests/foreman_proxy.pp, line: 95) on node centos7-devel.wareagle.example.com
```

The turning off of puppet features triggers this block https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/config.pp#L118

Perhaps that code could use an update to be smarter about determining management of the puppet group.

ekohl: Ah yes, we should. Then we can also get rid of https://github.com/theforeman/foreman-installer/blob/bc575c2d662dad4779623b89d24ca7750dcec3bd/config/katello-answers.yaml#L73 and https://github.com/theforeman/foreman-installer/blob/bc575c2d662dad4779623b89d24ca7750dcec3bd/config/foreman-proxy-content-answers.yaml#L21. For now I'll add it here too.

ehelms (approved):

roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
ehelms: Is this something you could capture in an installer issue? I think you'll word it better than I will.

PR #1393, opened 2021-07-28
Add 'Foreman Operations Collection' to Requirements

stejskalleos:


ekohl (changes requested):
I think you need to update. https://github.com/theforeman/forklift#quickstart does have a `ansible-galaxy collection install -r requirements.yml` line in the quickstart but it looks like you have an old checkout.

stejskalleos:
Oh, okay, I'm closing the issue then

PR #1394, opened 2021-08-02
Foreman 3.0 forklift verisions

pcreech:


ekohl (commented):
There's a typo in `versions` in the commit message. I'm also guessing the test failure is related. We should also not merge this until we actually have branched because it'll break nightly until there is an rc1.

pcreech:
Test failure is on a task that says "  TASK [Ensure 3.18 was properly skipped] "

ehelms:
Update the when condition here https://github.com/theforeman/forklift/blob/master/roles/forklift_versions/molecule/default/verify.yml#L100 to 4.1/4.2/nightly and I think that should solve this error and preserve the test

pcreech:
@ekohl @ehelms That worked.  We're green

ekohl (approved):
We'll break nightly after this, but RC1 PRs are open. To minimize the windows that they're broken we should merge this after those 2 are merged (https://github.com/theforeman/foreman-packaging/pull/6996 / https://github.com/theforeman/foreman-packaging/pull/6997).

PR #1402, opened 2021-08-19
Make hostname a deployment setting

fao89:


ekohl (changes requested):
I really dislike this. Our settings are global and we are by definition a multi box setup. :-1: unless you actually make a good case why this is a good idea.

fao89:
I don't have a good case, I'm closing it

chris1984:
@fao89 Ruby is better than Python

PR #1406, opened 2021-08-30
Allow machines to mount to its own folders simultaneously

ofedoren:
Currently if one runs `vagrant up box1` with sshfs enabled then it will mount to e.g. `/box1_dir`, but if one runs `vagrant up box2` with `/box2_dir` mount dir after, then the `box1` will use the `/box2_dir` in the next `vagrant up box1`.

Fixed only for sshfs, since that's what I'm using all the time.


ekohl (approved):
Looks correct to me. I think NFS needs a similar treatment, but I don't have a setup to test that so I'm merging this now. Thanks!

PR #1407, opened 2021-08-30
Remove goferd after katello-client tests

jturel:
This fixes an issue that has come up in the nightly pipeline where goferd is installed and running, but katello-agent has been removed - presumably after some initial run of the katello-client tests. when the katello-client tests run again katello-agent is installed but gofer doesn't auto-detect the agent plugin which causes the subsequent tests to fail. this fixes that by removing gofer after the client tests run

bonus: only run the installer to enable katello-agent if qpid is not installed already

### DIFF
--- a/bats/fb-katello-client.bats
+++ b/bats/fb-katello-client.bats
@@ -118,6 +118,7 @@ load fixtures/content
   timeout 300 hammer host package remove --host $HOSTNAME --packages walrus
 }
 
-@test "cleanup subscription-manager after content tests" {
+@test "clean up subscription-manager and gofer after content tests" {
   cleanSubscriptionManager
+  tPackageRemove gofer
 }


bats/fb-katello-client.bats
```diff
@@ -118,6 +123,7 @@ load fixtures/content
   timeout 300 hammer host package remove --host $HOSTNAME --packages walrus
 }
 
-@test "cleanup subscription-manager after content tests" {
+@test "cleanup subscription-manager and gofer after content tests" {
```
jeremylenz, line 126:
```suggestion
@test "clean up subscription-manager and gofer after content tests" {
```

jeremylenz (approved):
LGTM :+1:

jturel:
@ehelms what do ya think?

bats/fb-katello-client.bats
```diff
@@ -12,6 +12,11 @@ load fixtures/content
   if [[ $KATELLO_VERSION != 4.[1-9]* ]]; then
     skip "Enabling katello-agent explicitly is only available with Katello 4.1+"
   fi
+
+  if tPackageExists qpid-cpp-server ; then
```
ehelms, line 16: I am trying to think if this could cause problems as this alone is not a true indicator if katello-agent is enabled.

ehelms:
Should https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/client/katello-host-tools/katello-host-tools.spec#L271 be triggering a restart?

bats/fb-katello-client.bats
jturel: Yeah, it's quick and dirty. I could check the output of the ping api or something like that too, but I'd rather do so only if it's worth it

jturel:
> Should https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/client/katello-host-tools/katello-host-tools.spec#L271 be triggering a restart?

No, that comes into play when katello-agent is already installed and it gets updated, triggering itself to be reloaded. This PR addresses when katello-agent is installed when goferd is running already without the agent plugin

bats/fb-katello-client.bats
ehelms: Yea, I think we should drop this for now or derive it from `katello.yaml` or the answers file.

bats/fb-katello-client.bats
```diff
@@ -118,6 +123,7 @@ load fixtures/content
   timeout 300 hammer host package remove --host $HOSTNAME --packages walrus
 }
 
-@test "cleanup subscription-manager after content tests" {
+@test "clean up subscription-manager and gofer after content tests" {
   cleanSubscriptionManager
+  tPackageRemove gofer
```
ehelms, line 128: This looks good

bats/fb-katello-client.bats
jturel: dropping for now!

jturel:
@ehelms updated!

ehelms (approved):

PR #1408, opened 2021-09-03
Install consumer RPM with yum and not RPM

ehelms:
By using rpm to install the consumer RPM dependencies are not resolved
and the tests can end up in a bad state of available dependencies such
as subscription-manager which is removed when older RPMs are cleaned up.

### DIFF
--- a/bats/fb-katello-client.bats
+++ b/bats/fb-katello-client.bats
@@ -16,10 +16,6 @@ load fixtures/content
   foreman-maintain packages unlock -y
 }
 
-@test "install subscription manager" {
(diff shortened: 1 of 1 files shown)


ehelms:
This is aimed at fixing nightly failure -- https://community.theforeman.org/t/katello-nightly-rpm-pipeline-1077-failed/25124

I also do not fully see why this is not a problem *every* time on every operating system, something must be on the systems that prevents subscription-manager from being removed.

jlsherrill (approved):

PR #1411, opened 2021-09-13
set pulp_pkg_name_prefix to the correct Python 3.8 prefix for 3.15+

evgeni:
Requires https://github.com/pulp/pulp_installer/pull/758


ekohl (commented):
I may be hijacking this a bit, but have you considered implementing `Provides: pulpcore-plugin-$plugin` on all packages?

evgeni:
Right now, the installer also needs to install packages that are not Pulpcore plugins, so while yes, it would make things easier, it wouldn't fix them :(

ekohl:
To be fair, that's only the Pulp installer. On the Foreman side I think we don't do that.

evgeni:
True, but that's what forklift uses for the pulpcore pipelines  ;)

ekohl (approved):
That's why I said hijacking ;) I'm not going to touch pulp_installer, I just care about our side of it.

PR #1415, opened 2021-09-27
Omit bats test for registration instead of using empty string

jlsherrill:


evgeni (commented):
Omit only works for module parameters, not array entries

evgeni:
You'll need something like https://github.com/theforeman/forklift/blob/279f047780a2d9d5936f9a45234e19606e37e2b0/pipelines/vars/forklift_pulpcore.yml#L23

jlsherrill:
oh ignore me, i see your change is more involved, let me do that instead!

evgeni (approved):

jlsherrill:
my 4.0 test run passed (well smoker failed, but unrelated)

PR #1421, opened 2021-10-08
[WIP] Install epel repository on CentOS 8 (and Stream)

stejskalleos:
Without epel repository foreman-installer is failing
due to missing 'qpid-proton-c-devel' package


stejskalleos:
Should be fixed by https://github.com/theforeman/puppet-katello_devel/pull/269

PR #1440, opened 2021-11-09
Add Vagrant Config for Foreman 3.1

Odilhao:


ekohl (approved):
Merging this will break nightly upgrade pipelines until 3.1 / 4.3 RC1 is available. Not merging this until we're at the point where we actually want to run 3.1 pipelines.

Odilhao:
Merging this for RC1 pipelines

PR #1443, opened 2021-11-17
Update Pulpcore to 3.16 in nightly

ianballou:
This should go in after the Katello and packaging PRs:

- https://github.com/theforeman/foreman-packaging/pull/7221
- https://github.com/Katello/katello/pull/9758


evgeni:
I think this should also bump the default version in https://github.com/theforeman/forklift/blob/2d2749e73d1f3ab8ec2ba18ed7582524e796c7d3/roles/pulpcore_repositories/defaults/main.yml#L2

jlsherrill (approved):

PR #1444, opened 2021-11-18
create and fetch a backup as part of the tests

evgeni:


bats/fb-test-backup.bats
```diff
@@ -0,0 +1,14 @@
...
+# vim: ft=sh:sw=2:et
+
+set -o pipefail
+
+load os_helper
+load foreman_helper
+
+@test "perform backup" {
+  mkdir -p /var/tmp/foreman-backup
+  # this is a workaround for a check that is only relevant for DB dumps
```
ekohl, line 11: I recall there was a BZ for this in downstream. Do we have an upstream issue link we can track and put in a comment here?

ekohl (commented):
Foreman Maintain is not packaged on Debian and not installed by default on vanilla Foreman. Won't that create problems here?

I wonder if we can do a minimal verification of the created backup. Something like `tar taf /path/to/backup | grep whatever_foreman_db_filename`.

bats/fb-test-backup.bats
evgeni: I can do the :woman_dancing:

evgeni:
> Foreman Maintain is not packaged on Debian and not installed by default on vanilla Foreman. Won't that create problems here?

No, because the *Foreman* pipelines override the list of bats test anyways, the default set is really about Katello ;)

> I wonder if we can do a minimal verification of the created backup.

If you have suggestions, I'll gladly take them.

My original idea was to create another job, that will pull down the last successfull backup from Jenkins, and try to restore it ;)
And for the Katello case, we even have a few "this needs to work after restore" tests:
https://github.com/theforeman/forklift/blob/431f9e10fbfa829e82b84b4b23e1eb8497240030/bats/fb-katello-content-restore.bats#L10-L30

ekohl:
> If you have suggestions, I'll gladly take them.

I was thinking about listing the contents of the tar and checking for some known filenames. Perhaps the metadata filename at least?

> My original idea was to create another job, that will pull down the last successfull backup from Jenkins, and try to restore it ;)

I like the idea of a pipeline that starts from scratch and restores a backup to see if it still performs its function.

evgeni:
Well, metadata is outside the tarball, so I now just check if a certain set of files is present, but not look *inside* them.

bats/fb-test-backup.bats
evgeni: done

bats/fb-test-backup.bats
```diff
@@ -0,0 +1,30 @@
...
+
+load os_helper
+load foreman_helper
+
+BACKUP_DIR=/var/tmp/foreman-backup
+
+@test "perform backup" {
+  mkdir -p /var/tmp/foreman-backup
+  # this is a workaround for a https://projects.theforeman.org/issues/33947
+  chmod 777 /var/tmp/foreman-backup
```
ekohl, line 14:
```suggestion
  mkdir -p $BACKUP_DIR
  # this is a workaround for a https://projects.theforeman.org/issues/33947
  chmod 777 $BACKUP_DIR
```

bats/fb-test-backup.bats
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bats
+# vim: ft=sh:sw=2:et
+
+set -o pipefail
+
+load os_helper
+load foreman_helper
+
+BACKUP_DIR=/var/tmp/foreman-backup
```
ekohl, line 9: Would it make sense to create a temporary directory?
```suggestion
BACKUP_DIR=$(mktemp -d --tmpdir=/var/tmp/foreman-backup)
```

evgeni: What would be the benefit? The test can be run multiple times?

I've added the `--preserve-directory` so that FM doesn't create it's timestamped dir so the tests don't need to *find* the result. That problem wouldn't exist with `mktemp`, but I am not convinced we gain anything (but less readable paths)

ekohl: Running it multiple times indeed. We generally try to keep our test suite in a way that you can rerun it. Another would be to remove the directory if already present.

evgeni: I'd go with the removal before run then.

ekohl (approved):
I'd consider this to make it Foreman on EL compatible:
```bash
tPackageExists foreman-maintain || tPackageInstall foreman-maintain
```

Ideally this (and the `mkdir -p`) would be in `setup_file` but AFAIK the bats version in EPEL7 is too old to support that.

I believe this should work since there's a `Provides: foreman-maintain = %{version}` line in the spec.

ekohl (approved):
Thanks for shaving all my yaks. Shame they'll be cold this winter though ;)

evgeni:
`tPackageExists` doesn't support provides. And yes, no `setup_file` in EPEL.

So I've implemented the setup as a test.

ekohl:
> `tPackageExists` doesn't support provides

Surprising. I thought `rpm -q` did do so. I think it can be modified by using `--whatprovides` but in this case I think we're good since there's no SCL involved and also no Debian packaging convention.

PR #1447, opened 2021-12-01
don't try to install rpmdevtools

evgeni:
we dropped the usage of it in e16a19d7821a4c44e467352e7ad69d1fa59ffa1a,
but didn't drop installing it

Fixes: e16a19d7821a4c44e467352e7ad69d1fa59ffa1a


evgeni:
unblocking @pcreech

PR #1448, opened 2021-12-02
Debian 11 support

evgeni:


roles/puppet_repositories_bullseye/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: 'Install Puppet repository'
```
ehelms, line 2:
```suggestion
- name: 'Install Buster Puppet repository on Bullseye'
```

And include a comment with the ticket from puppetlabs?

ehelms (approved):

PR #1449, opened 2021-12-06
Add a CentOS 9 Stream box

ekohl:
Currently centos/stream9 doesn't exist on vagrantup.com but there is a direct link. This allows early work to start while CentOS works on automatically updating vagrantup.com.

### DIFF
--- a/vagrant/boxes.d/00-centos.yaml
+++ b/vagrant/boxes.d/00-centos.yaml
@@ -36,3 +36,8 @@ boxes:
     scenarios:
       - foreman
       - katello
+
+  centos9-stream:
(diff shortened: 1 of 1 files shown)


ekohl:
This came up in https://lists.centos.org/pipermail/centos-devel/2021-December/098745.html. It also brought up https://bugs.centos.org/view.php?id=18028 and tells us the stream8 box is also outdated.

evgeni:
"fun"

And I was complaining about the c8s box on #centos-stream the other day and noone pointed me at the other souro :(

evgeni (approved):

PR #1451, opened 2021-12-16
bats cleanup

evgeni:


bats/fb-test-foreman.bats
```diff
@@ -17,13 +17,6 @@ load foreman_helper
 }
 
 @test "check hammer ping" {
-  FOREMAN_VERSION=$(tForemanVersion)
-  if [ $FOREMAN_VERSION == '1.24' ]; then
```
evgeni, line 21: need to check if we don't need this anywhere, hold tight!

ekohl (approved):

bats/fb-test-foreman.bats
evgeni: I quote "dead dead EOL" ;)

PR #1452, opened 2022-01-03
generate correct pkg prefix for pulpcore packages

evgeni:


ekohl (approved):
:+1: for more future compatible. This would also work on EL9.

PR #1453, opened 2022-01-11
force hammer cache reload when enabling puppet

evgeni:
no idea why/how this worked before, but you need to drop the hammer apidoc cache if you change the apidoc


ekohl (commented):
I'm ok with a workaround, but could we open an issue in the installer/hammer to document this workaround?

I can imagine that in some conditions hammer could check if its cache is still fresh, for example if a command is not found.

evgeni:
In theory, Apipie "knows" the cache expired and loads a new apidoc, but it might be not enough to signal Hammer "hey, rebuild your commands".

https://github.com/Apipie/apipie-bindings/blob/f7c10b0ea1c794ca6d9754885218d2f2bc2f15c4/lib/apipie_bindings/api.rb#L260-L266

ekohl:
Sounds like we should have something in Hammer's bug tracker, unless @ofedoren can enlighten us.

ekohl:
For context, this comes from https://community.theforeman.org/t/katello-nightly-rpm-pipeline-1196-failed/26792.

ofedoren:
For a workaround that should do this. When I was spinning up a new instance and then manually installed Puppet, I've noticed that the API docs are updated, but hammer need to download a new version explicitly with -r option :/

Nitpick: you don't have to have additional command such as ping to reload the cache, simple --reload-cache will do and you'll save some time by not pinging all the services unless you need to.

evgeni:
Added a reference to https://projects.theforeman.org/issues/28283

And I like the `ping`, it's not that expensive, and ensures the app is actually up and running after the installer run.

ekohl (approved):

PR #1457, opened 2022-01-21
drop bullseye puppet workaround

evgeni:


ekohl (approved):
Yay, fewer workarounds.

PR #1458, opened 2022-01-24
filter out ostree param for old installer

evgeni:


roles/foreman_installer/tasks/install.yml
```diff
@@ -8,7 +8,7 @@
 
 - name: 'Join options'
   set_fact:
-    foreman_installer_options_joined: "{{ foreman_installer_options | join(' ') }} {{ foreman_installer_options_internal_use_only | join(' ') }}"
+    foreman_installer_options_joined: "{{ foreman_installer_options | difference([] if foreman_installer_version is version('3.1', '>=') else ['--foreman-proxy-content-enable-ostree true']) | join(' ') }} {{ foreman_installer_options_internal_use_only | join(' ') }}"
```
ekohl, line 11: Is this also something that should be applied to `upgrade.yml`?

roles/foreman_installer/tasks/installer_version.yml
```diff
@@ -3,6 +3,10 @@
     src: /usr/share/foreman-installer/VERSION
   register: foreman_installer_version_file
 
+- name: 'Decode installer version'
+  set_fact:
+    foreman_installer_version: "{{ foreman_installer_version_file['content'] | b64decode }}"
```
ekohl, line 8: Can you also modify `upgrade.yml` to use this new variable rather than `b64decode` it manually?

roles/foreman_installer/tasks/install.yml
evgeni: no, because that is calling `install.yml` again, not joining the options itself.

roles/foreman_installer/tasks/installer_version.yml
evgeni: sure

ekohl (approved):
I'm not that happy with this workaround in the general installer, but I guess we'll have to accept it for now.

PR #1461, opened 2022-01-26
support deploying with custom certs

evgeni:


ekohl (commented):
Btw, ownca here should not be confused with Ansible's ownca in https://docs.ansible.com/ansible/latest/collections/community/crypto/x509_certificate_module.html

PR #1462, opened 2022-01-26
make fips role work with el7 without separate /boot

evgeni:
also, don't use findmnt, ansible has all the data already in facts

### DIFF
--- a/roles/fips/tasks/main.yml
+++ b/roles/fips/tasks/main.yml
@@ -2,16 +2,7 @@
 - include_tasks: 'el{{ ansible_distribution_major_version }}.yml'
 
 - name: reboot vm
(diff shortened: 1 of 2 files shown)


roles/fips/tasks/main.yml
```diff
@@ -1,17 +1,11 @@
 ---
 - include_tasks: 'el{{ ansible_distribution_major_version }}.yml'
+  when: not ansible_fips
```
ekohl, line 3: Is this a built in?

evgeni: yeah, since ansible 2.4 or so: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts/system/fips.py

ekohl (approved):

PR #1464, opened 2022-02-01
enable katello:el8 module for 4.3+ instead of module_hotfixes

evgeni:


ekohl (approved):
Technically speaking it could be a single module that's enabled since katello should pull in foreman as a dependency.

I'd also like it if we didn't hardcode el8 and instead use the OS major version, but I suppose el9 is still a bit in the future.

evgeni:
yeah, it certainly can get some cleanup, I'll put that on my todo

PR #1465, opened 2022-02-02
only enable katello module on non-foreman installs

evgeni:
katello repo version var is always defined in our pipes

Fixes: 4a31dd26e81f7df67de3e26d752c7bfaf32b0f20


ekohl (approved):
I should have said something in my review but I assumed it would be ok.

PR #1469, opened 2022-02-08
Drop Foreman 1.24 - 2.3 & CentOS 8 boxes

ekohl:
Foreman 1.24 - 2.3 are unsupported and no longer needed for upgrade testing.

CentOS 8 has been removed from the mirrors and has moved to the vault. This means they no longer provision without additional work. In 2.4 it's kept in since 2.4 didn't support Stream officially but it at least indicates it was built on EL8.


ekohl:
Looks like the tests need to be modified. Should we also drop the < 2.1 compatibility code?

ehelms:
Replacement -- https://github.com/theforeman/forklift/pull/1497

PR #1477, opened 2022-03-21
allow ansible_repositories on other RH family OSes

evgeni:
we need this on Alma and Rocky too


evgeni:
damn, ansible-lint 6 is picky

ekohl (approved):

PR #1478, opened 2022-03-21
add support for almalinux and rocky in os_helper

evgeni:


ekohl (approved):
I think we can safely ignore ansible-lint for this one.

PR #1484, opened 2022-04-06
Foreman templates git import test

adamruzicka:


ekohl (approved):
:+1: assuming the SELinux fix is cherry picked.

adamruzicka:
SELinux fix CPs just went in.

PR #1487, opened 2022-04-22
remame foreman_proxy::plugin::remote_execution::{ssh,script}

wbclark:
this was changed in puppet-foreman_proxy with https://github.com/theforeman/puppet-foreman_proxy/pull/734/commits/3a35fc75511d559f0470040bbc024e46ce3ba081

### DIFF
--- a/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
+++ b/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
@@ -35,5 +35,5 @@ foreman_proxy::plugin::dynflow: true
 foreman_proxy::plugin::monitoring: false
 foreman_proxy::plugin::omaha: false
 foreman_proxy::plugin::openscap: false
-foreman_proxy::plugin::remote_execution::ssh: true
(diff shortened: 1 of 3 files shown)


evgeni (approved):

roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
```diff
@@ -35,5 +35,5 @@ foreman_proxy::plugin::dynflow: true
 foreman_proxy::plugin::monitoring: false
 foreman_proxy::plugin::omaha: false
 foreman_proxy::plugin::openscap: false
-foreman_proxy::plugin::remote_execution::ssh: true
+foreman_proxy::plugin::remote_execution::script: true
```
wbclark, line 38: this will break older versions so I think I need to use the templating to make change only when `if foreman_provisioning_foreman_version == 'nightly'`

evgeni: *devel* ;)

PR #1490, opened 2022-04-22
rename rex-ssh to rex-script only for recent versions

wbclark:


evgeni:
`pipeline_version` is undefined outside of pipelines, and that's where these playbooks are used :(

wbclark:
> `pipeline_version` is undefined outside of pipelines, and that's where these playbooks are used :(

Will `foreman_repositories_version` suffice? If that is not defined, how is does it determine which version to install?

roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
```diff
@@ -35,5 +35,5 @@ foreman_proxy::plugin::dynflow: true
 foreman_proxy::plugin::monitoring: false
 foreman_proxy::plugin::omaha: false
 foreman_proxy::plugin::openscap: false
-foreman_proxy::plugin::remote_execution::script: true
+"{{ 'foreman_proxy::plugin::remote_execution::script: false' if (foreman_installer_version == 'nightly' or foreman_installer_version is version('4.5', '>=')) else 'foreman_proxy::plugin::remote_execution::ssh: false' }}"
```
evgeni, line 38: this template is for devel scenario only, which implies nightly.

playbooks/luna.yml
```diff
@@ -21,6 +21,6 @@
       - "--enable-foreman-proxy-plugin-ansible"
       - "--enable-foreman-proxy-plugin-discovery"
       - "--enable-foreman-proxy-plugin-openscap"
-      - "--enable-foreman-proxy-plugin-remote-execution-ssh"
+      - "{{ '--enable-foreman-proxy-plugin-remote-execution-script' if (foreman_repositories_version == 'nightly' or foreman_repositories_version is version('4.5', '>=')) else '--enable-foreman-proxy-plugin-remote-execution-ssh' }}"
```
evgeni, line 24: this playbook is only used for `centosX-luna` which isn't versioned and implies nightly.

evgeni (commented):
I think I'd just keep them on "script", don't think anyone uses *these* with non-nightly.

wbclark:
@evgeni is this one no longer needed?

evgeni:
correct

PR #1499, opened 2022-05-12
Add Regexp as permitted class for Psych

fao89:
When updating to fedora 36, we get the following error:
```
Vagrant failed to initialize at a very early stage:

There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /home/faguiard/project/pulp_installer/Vagrantfile
Line number: 0
Message: Psych::DisallowedClass: Tried to load unspecified class: Regexp
```

https://stackoverflow.com/questions/71332602/upgrading-to-ruby-3-1-causes-psychdisallowedclass-exception-when-using-yaml-lo

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,6 +49,7 @@ jobs:
           - '2.6'
           - '2.7'
           - '3.0'
+          - '3.1'
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


evgeni:
Why would it end up trying to get a regexp there to begin with?

fao89:
Oh, the issue is on pulp_installer, it is getting the Regexp from here:
https://github.com/pulp/pulp_installer/blob/main/vagrant/boxes.d/00-fedora.yaml#L11

PR #1501, opened 2022-05-12
always use YAML.safe_load to load YAML

evgeni:
and pass `permitted_classes: [Regexp]` when loading boxes, as we need
regexp when using cloudy backends


fao89 (approved):
it worked!

Thank you!

ehelms (approved):

PR #1502, opened 2022-05-12
Add role to spin up a basic external postgresql database

ehelms:
This is pseudo-draft status. It could go in as is but I am hoping to take it for a spin against a Katello at least. My intent for this is to make it easier to test external database use cases (maybe pipeline it one day?).

### DIFF
--- a/playbooks/fips.yml
+++ b/playbooks/fips.yml
@@ -1,6 +1,6 @@
 ---
 - hosts: all
   become: yes
-  gather_facts: no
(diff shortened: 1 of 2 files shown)


evgeni:
Why bother adding el7 support? The os of the db should have little to no influence on the result of the setup.

ehelms:
This is ready for further review, I have added a variety of features to it as well to allow spinning up a database with either md5 or scram-sha-256.

PR #1515, opened 2022-06-23
Use 'python3' interpreter specifically

pcreech:
On distributions without a default 'python' interpreter,
the inventory script will fail to execute properly preventing host matching

Failure was noticed on RHEL8

### DIFF
--- a/inventories/vagrant.py
+++ b/inventories/vagrant.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # Adapted from Mark Mandel's implementation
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
Given we're killing EL 7 I think it's fine to move to Python 3 only. However, I would prefer to also kill the Python 2 compat code, which is the import and DEVNULL bit. For the latter I'd just use subprocess.DEVNULL everywhere instead of assigning a local variable.

ekohl:
Sounds like something we should fix as well, or ensure python3 is installed on EL7 but it'd be better to get rid of EL7 there too.

PR #1516, opened 2022-06-23
add the opentelemetry callback

evgeni:


ekohl (commented):
How does this affect users running it locally?

evgeni:
As long as they don't have the opentelemetry python sdk installed *and* have telemetry environment variables (endpoint, token) set, it's NOOP (well, a warning and then a noop)

ehelms (approved):

PR #1523, opened 2022-06-30
Use development tuning profile for devel

ehelms:
Depends on the outcome of https://github.com/theforeman/foreman-installer/pull/788


ehelms:
The mentioned PR has been merged, so development installations will now have access to the development profile. That is, this can be review now.

PR #1525, opened 2022-07-01
hack around python vs python3 binary names

evgeni:
this is a hack! please remove once we don't need any python2 support
anymore and replace with a python3 shebang


evgeni:
cc @pcreech

ehelms (approved):
Mind opening an issue that points to this to increase visibility?

evgeni:
#1527 it is

PR #1528, opened 2022-07-01
Use python 2 and 3 compatible split()

ehelms:
Python 2 does not support maxsplit keyword on the split method.
This change can be reverted once our environments are off of EL7
based infra.


ehelms:
@Odilhao This should fix the Pulpcore pipeline error.
@evgeni For awareness.

evgeni:
me--

evgeni (approved):

Odilhao (approved):

PR #1530, opened 2022-07-04
install telemetry deps when seting up forklift

evgeni:


evgeni:
> I guess you intend to set `forklift_telemetry` in `jenkins-jobs`.

this is exactly the intent here, yes.

PR #1531, opened 2022-07-04
support psych before 4.0

evgeni:


vagrant/lib/forklift/box_factory.rb
```diff
@@ -42,7 +42,11 @@ def filter_boxes!
 
     def load_box_file(file)
       file = File.read(file)
-      YAML.safe_load(ERB.new(file).result, permitted_classes: [Regexp])
+      if Gem::Version.new(Psych::VERSION) < Gem::Version.new('4.0')
```
ekohl, line 45: In other places I checked `RUBY_VERSION >= '2.7'`. Ruby < 2.7 doesn't allow keyword arguments this way, but 2.7 does.

ekohl (approved):

vagrant/lib/forklift/box_factory.rb
evgeni: But the param comes from Psych, so I prefer this ;)

ekohl: The param is already there in older versions. Ruby 2.7:
```console
$ scl enable tfm ruby <<EOF
> require 'yaml'
> puts Psych::VERSION
> puts YAML.safe_load('{}', permitted_classes: [Regexp])
> EOF
3.1.0
{}
```

I don't know which version added permitted_classes, but that's been there for a long time AFAIK.

evgeni: 3.1.0 renamed `whitelist_classes` to `permitted_classes`: https://github.com/ruby/psych/commit/682abf20f0ddbc8d83e816022953536dad80c05b
4.0.0 made it "mandatory": https://github.com/ruby/psych/commit/0767227051dbddf1f949eef512c174deabf22891

ekohl: Psych 4 has `ruby_version >= 2.4` so you could have an incompatible setup this way, but I'm ok with waiting until this hits us to fix it.

PR #1532, opened 2022-07-08
Skip ostree bats tests if not enabled

ehelms:
While this pattern could be applied more broadly, this is to tackle this specific use case first and then follow ups could organize better and make use of it.

### DIFF
--- a/bats/fb-katello-content.bats
+++ b/bats/fb-katello-content.bats
@@ -77,6 +77,7 @@ setup() {
   fi
 
   tSkipIfOlderThan43
+  tSkipUnlessContentType 'ostree'
 
   hammer repository create --organization="${ORGANIZATION}" --url=https://fixtures.pulpproject.org/ostree/small/ \
     --product="${PRODUCT}" --content-type="ostree" --name "${OSTREE_REPOSITORY}" | grep -q "Repository created"
@@ -88,7 +89,8 @@ setup() {
   fi
 
   tSkipIfOlderThan43
-  
+  tSkipUnlessContentType 'ostree'
+
   hammer repository synchronize --organization="${ORGANIZATION}" \
     --product="${PRODUCT}" --name="${OSTREE_REPOSITORY}"
 }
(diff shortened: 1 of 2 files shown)


pcreech (approved):

ehelms:
Output from testing it:

```
 ✓ Create an Empty Organization
 ✓ create an Organization
 ✓ create a product
 ✓ create package repository
 ✓ upload package
 ✓ sync repository
 ✓ create a file repository
 ✓ sync file repository
 ✓ fetch file from file repository
 ✓ create a container repository
 ✓ sync container repository
 - create an ostree repository (skipped: Content type ostree is not enabled)
 - sync ostree repository (skipped: Content type ostree is not enabled)
 - create puppet repository (skipped: Puppet content is not available in scenarios with only Pulp 3)
 - upload puppet module (skipped: Puppet content is not available in scenarios with only Pulp 3)
 - upload ostree_ref (skipped: Content type ostree is not enabled)
```

ekohl (commented):
I always worry about these kind of tests. How do you know if the ostree content is ever really tested and something didn't just fail? Do we have a positive test which fails if ostree isn't present?

In forklift we have `--foreman-proxy-content-enable-ostree` which is authoritative. Can we somehow utilize that: tell forklift to expect ostree in some cases?

ehelms:
> I always worry about these kind of tests. How do you know if the ostree content is ever really tested and something didn't just fail? Do we have a positive test which fails if ostree isn't present?
> 
> In forklift we have `--foreman-proxy-content-enable-ostree` which is authoritative. Can we somehow utilize that: tell forklift to expect ostree in some cases?

To achieve that, it likely needs more thought and break-down of the tests given that it could be the scenario itself defining what is enabled or disabled which makes it hard to do a blanket enablement for a test run in those scenarios.

bats/fb-katello-content.bats
```diff
@@ -77,6 +77,7 @@ setup() {
   fi
 
   tSkipIfOlderThan43
+  tSkipUnlessContentType 'ostree'
```
ekohl, line 80: I was wondering if it could be more efficient and only check it once, but I think `setup_file` isn't in some older bats versions we may test with.

ekohl (commented):
Just thinking out loud (because I thought about a similar case in the past just for smoker): you can have an env var (like EXPECT_OSTREE) where missing means auto-detection, YES means failure if it isn't present and NO means failure if it is present.

bats/fb-katello-content.bats
evgeni: Given EL7 is dead, we could just stop installing bats via package (from epel) and always use git?

ekohl: I thought about the same thing. Probably a good thing for a new PR.

ehelms:
> Just thinking out loud (because I thought about a similar case in the past just for smoker): you can have an env var (like EXPECT_OSTREE) where missing means auto-detection, YES means failure if it isn't present and NO means failure if it is present.

I can work with that. I am going to opt to take in this change as is to unblock and then work on this concept since it will take a few iterations of testing to make sure I got it right.

PR #1541, opened 2022-07-19
Updating Pulpcore to 3.15.9-1

Odilhao:
With the merge of #1539 we got hit by another issue with the install for EL7, EL8 now is working.

3.15.9-1 brings a fix for [ #pulp_installer/pull/914](https://github.com/pulp/pulp_installer/pull/914) that aims to fix the EL7 install Issue

### DIFF
--- a/requirements-pulp.yml
+++ b/requirements-pulp.yml
@@ -1,3 +1,3 @@
 collections:
   - name: pulp.pulp_installer
-    version: 3.15.9
+    version: 3.15.9-1


Odilhao:
Just tested locally, and managed to install 3.17 on Centos7

Odilhao:
Also worked on 3.15

Odilhao:
Had to make the `pulp_pkg_undeclared_deps` empty for >3.15 installations.

pipelines/pulpcore/02-install.yaml
```diff
@@ -3,6 +3,9 @@
   hosts:
    - "{{ forklift_server_name }}"
   become: True
+  vars:
+    pulp_pkg_undeclared_deps: '{}'
```
evgeni, line 7: a string with `{}` in it? why

Odilhao: I thought that I could just empty the list `{}`

evgeni: You can, but then it needs to be an empty list (`[]`) or an empty dictionary (`{}`), not a string `'{}'`

PR #1542, opened 2022-07-21
Only set --tuning on katello scenario for foreman_installer role

ehelms:
To fix nightly breakage.

### DIFF
--- a/roles/foreman_installer/tasks/install.yml
+++ b/roles/foreman_installer/tasks/install.yml
@@ -25,7 +25,7 @@
       {{ foreman_installer_command }} {{ (foreman_installer_verbose|bool) | ternary("-v", "") }}
       {{ (foreman_installer_no_colors|bool) | ternary("--no-colors", "") }}
       {{ (foreman_installer_version is version('3.4', '<') and foreman_installer_disable_system_checks|bool) | ternary("--disable-system-checks", "") }}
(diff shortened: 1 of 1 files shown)


roles/foreman_installer/tasks/install.yml
```diff
@@ -25,7 +25,7 @@
       {{ foreman_installer_command }} {{ (foreman_installer_verbose|bool) | ternary("-v", "") }}
       {{ (foreman_installer_no_colors|bool) | ternary("--no-colors", "") }}
       {{ (foreman_installer_version is version('3.4', '<') and foreman_installer_disable_system_checks|bool) | ternary("--disable-system-checks", "") }}
-      {{ (foreman_installer_version is version('3.4', '>=') and foreman_installer_disable_system_checks|bool) | ternary("--tuning development", "") }}
+      {{ (foreman_installer_version is version('3.4', '>=') and foreman_installer_disable_system_checks|bool and foreman_installer_scenario in ['katello']) | ternary("--tuning development", "") }}
```
ekohl, line 28: You use `in` because you foresee other scenarios doing the same?

ekohl (approved):

roles/foreman_installer/tasks/install.yml
ehelms: Aye, thinking of bringing tuning to other scenarios so leaving the door open for now.

PR #1544, opened 2022-07-25
Drop Ansible 2.9 repository setup

ekohl:
ansible-core is in CentOS Stream which means Ansible 2.9 can no longer be installed. Setting up this repository only slows things down.

Submitting this as a draft PR. I still need to verify this works, but I'll use this PR to get it to my desktop.


ekohl:
On an existing installation I tried `dnf remove centos-release-ansible-29 -y && dnf list extras`. Nothing came up.

IIRC ansible-core also became part of RHEL 8.6.

ehelms (approved):

PR #1547, opened 2022-08-02
Make forklift compatible with ansible-core

ekohl:
The goal is to get our CI testing on CentOS CI to work with ansible-core.

Right now this is a draft, but I'll continue iterating on this until I get a complete run.

### DIFF
--- a/playbooks/setup_forklift.yml
+++ b/playbooks/setup_forklift.yml
@@ -31,6 +31,16 @@
           - rubygem-deep_merge
         state: 'present'
 
+    - name: 'install telemetry system dependencies'
+      package:
+        name:
+          - python3-devel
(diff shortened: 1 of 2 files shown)


playbooks/setup_forklift.yml
```diff
@@ -31,6 +31,16 @@
           - rubygem-deep_merge
         state: 'present'
 
+    - name: 'install telemetry system dependencies'
```
ekohl, line 34: I'm not sure this is the right solution. Right now it's compiling the whole SDK which takes a few minutes. In CI we use wheels, so that's probably better.

ekohl:
I'd like to merge this now so we can at least move forward.

ehelms (approved):

PR #1551, opened 2022-08-05
Foreman 3.4

pcreech:


vagrant/config/versions.yaml
```diff
@@ -74,6 +74,16 @@ installers:
       - 'debian11'
       - 'ubuntu2004'
 
+  - foreman: '3.4'
+    katello: 'nightly'
```
ekohl, line 78: Please set this correctly

ekohl (changes requested):

vagrant/config/versions.yaml
pcreech: Done.  I wasn't sure if I should use nightly or 4.6, due to the order of operations (i.e. use nightly first, then update when 4.6 is branched, or just use 4.6 from the start).

ekohl (approved):
Merging this will break upgrade pipelines until 3.4.0-rc1 is available. Please do so just before you run the release pipelines.

PR #1552, opened 2022-08-08
Remove attach subscription to AK test since orgs are in SCA by default

chris1984:
Since the new Candlepin 4.2 is tagged into nightly, by default orgs are in SCA mode. This means that you can not attach a subscription to an activation key anymore. 

After discussion with @parthaa and @jeremylenz it was decided since entitlement mode is being depreciated and eventually removed, that removing this test is fine.

### DIFF
--- a/bats/fb-katello-content.bats
+++ b/bats/fb-katello-content.bats
@@ -297,6 +297,8 @@ setup() {
 }
 
 @test "add subscription to activation key" {
+  tSkipIfNewerthan45
+
   sleep 10
   activation_key_id=$(hammer --csv --no-headers activation-key info --organization="${ORGANIZATION}" \
     --name="${ACTIVATION_KEY}" | cut -d, -f2)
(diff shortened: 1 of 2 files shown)


evgeni:
This will also remove the test fron older Katello releases (which don't have SCA by default).

If you say that's a risk that you accept, we're good. Just wanted to explicitly point out the multi-version implication.

chris1984:
> This will also remove the test fron older Katello releases (which don't have SCA by default).
> 
> If you say that's a risk that you accept, we're good. Just wanted to explicitly point out the multi-version implication.

Is there a way to specify a release? I guess we could do if candlepin < 4.2 then run the test or something similar, would that work?

evgeni:
We have the possibility to say "skip if <condition>", like here:

https://github.com/theforeman/forklift/blob/8acf01f01809459f123d754e910b7a36f4ee3b1c/bats/fb-katello-content.bats#L157
https://github.com/theforeman/forklift/blob/8acf01f01809459f123d754e910b7a36f4ee3b1c/bats/foreman_helper.bash#L31-L36

So we could do something like "skip if newer than 4.5"?

chris1984:
@evgeni how does this look?

ekohl (approved):

PR #1553, opened 2022-08-08
Allow specifying custom user and group in devel scenario

wbclark:
puppet-katello_devel supports these options but passing them as installer args doesn't set the certs::group parameter. this approach creates a working installation by also properly setting certs::group so that it is inherited by the apache certificate.

### DIFF
--- a/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
+++ b/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
@@ -1,11 +1,11 @@
 ---
 certs:
-  group: vagrant
+  group: {{ foreman_installer_devel_scenario_group }}
   deploy: true
   generate: true
 katello_devel:
-  deployment_dir: /home/vagrant
-  user: vagrant
+  deployment_dir: /home/{{ foreman_installer_devel_scenario_user }}
+  user: {{ foreman_installer_devel_scenario_user }}
 foreman_proxy_content:
   enable_katello_agent: true
   proxy_pulp_yum_to_pulpcore: true
--- a/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
+++ b/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
@@ -1,6 +1,6 @@
 ---
 certs:
-  group: {{ foreman_installer_devel_scenario_group }}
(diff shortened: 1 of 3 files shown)


wbclark:
This works for me when I test with the following box definition:

```
centos8-nodejs14-katello-devel:
  primary: true
  box: centos8-stream
  ansible:
    playbook:
      - 'playbooks/katello_devel.yml'
    group: 'devel'
    variables:
      ssh_forward_agent: true
      foreman_devel_github_push_ssh: true
      katello_devel_github_username: wbclark
      katello_repositories_environment: staging
      katello_devel_user: wclark
      katello_devel_group: wclark
      foreman_installer_options:
        - "--foreman-proxy-content-enable-ostree=true"
        - "--katello-devel-npm-timeout=3600"
        - "--katello-devel-modulestream-nodejs=14"
```

roles/katello_devel/meta/main.yml
```diff
@@ -11,6 +11,8 @@ dependencies:
   - role: postgresql_scl
     when: ansible_distribution_major_version == "7"
   - role: foreman_installer_devel_scenario
+    foreman_installer_devel_scenario_user: "{{ katello_devel_user if katello_devel_user is defined else 'vagrant' }}"
+    foreman_installer_devel_scenario_group: "{{ katello_devel_group if katello_devel_group is defined else 'vagrant' }}"
```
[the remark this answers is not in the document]
wbclark: yes, for sure

roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml
```diff
@@ -1,11 +1,11 @@
 ---
 certs:
-  group: vagrant
+  group: {{ foreman_installer_devel_scenario_group }}
```
ekohl, line 3: Thinking more: should we even have a default for the group or should it default to the user instead:
```suggestion
  group: {{ foreman_installer_devel_scenario_group | default(foreman_installer_devel_scenario_user) }}
```

wbclark: I like the suggestion, because it allows the user to pass one less variable in the case where the user and group should match (probably most of the time) but still allows for the full customization in case it is required.

ekohl (approved):

PR #1554, opened 2022-08-09
disable sync folder for CI

evgeni:


ekohl:
I thought it was weird that Rubocop passes on Ruby 3.1, but not on older versions. Then I saw this:
https://github.com/theforeman/forklift/blob/0753db111d5c5262c17db240485925f8b8730c63/vagrant/Gemfile#L14-L18

ekohl (approved):

PR #1558, opened 2022-08-09
only check for product subs if no SCA enabled

evgeni:


evgeni:
cc @chris1984 @parthaa -- would you mind double checking my logic here?

currently, nightly fails with:

     `subscription-manager list --consumed | grep "${PRODUCT}"' failed

on SCA-by-default enabled Candlepin

jeremylenz (approved):
Thanks @evgeni! LGTM :+1:

chris1984 (approved):
Looks correct to me,

PR #1562, opened 2022-08-22
Remove checks related to no longer supported Vagrant versions

wbclark:
These checks were required in CI back in 2017. Since we
only support Vagrant 2.2+ presently, they can be safely
removed.

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -7,11 +7,6 @@ module Forklift
 
     VAGRANTFILE_API_VERSION = '2'.freeze
 
-    if Gem.loaded_specs['vagrant']
-      SUPPORT_NAMED_PROVISIONERS = Gem.loaded_specs['vagrant'].version >= Gem::Version.create('1.7')
-      SUPPORT_BOX_CHECK_UPDATE = Gem.loaded_specs['vagrant'].version >= Gem::Version.create('1.5')
-    end
-
     def initialize(boxes)
       @ansible_groups = {}
       @boxes = boxes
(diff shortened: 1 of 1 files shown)


vagrant/lib/forklift/box_distributor.rb
```diff
@@ -184,12 +179,8 @@ def configure_ansible(machine, ansible, box_name)
...
       [playbooks].flatten.each_with_index do |playbook, index|
-        if SUPPORT_NAMED_PROVISIONERS
-          args = ["main#{index}"]
-          kwargs = { type: 'ansible' }
-        else
-          args = [:ansible]
-        end
+        args = ["main#{index}"]
+        kwargs = { type: 'ansible' }
         machine.vm.provision(*args, **kwargs) do |ansible_provisioner|
```
ekohl, line 184:
```suggestion
        machine.vm.provision("main#{index}", type: 'ansible') do |ansible_provisioner|
```

ekohl (approved):
I recall last time we had issues with Vagrant on EL7, mostly regarding downstream. I'm not sure those are still an issue. @ehelms knew last time.

I'd also be fine merging this and letting it break.

ehelms (approved):

evgeni:
Even on EL7 we deploy 2.2.x:

https://github.com/theforeman/forklift/blob/5683c94fb95fca963f98512d5094c4b7e964c52d/roles/vagrant/defaults/main.yml#L3

vagrant/lib/forklift/box_distributor.rb
wbclark: TY

wbclark:
If someone would kindly merge this, I do not have the permissions to do so

PR #1563, opened 2022-08-22
Rename unpriviledged_user role to unprivileged_user

wbclark:
The spelling is corrected with a redirect and deprecation
notice in place to maintain backwards compatibility while
informing users who use this role in their playbooks.


wbclark:
NOTE: I need to make some other enhancements to this role and figured tackling this first then building on top of the renamed role was the best approach. That said, I could also simply create a new role to serve my needs and leave this one unchanged if that approach is preferred.

roles/unpriviledged_user/tasks/main.yml
```diff
@@ -1,62 +1,11 @@
...
-  become: true
+- name: Print message informing the user that that the role has been renamed
+  ansible.builtin.debug:
+    msg: "Deprecation Notice: this role has moved! Please use roles/unprivileged_user instead."
+
+- name: Include moved role from new location
+  ansible.builtin.include_role:
+    name: unprivileged_user
+  vars:
+    unprivileged_user_username: "{{ unpriviledged_user_username | default('vagrant') }}"
```
wbclark, line 10: While `roles/unpriviledged_user/defaults/main.yml` is removed, the behavior is preserved by setting the default here.

ehelms (commented):
I think you are safe to just rename the role and not keep around the compatibility. As we do not version Forklift, there is no guarantee when we'd clean up that compatibility layer and the change for any user of it seems small enough in their environment to just make the change.

wbclark:
> I think you are safe to just rename the role and not keep around the compatibility. As we do not version Forklift, there is no guarantee when we'd clean up that compatibility layer and the change for any user of it seems small enough in their environment to just make the change.

Thanks for the suggestion. I pushed an additional commit which removes the previous name entirely.

PR #1568, opened 2022-08-26
Add documentation using sphinx

ekohl:
Example: https://ekohl.github.io/forklift/

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,42 @@
+name: Deploy docs to Pages
+
+on:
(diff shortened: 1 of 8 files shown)


ehelms:
> An alternative to sphinx may be https://www.mkdocs.org/ which is lighter than sphinx and used within CentOS and Fedora.

This looks nice since we already use markdown for our docs. I'd say give that a go.

ekohl:
Ok, replacement @ https://github.com/theforeman/forklift/pull/1582

PR #1570, opened 2022-08-31
Unprivileged user groups

wbclark:


roles/unprivileged_user/tasks/main.yml
```diff
@@ -1,17 +1,29 @@
...
+- name: "Create the {{ unprivileged_user_username }} user"
   ansible.builtin.user:
     name: "{{ unprivileged_user_username }}"
-    comment: "John Vagrant"
+    groups: "{{ unprivileged_user_additional_groups + [unprivileged_user_groupname] }}"
+    append: yes
   become: true
 
-- name: "Add {{ unprivileged_user_username }} to sudoers"
+- name: "Grant passwordless sudo via {{ unprivileged_user_groupname }} group"
```
ekohl, line 20: Can you explain this change? Why would you grant it via the group if the role is to create an unprivileged **user**?

From a design point of view I would rather expect this as a role `unprivileged_group` and change `unprivileged_user` to become part of that group. Or, since the whole role becomes just a tiny wrapper for `user`, drop the whole role?

 Some motivation why this is beneficial would be appreciated.

wbclark: I was initially looking at the standard `wheel` group pattern but that's RH/Fedora only. Ubuntu has `admin`. So this seemed like a general way to allow granting this access to multiple users if that is ever desired.

I'm not strongly set on that approach, it could be granted by the username only. I still _need_ the capability to add multiple groups to the user though, because that is required to build hotfix RPMs, and a portable hotfix environment is the end goal.

ekohl: I think you should question at which point the role becomes too complex and try to do too many things. At this point the module is already hard to follow. It may be easier to first start using https://docs.ansible.com/ansible/latest/collections/ansible/posix/authorized_key_module.html + https://docs.ansible.com/ansible/latest/collections/community/general/sudoers_module.html and then see what other changes are needed. If you have those modules, it may be easier to write a fresh role for your use case.

wbclark: I have a working local branch with a `devel_user` role which uses the authorized_keys module, but when I saw how much overlap exists with this role `unprivileged_user` I decided to look at updating that instead. I didn't include any switch to the authorized_keys module yet due to the way this role falls back to copying `/root/.ssh/authorized_keys` but I could do so.

I can open an additional PR with the separate role as an alternative so that you can more easily compare the approaches. There really is a large amount of overlap, and the only change to _default/existing_ behavior was sudo access via the group rather than the user, which I don't really need -- I just went that route because it's how I usually see it done.

ekohl: The goal of this role was originally to mimic what vagrant did.

I'll also admit that I don't really use this at all. All my development is in a persistent host system. I'm starting to use containers more where needed.

That means I won't block on this, but forklift has always been at great risk of becoming an unmaintainable mess where nobody knows what certain parts do but they're afraid of touching them because they fear breaking them. This has that potential too.

wbclark:
Thanks for the recommendations @ekohl 

I've updated with the changes to community.general.sudoers and ansible.posix.authorized_key

PR #1571, opened 2022-09-01
Unprivileged user kerberos

wbclark:
This builds on top of #1570 and adds capabilities to optionally install kerberos client packages, copy kerberos configuration from the host system, and copy credential cache from the host system. These features are opt-in; they are disabled by default.

### DIFF
--- a/roles/unprivileged_user/tasks/main.yml
+++ b/roles/unprivileged_user/tasks/main.yml
@@ -27,24 +27,20 @@
     create: yes
   become: true
 
-- name: "Create {{ unprivileged_user_username }} .ssh"
-  ansible.builtin.file:
-    path: "/home/{{ unprivileged_user_username }}/.ssh"
-    owner: "{{ unprivileged_user_username }}"
-    group: root
-    state: directory
-    mode: 0700
+- name: "Add public key to authorized_keys from Host Machine"
+  ansible.posix.authorized_key:
+    user: "{{ unprivileged_user_username }}"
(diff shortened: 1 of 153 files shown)


roles/unprivileged_user/tasks/main.yml
```diff
@@ -66,3 +66,40 @@
...
+        state: present
+      when: ansible_os_family == "RedHat"
+
+    - name: "Install krb5-user on Debian based distributions"
+      ansible.builtin.apt:
+        name: "krb5-user"
+        state: present
+      when: ansible_os_family == "Debian"
+
+    - name: "Copy Kerberos client config from Host"
```
ekohl, line 86: This seriously feels out of scope for this role. Roles should be small and focused. When I read `unprivileged_user` I don't think it essentially means "reconfigure the whole system".

What I'd do is create a role to make the system a kerberos client, with an option to copy from the host. You can then combine those in a playbook. Composition is a great thing and playbooks are the right place for that.

ekohl (changes requested):

roles/unprivileged_user/tasks/main.yml
wbclark: With the default parameter values, the behavior is exactly the same because none of the extra tasks here will run. But I don't mind extracting it out into a separate role to keep things organized.

wbclark:
Closing in favor of #1574

PR #1572, opened 2022-09-01
Add an inherit_timezone role to copy timezone from host

wbclark:


ekohl (commented):
A bit of bikeshedding, but should the role be named `timezone_from_host`?

wbclark:
> A bit of bikeshedding, but should the role be named `timezone_from_host`?

Sure. After thinking about it, I went with `inherit_timezone`. That way similar host->guest inheritance roles could have a similar naming scheme, e.g. `inherit_kerberos`, and be grouped together.

ekohl (approved):
I like inherit.

PR #1573, opened 2022-09-01
Configure ssh.username and ssh.keys_only per box

wbclark:
Also fixes a bug where ssh.forward_agent config
on a single box would apply to all boxes, and
a separate bug where a box configured with
forward_agent = false would always defer to
settings.yaml.

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -60,13 +60,23 @@ module Forklift
       end
     end
 
+    def fetch_from_box_or_settings(box, key)
+      box.fetch(key) { @settings.fetch(key, nil) }
+    end
+
     def define_vm(config, box = {})
       primary = box.fetch('primary', false)
       autostart = box.fetch('autostart', false)
       config.vm.define box.fetch('name'), primary: primary, autostart: autostart do |machine|
         machine.vm.box = box.fetch('box_name', nil)
         machine.vm.box_version = box.fetch('box_version', nil)
-        config.ssh.forward_agent = box.fetch('ssh_forward_agent', nil) || @settings.fetch('ssh_forward_agent', false)
+
+        %w[username forward_agent keys_only].each do |key|
+          unless (value = fetch_from_box_or_settings(box, "ssh_#{key}")).nil?
+            machine.ssh.send("#{key}=", value)
+          end
+        end
+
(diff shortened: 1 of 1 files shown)


vagrant/lib/forklift/box_distributor.rb
```diff
@@ -60,13 +60,31 @@ def distribute!
...
+    def fetch_from_box_or_settings(box, key, default_value)
+      if !(value_from_box = box.fetch(key, nil)).nil?
+        value_from_box
+      else
+        @settings.fetch(key, default_value)
+      end
+    end
+
+    def ssh_config_keys_default_values
+      [%w[username vagrant], ['forward_agent', false], ['keys_only', true]]
```
ekohl, line 72: Feels like this could be a a hash
```suggestion
      {
        username: 'vagrant',
        forward_agent: false,
        keys_only: true,
      }
```

Also, I'd consider not setting the SSH username and relying on Vagrant to take care of that.

vagrant/lib/forklift/box_distributor.rb
```diff
@@ -60,13 +60,31 @@ def distribute!
       end
     end
 
+    def fetch_from_box_or_settings(box, key, default_value)
+      if !(value_from_box = box.fetch(key, nil)).nil?
+        value_from_box
+      else
+        @settings.fetch(key, default_value)
+      end
```
ekohl, line 68: The goal of [fetch](https://ruby-doc.org/core-3.1.2/Hash.html#method-i-fetch) is to provide a default if it isn't set. You can use that to simplify:
```suggestion
      box.fetch(key, @settings.fetch(key, default_value))
```

Or in a block if the computation of the value is expensive/dynamic:

```suggestion
      box.fetch(key) { @settings.fetch(key, default_value) }
```

vagrant/lib/forklift/box_distributor.rb
wbclark: I tried this initially but couldn't find a satisfactory solution. Looking again, I think https://apidock.com/ruby/v2_6_3/Hash/each_pair is what I need

Really, Vagrant should be able to set defaults for all of these. I initially tried to copy what looked like the intent of the existing pattern with `forward_agent` but looking again at https://www.vagrantup.com/docs/vagrantfile/ssh_settings I don't see any reason why we'd need to have defaults here.

I will try a version without them.

ekohl (approved):
Thanks!

PR #1574, opened 2022-09-05
Add an inherit_kerberos role

wbclark:
The role installs kerberos client packages
and optionally copies the configuration
and credential cache from the host machine.


roles/inherit_kerberos/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+inherit_kerberos_config: '/etc/krb5.conf'
+inherit_kerberos_ccache: False
```
evgeni, line 3: I'd not define this as false (it's a path in reality), but check for "is defined" below?

wbclark: Is there a particular reason to prefer one of these patterns over the other?

```yaml
  when: inherit_kerberos_ccache is defined
```

vs.

```yaml
  when: inherit_kerberos_ccache | default(False)
```

I've seen both used pretty frequently and I'm wondering which is "more correct" and why.

evgeni: IMHO only `is defined` is "correct" in the case where `inherit_kerberos_ccache` is not a boolean as such.

If I am reading it correctly, the logic here is: "The user can opt-in to provide additional data (path) to be copied into the VM", which means that the variable is either undefined (the user did not opt-in) or a string/path (the user wants this path to be copied). If Ansible would be type-safe (lol), you wouldn't be able to pass a boolean here, as it's not a string. And the `default(False)` should also break.

It would be different if the variable itself would already denote a boolean (do/don't copy), where you could just set the default to `false` in the defaults file.

wbclark:
I've rewritten this to use the `when: pathname is defined` pattern

roles/inherit_kerberos/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+      - "krb5-workstation"
+      - "krb5-libs"
+    state: present
+  when: ansible_os_family == "RedHat"
+
+- name: "Install client packages on Debian based distributions"
+  ansible.builtin.apt:
+    name: "krb5-user"
+    state: present
+  when: ansible_os_family == "Debian"
```
evgeni, line 14: you culd use `ansible.builtin.package` which works across distros, and load the correct list of packages using a vars file.

https://github.com/theforeman/forklift/blob/12c6385b216e77dc4565d5ef060deae3745cd24e/roles/bats/tasks/main.yml#L3
https://github.com/theforeman/forklift/blob/12c6385b216e77dc4565d5ef060deae3745cd24e/roles/bats/vars/RedHat.yml#L2

Not saying you *have* to, but something to consider :)

roles/inherit_kerberos/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    owner: root
+    group: root
+    mode: '0644'
+  when: inherit_kerberos_config is defined
+
+- name: "Copy Kerberos credential cache from Host"
+  ansible.builtin.copy:
+    src: "{{ inherit_kerberos_ccache }}"
+    dest: "{{ inherit_kerberos_ccache }}"
+    owner: "{{ inherit_kerberos_local_user_name }}"
```
evgeni, line 29: where does `inherit_kerberos_local_user_name` come from? I don't see it defined anywhere.

evgeni:
Added some minor comments. Would be cool if there would be a few lines in the readme, how to use this role?

roles/inherit_kerberos/tasks/main.yml
wbclark: This was originally written to be part of the `unprivileged_user` role and the value of `unprivileged_user_username` was passed for the owner. The approach was to give `unprivileged_user` additional features to have more general user creation/setup powers, while leaving the default behavior essentially the same in setting up the `vagrant` user for images that don't already have it; therefore the assumption at that time was that this block does not run by default, so an user passing the extra stuff required to run these extra bits would be doing it to setup their personal or devel user and not the default `vagrant` user in that role.

So with the request to split this out into more of a standalone role, instead I assume that the value will be provided by whatever role or playbook includes this role.

evgeni: Ah, got it!

wbclark:
@evgeni some additional context --

The goal of these efforts is to create a portable hotfix environment which is fully encoded in a block of yaml so that other developers can simply copy & paste the box, edit as needed, and `vagrant up`

I could add a very basic `README.md` for the role if that is what we need to get it merged. The lack of documentation in this project is a real concern. If we do not also follow through on documenting the other roles then I am not sure whether it is worth documenting this one, and I don't have tons of time to devote to that larger effort at present.

What do you think?

ekohl:
> I could add a very basic `README.md` for the role if that is what we need to get it merged. The lack of documentation in this project is a real concern. If we do not also follow through on documenting the other roles then I am not sure whether it is worth documenting this one, and I don't have tons of time to devote to that larger effort at present.

I think a policy where we document new roles is a good start.

On a related matter I opened https://github.com/theforeman/forklift/pull/1568. It doesn't document the ansible roles, but that could be a nice addition.

evgeni (approved):
the existing comments can be adressed later

PR #1575, opened 2022-09-05
Optionally use ssh settings only with vagrant ssh command

wbclark:
This allows bootstrapping the user on a base image, by using
default ssh settings during provisioning and user defined
ssh settings when doing `vagrant ssh`.

TESTING INSTRUCTIONS:

1. Create `user_playbooks/ssh-test.yml` to bootstrap the user (edit the usernames to match those you will test):

```yaml
---
- hosts: all
  become: true
  vars:
    unprivileged_user_username: wclark
    unprivileged_user_import_ssh_pub_key_github: wbclark

    - name: "Include unprivileged_user role"
      ansible.builtin.include_role:
        name: unprivileged_user
```

NOTE: As written, this requires the feature to add the user's ssh public key to authorized_keys directly from github via their github username, from #1570 

2. Add the following within `vagrant/boxes.d`:

```yaml
fedora34-ssh-test:
  box: fedora34
  ssh_forward_agent: true
  ssh_username: wclark
  ssh_keys_only: false
  use_ssh_settings_vagrant_ssh_only: true
  ansible:
    playbook:
      - 'user_playbooks/ssh-test.yml'
```

3. Make sure ssh-agent is started and your private key has been added to it

4. Run `vagrant up fedora34-ssh-test` and observe this still uses the default `vagrant` user to run provisioning tasks, including creating the new user and managing its authorized_keys.

5. Run `vagrant ssh-config fedora34-ssh-test` and observe the `ssh_username` from the box definition defined here, even though it was not used in step 4.

6. Run `vagrant ssh fedora34-ssh-test` and observe that it authenticates as the `ssh_username` with credentials from the ssh-agent.

7. As the `ssh_username` user, logged in on fedora34-ssh-test box, test that the ssh-agent was forwarded to the box by using the stored credentials to authenticate to github: `ssh -T git@github.com`

EXPECTED RESULTS: `vagrant up` uses the default `vagrant` user instead of `ssh_username`, while `vagrant ssh-config` and `vagrant ssh` use `ssh_username`. all above steps are successful

ACTUAL RESULTS (prior to this commit): `vagrant up` fails as it tries to use the `ssh_username` which doesn't exist until after provisioning has been run. It will appear to hang on `waiting for ssh to become available` while adding `VAGRANT_LOG=debug` will show it attempting the connection with the not-yet-existing user.

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -60,8 +60,24 @@ module Forklift
       end
     end
 
-    def fetch_from_box_or_settings(box, key)
-      box.fetch(key) { @settings.fetch(key, nil) }
+    def fetch_setting(box, key, default = nil)
+      box.fetch(key) { @settings.fetch(key, default) }
+    end
+
+    def argv_include_any?(tokens)
+      tokens.any? { |token| ARGV.include?(token) }
+    end
+
+    def add_ssh_setting(box, machine, key)
+      return if (value = fetch_setting(box, "ssh_#{key}")).nil?
+      machine.ssh.send("#{key}=", value)
+    end
+
+    def add_ssh_settings(box, machine, keys)
+      return if fetch_setting(box, 'use_ssh_settings_vagrant_ssh_only', false) && !argv_include_any?(%w[ssh ssh-config])
+      keys.each do |key|
+        add_ssh_setting(box, machine, key)
+      end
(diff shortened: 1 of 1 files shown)


wbclark:
Pushed a fix of a bug I introduced while appeasing the linter. After re-testing it, it is once again working.

kirudag:
Good

PR #1577, opened 2022-09-06
Ensure vagrant up runs sequentially

evgeni:


evgeni:
That's opt-in, `forklift_vagrant_lock` needs to be defined for it to trigger

ekohl (commented):
Why is this needed? Isn't `forklift_name` already unique? If it does, why are there conflicts? If it's not unique, wouldn't a run after this try to create the same boxes?

evgeni:
It is, but that's not the problem.

When run on CI, we always download the base boxes fresh and adding them to libvirt. Those operations use the non-unique base names (`centos8-stream`, `almalinux8` etc) and produce clashes.

And before you ask, yes, this needs to be locked across "all" parallel runs, as `vagrant-libvirt` tries to manage the `default` storage and that also ends up in race conditions.

roles/forklift/tasks/up.yml
```diff
@@ -1,12 +1,36 @@
...
+      when:
+        - forklift_vagrant_lock is defined
+
+    - name: 'Write box file'
+      copy:
+        dest: "{{ forklift_directory }}/vagrant/boxes.d/80-tmp-{{ forklift_name }}.yaml"
+        content: "{{ forklift_boxes | to_yaml }}"
+
+    # using --no-parallel here to avoid problems with libvirt storage volumes
+    # see https://github.com/vagrant-libvirt/vagrant-libvirt/issues/850
```
ekohl, line 25: That issue is closed now. Should we try without `--no-parallel` again?

ekohl (approved):
Ah, that feels like a bug in vagrant-libvirt itself where it should actually create a lock in the plugin. Can you create an issue and link to that in the role?

roles/forklift/tasks/up.yml
evgeni: We probably should, yeah.

evgeni:
> Ah, that feels like a bug in vagrant-libvirt itself where it should actually create a lock in the plugin. Can you create an issue and link to that in the role?

The download part is a bug in vagrant core, the libvirt upload parts are bugs in vagrant-libvirt.

I can open bugs, yes.

PR #1578, opened 2022-09-09
don't pass `some_var|default(omit)` to roles

evgeni:
`omit` only works for modules as they have special code to filter this out, but not for roles. Omit (hah!) passing `omit` in katello_devel and make the user set `foreman_installer_devel_scenario_user` as they require themself.

Otherwise you get errors like `Could not create user __omit_place_holder__67702be7f0289f0f990ab1ec3b121327f51b2d5f` when not setting the variable.

Fixes: 06ffd5ebeaaf9c0cc639e8ec2218a39bbd1ff65d


evgeni:
:eyes: @wbclark

wbclark (approved):

PR #1580, opened 2022-09-09
Add a git_repo role

wbclark:
This role uses a single ansible data structure to manage a set of git repositories with potentially multiple remotes and multiple branches per remote.

### DIFF
--- a/roles/git_repo/tasks/branch.yml
+++ b/roles/git_repo/tasks/branch.yml
@@ -1,0 +1,5 @@
+---
+- name: "Create local repository branch tracking specified remote branch"
+  ansible.builtin.command:
+    cmd: "git branch -f {{ branch }} {{ remote.name }}/{{ branch }}"
+    chdir: "{{ git_repo_repository.dir }}/{{ git_repo_repository.name }}"
--- a/roles/git_repo/README.md
+++ b/roles/git_repo/README.md
@@ -1,0 +1,85 @@
+This role clones a set of git repositories, handling multiple remotes and managing different branches for each remote. It also provides basic support for bundle installs for Ruby projects, and pip installs within a virtual environment for Python projects. The role aims to present a single, simple interface in Ansible variables for managing a collection of locally cloned git repositories.
+
+Role Variables
+--------------
+
+The main data structure for the role is the list of `git_repositories`. Each repository in the list requires the following attributes:
+
+- `name`: The local name of the repository.
+- `dir`: The parent directory to clone the repository from.
+- `remotes`: List of repository remotes, each with `name`, `url`, and `branches`.
+
+Repositories may also have the following optional attributes:
+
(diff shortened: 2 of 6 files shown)


wbclark:
Not yet properly tested, due to difficulty getting #1579 to play nicely with https://github.com/theforeman/forklift/commit/5683c94fb95fca963f98512d5094c4b7e964c52d

wbclark:
I added a README for this role explaining how it is used, and marked it ready for review

ianballou:
I haven't tested this entire thing, but it has been in use with our hotfixing process for over a year now and has been working well.

roles/git_repo/tasks/branch.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: "Create local repository branch tracking specified remote branch"
+  ansible.builtin.command:
```
wbclark, line 3: I've tried to use the git_config module for this sort of thing, which was suggested by the maintainers of the git module as that module doesn't support everything that we need to do here, but branch creation was the one area where I could never git_config to work.

Doing it with the command module smells somewhat hacky, but it's not only worked very reliably, it also ends up being quite a bit simpler.

ianballou (commented):
This is looking good to me! How I tested:

```
[ian@manicotto forklift]$ cat playbooks/repositories_test.yml
---
- hosts: all
  tasks:
    - name: "Setup git repositories"
      ansible.builtin.include_role:
        name: git_repo
      loop: "{{ git_repositories }}"
      loop_control:
        loop_var: repo
      vars:
        git_repo_repository: "{{ repo }}"
```

```
  fedora39:
    box_name: 'fedora/39-cloud-base'
    pty: true
    ansible:
      playbook: 'playbooks/repositories_test.yml'
      variables:
        git_repositories:
          - name: 'foreman'
            dir: '/home/vagrant'
            remotes:
              - name: 'origin'
                url: 'https://github.com/theforeman/foreman.git'
                branches:
                  - 'develop'
          - name: 'katello'
            dir: '/home/vagrant'
            remotes:
              - name: 'origin'
                url: 'https://github.com/Katello/katello.git'
                branches:
                  - 'master'
            python_packages:
              - 'ansible'
              - 'obal'
              - 'tito'
 ```

Only snag I hit was that git wasn't on my box. Should a role be responsible for ensuring certain dependencies are installed? I'm not sure if that's a usual Ansible expectation.

ianballou:
(For future reference, I took that playbook from here: https://github.com/wbclark/forklift/commit/bb6de0d3b7fe24d11730045a2b27c3202ce8221b)

ianballou (approved):
Looks good to me! In the future we can revamp the git role to use the generic package action so that other distros are supported.

evgeni (approved):
As said on Matrix: I think it's too complicated, but I don't mind having it if it's useful for y'all

Mergin.

PR #1581, opened 2022-09-12
Support Fedora 35 and Fedora 36 as guest OS

wbclark:


wbclark:
34 and older are also EOL now. If it is desired, I can remove these as well.

evgeni (approved):
If 34 isn't used by anything anymore, I guess it can be removed, but not a blocker

evgeni:
`vagrant/boxes.d/03-packaging.yaml` is using fedora 33 as base, probably should just be updated

PR #1582, opened 2022-09-22
Add docs using mkdocs

ekohl:
Example: https://ekohl.github.io/forklift/

Replacement for https://github.com/theforeman/forklift/pull/1568

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,42 @@
+name: Deploy docs to Pages
+
+on:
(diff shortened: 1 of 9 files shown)


ehelms:
Looks great to me, is it possible to use https://github.com/theforeman/forklift/blob/master/docs/forklift.png as the favicon? or would we need a specifically formatted version of it?

ekohl:
Done.

ehelms (approved):

PR #1583, opened 2022-09-28
install vagrant and vagrant-libvirt from the copr

evgeni:


roles/vagrant/tasks/main.yml
```diff
@@ -1,15 +1,23 @@
...
   get_url:
-    url: https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo
-    dest: /etc/yum.repos.d/hashicorp.repo
+    url: https://copr.fedorainfracloud.org/coprs/pvalena/rubygems/repo/epel-{{ ansible_distribution_major_version }}/pvalena-rubygems-epel-{{ ansible_distribution_major_version }}.repo
+    dest: /etc/yum.repos.d/pvalena-rubygems.repo
+
+- name: 'install pvalena/vagrant copr'
+  get_url:
+    url: https://copr.fedorainfracloud.org/coprs/pvalena/vagrant/repo/epel-{{ ansible_distribution_major_version }}/pvalena-vagrant-epel-{{ ansible_distribution_major_version }}.repo
+    dest: /etc/yum.repos.d/pvalena-vagrant.repo
```
ekohl, line 10: Thoughts about using https://docs.ansible.com/ansible/latest/collections/community/general/copr_module.html?

roles/vagrant/tasks/main.yml
```diff
@@ -1,15 +1,23 @@
...
 
 - name: 'install vagrant'
   yum:
```
ekohl, line 13: Just a thought: should this be `package` instead of `yum`?

evgeni: given the rest won't work on non-EL, and there is no real difference between `yum` and `dnf`, "yes, but it's just cosmetics"

roles/vagrant/tasks/main.yml
evgeni: I prefer to stay away from `community.general`

ekohl: Fair enough.

ekohl (approved):

PR #1585, opened 2022-09-28
only become when needed during forklift setup

evgeni:


playbooks/setup_forklift.yml
```diff
@@ -34,6 +34,7 @@
         state: 'present'
 
     - name: 'install telemetry dependencies'
+      become: true
```
ekohl, line 37: I assume you install these system wide to speed things up on reuse between multiple users.

evgeni: that's correct

ekohl (approved):

PR #1588, opened 2022-09-28
Use the right flag to Ansible command

lfu:
Use the right flag to Ansible command to update the installed collections in requirements.yml.

https://issues.redhat.com/browse/SAT-11454

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,5 +296,5 @@ collections:
 Then run `ansible-galaxy` install:
 
 ```
-ansible-galaxy collection install -r requirements.yml --force
+ansible-galaxy collection install -r requirements.yml --force-with-deps
 ```


wbclark (commented):
Hey @lfu , what issue were you facing when you didn't use `--force-with-deps`?

In the past it was possible to install some of these ansible collections using an RPM. I'm not sure if that is still the case today, but if it is, we probably don't want to suggest `--force` or `--force-with-deps` and overwrite the user's RPM installed collections.

I usually install within a python venv as below:

```
python3 -m venv env
source env/bin/activate
pip install ansible apypie
ansible-galaxy collection install -r requirements.yml
```

(If I need to remind myself that the collection is installed locally this way and not system wide, I check the output  `which ansible-galaxy`)

What do you think of this approach @lfu ?

lfu:
@wbclark Thanks for reviewing! This change was made according to the instruction in the Jira card.
@parthaa @sjha4 Any suggestion?

sjha4:
@wbclark , @lfu : I remember this discussion from offline where this is needed on forklift that's already installed the collections in requirements.yml. To update the collections, the flag is required. It is fine IMO to add that flag in the Readme instructions.

wbclark:
It looks like requirements just has foreman infrastructure collection and community general. I am good with this, then. I can do a separate PR later to include alternate install instructions for a venv.

ekohl:
This is why you always specify the "why" in an issue, not the "how". Please modify the commit message to describe the reasoning.

lfu:
Updated.

ekohl (approved):
Thanks!

PR #1589, opened 2022-09-28
Update rpm packaging box to F36 and include semver + missing pkgs

chris1984:
* Update the box to use Fedora 36 since 33 is no longer present to be downloaded
* When using the `bump_rpm` script I noticed that it failed with missing `jq` and python `semver`
* Updated the playbook to include those 2 packages and use `python3-pip`
* Added `vim` and `wget` since those are handy

Python traceback is as follows:
```python
Traceback (most recent call last):
  File "/home/vagrant/foreman-packaging/update-requirements", line 8, in <module>
    import semver
ModuleNotFoundError: No module named 'semver'
```
Results of spinup:

```bash
rpm-packaging              : ok=6    changed=5    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

* Let me know if you want a better commit message, for the extra packages side
* Created this pr with the new setup, script works well with F36 https://github.com/theforeman/foreman-packaging/pull/8542

### DIFF
--- a/playbooks/rpm_packaging.yml
+++ b/playbooks/rpm_packaging.yml
@@ -18,17 +18,21 @@
           - ruby-devel
           - curl-devel
           - gem
-          - python3-click
           - python3-ruamel-yaml
-          - crudini
           - tito
           - mock
+          - jq
+          - python3-pip
+          - wget
+          - vim
+          - python3-semver
       become: true
 
     - name: Install obal
       pip:
         name: obal
         state: present
+        executable: pip3
       become: true
 
     - name: Clone foreman-packaging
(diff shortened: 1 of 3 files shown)


evgeni:
Any reason not to use the packaged `python3-semver` from Fedora?

playbooks/rpm_packaging.yml
```diff
@@ -23,12 +23,18 @@
           - crudini
```
ekohl, line 23: This is no longer needed so I opened https://github.com/theforeman/foreman-packaging/pull/8545.

Also note that since https://github.com/theforeman/foreman-packaging/commit/37558e583e65819adb125942a3683785922004f0 click (on line 21) is no longer needed.

evgeni (approved):
I'm cool with the changes as they are, but feel free to also incorporate ewouds feedback!

chris1984:
@ekohl removed `crudini` and `click` per your feedback

PR #1590, opened 2022-09-29
place bats and sos results in correctly named folders

evgeni:
before this change, these folders didn't include whether it's a instal or upgrade pipe. this wasn't a problem as we were running one pipe per box, but now we are running multiple and things break

### DIFF
--- a/playbooks/collect_debug.yml
+++ b/playbooks/collect_debug.yml
@@ -3,7 +3,7 @@
   become: true
   vars:
     bats_output_dir: '/root/bats_results'
-    remote_dir: "/tmp/debug-{{ pipeline_type | default('foreman') }}-{{ pipeline_version | default('nightly') }}-{{ pipeline_os | default('el7') }}"
+    remote_dir: "/tmp/debug-{{ pipeline_type | default('foreman') }}-{{ pipeline_version | default('nightly') }}-{{ pipeline_os | default('el7') }}-{{ pipeline_action | default('install') }}"
     smoker_output_dir: '/home/vagrant/smoker'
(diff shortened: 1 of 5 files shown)


playbooks/collect_debug.yml
```diff
@@ -3,7 +3,7 @@
   become: true
   vars:
     bats_output_dir: '/root/bats_results'
-    remote_dir: "/tmp/debug-{{ pipeline_type | default('foreman') }}-{{ pipeline_version | default('nightly') }}-{{ pipeline_os | default('el7') }}"
+    remote_dir: "/tmp/debug-{{ pipeline_type | default('foreman') }}-{{ pipeline_version | default('nightly') }}-{{ pipeline_os | default('el7') }}-{{ pipeline_action | default('install') }}"
```
ekohl, line 6: Thoughts about storing it in `/tmp/forklift-debug-$user/$pipeline`. Then you could also collect `/tmp/forklift-debug-*/**`, without knowing the exact pipeline.

evgeni: I did (not exactly `forklift-debug-$user` but similar) and opted for "let's not change too much right now"

ekohl:
This does require https://github.com/theforeman/jenkins-jobs/pull/266 as well, right?

evgeni:
> This does require [theforeman/jenkins-jobs#266](https://github.com/theforeman/jenkins-jobs/pull/266) as well, right?

Yes.

Well, doesn't *require*, but Jenkins would fail to fetch the bits without it.

ekohl (approved):

PR #1592, opened 2022-10-04
Always use graphics on Debian 11

ekohl:
Since 6f58dc045513d7091a4d1f0fa75c470208bd05c9 boxes in CI are provisioned without graphics. This was done because multiple users tried to allocate the same port for VNC. However, Debian 11 fails to boot up and fails with a timeout. This ignores the graphics_type option for Debian. This should be OK since Debian 11 only runs under a single user, so it can safely use ports for VNC. No other user will try to allocate the same ports.

Fixes: 6f58dc045513d7091a4d1f0fa75c470208bd05c9

### DIFF
--- a/vagrant/lib/forklift/box_distributor.rb
+++ b/vagrant/lib/forklift/box_distributor.rb
@@ -271,6 +271,8 @@ module Forklift
         add_disks(box, p)
 
         merged_options(box, 'libvirt_options').each do |opt, val|
+          # Debian 11 fails to boot without graphics
+          next if opt == 'graphics_type' && box['box_name'] == 'debian/bullseye64'
           p.instance_variable_set("@#{opt}", val)
         end
       end


evgeni:
The problem with graphics is, while each qemu process runs as a different user, all vnc ports are from the 5900+ port space, so they will very well still clash if both processes decide to use e.g. 5901.

evgeni:
So here is a smarter fix for it: https://github.com/theforeman/forklift/pull/1593 :)

ekohl:
> And we run two debian users: install and upgrade.

Oh yes, I forgot about that.

evgeni:
#1593 is merged and https://ci.theforeman.org/job/foreman-nightly-deb-pipeline/1043/ was green, so I am closing this as "not needed anymore"

PR #1596, opened 2022-10-25
explicitly clean yum cache after configuring yum repositories

evgeni:


roles/katello_repositories/tasks/main.yml
```diff
@@ -9,3 +9,11 @@
...
   command: subscription-manager repos --enable codeready-builder-for-rhel-8-x86_64-rpms
   when: ansible_distribution_major_version == "8" and ansible_distribution == "RedHat"
+
+- name: 'Clean yum'
+  command: 'yum clean all'
+  args:
+    warn: false
+  tags:
+    - packages
+  when: ansible_os_family == 'RedHat'
```
ekohl, line 19: I think DNF does this correctly so should we limit this to EL7?

evgeni: We have the same structure in other modules.

Certainly can, but I guess that cleanup is needed in more places then :)

evgeni:
FTR, this does fix the issue described in https://community.theforeman.org/t/katello-4-4-2-release-process/30636, even tho I am not sure why it's needed now and wasn't needed in the past.

ekohl (approved):

PR #1601, opened 2022-11-07
Use Puppet 7 for Katello devel setups

wbclark:
This is a short term workaround. The permanent fix is a new release of theforeman.operations


wbclark:
Amended commit message to clarify this is for katello_devel setups only

PR #1604, opened 2022-11-11
Don't overwrite katello-devel-answers.yaml

wbclark:
This allows re-running the provisioner without overwriting the user's changes to the configuration, so that the installer in devel setups more closely matches the familiar standard behavior.


roles/foreman_installer_devel_scenario/files/katello-devel.yaml
```diff
@@ -5,8 +5,6 @@
 :log_name: katello-devel.log
 :log_level: DEBUG
 :no_prefix: false
-# Forklift always overwrites the answers so let's not pretend we save them
```
ehelms, line 8: @wbclark is this no longer true?

wbclark: @ehelms with this change, yes the comment becomes no longer true, so we can really save the scenario answers now

`ansible.builtin.copy` module has a `force` parameter which defaults to `yes` -- see https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html#parameter-force

This PR explicitly sets `force: no` so that the devel installer role creates the devel scenario answers file if it doesn't already exist, but it will no longer overwrite the file if it has changes.

My assumption here is that if the developer _wants_ to reset to the initial state, a better way is to just destroy and recreate the box.

So in summary, the commit introduces two changes:

1. Don't overwrite the devel scenario answers file if we re-run the playbook

2. Do save config changes in the devel scenario answers file, so the installer can be used to make persistent changes (the same way we already use it in all the non-devel scenarios)

roles/foreman_installer_devel_scenario/tasks/main.yml
```diff
@@ -20,6 +20,7 @@
   template:
     src: "{{ role_path }}/templates/{{ item }}"
     dest: /etc/foreman-installer/scenarios.d/{{ item }}
+    force: no
```
wbclark, line 23: @ehelms thinking about it more, I think I should also re-name this task to something like `Create initial devel scenario answers file` so that it's clear we won't overwrite the file once it exists

wbclark: I pushed an update which also renames this task so that the new behavior is more clear

wbclark:
Test failure unrelated but https://github.com/theforeman/forklift/pull/1605 should fix that

wbclark:
Tests are now :green_apple:

roles/foreman_installer_devel_scenario/tasks/main.yml
```diff
@@ -16,10 +16,11 @@
   with_items:
     - katello-devel.yaml
 
-- name: 'Copy answers file'
+- name: "Create devel scenario answers file if it doesn't already exist"
```
ehelms, line 19: This will result in some options to this role only affecting the first run of the role and then never again -- that is generally an unexpected situation for Ansible roles.

If we are going this route, I should think we need to go all the way and make it such that the dynamic values we do have (https://github.com/theforeman/forklift/blob/master/roles/foreman_installer_devel_scenario/templates/katello-devel-answers.yaml#L3-L8) get passed to the installer as parameters.

wbclark: Thanks, I pushed an additional commit which moves these dynamic values out of the answers file and into `foreman_installer_options_internal_use_only`, which is probably where they should have been all along. (What clown put dynamic values in the answers file? Oh right, it's me. I'm the clown.)

So, the behavior should now be consistent with what you expect from an ansible role. Updating the `foreman_installer_devel_scenario_user` or `foreman_installer_devel_scenario_group` will again properly result in the installer making the appropriate changes.

I also removed `generate` and `deploy` from the certs params in the answers file, because the values we were passing were already the defaults for that module, so I couldn't see any reason why those needed to be there. I can revert that if it's needed for any reason.

I've tested this in my own setup, which does include a non-default value for `foreman_installer_devel_scenario_user` and it's all working just fine for me.

roles/foreman_installer_devel_scenario/tasks/main.yml
```diff
@@ -16,10 +16,11 @@
...
   template:
     src: "{{ role_path }}/templates/{{ item }}"
     dest: /etc/foreman-installer/scenarios.d/{{ item }}
+    force: no
```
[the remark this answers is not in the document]
ehelms: I can see that being useful.

ehelms (approved):
Feel free to do the force as a follow up or in this PR.

PR #1605, opened 2022-11-15
Don't test Ruby 2.0

wbclark:
Ruby 2.0 is very old at this point and tests are failing with `Gem::InstallError: rake requires Ruby version >= 2.2.`

Even Ruby 2.6 is EOL now so we could remove this as well (but it's not currently causing test failures)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,7 +50,6 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - '2.0'
           - '2.6'
           - '2.7'
           - '3.0'


ehelms (approved):

evgeni:
And yet, our CI is still running on EL7 which has Ruby 2.0 as the default ruby, and that's why we test it.

ekohl:
At least in upstream we should no longer rely on EL7. Perhaps there's something downstream?

.github/workflows/main.yml
```diff
@@ -50,7 +50,6 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - '2.0'
```
ekohl, line 53: Shouldn't this have been replaced with Ruby 2.5 since that's what EL8 uses by default?

PR #1606, opened 2022-11-17
Set Ruby 2.5 as the minimum version

ekohl:
Follow up on https://github.com/theforeman/forklift/pull/1605. Ruby 2.5 was chosen because EL8 ships Ruby 2.5 by default.


vagrant/Rakefile
```diff
@@ -12,9 +12,11 @@ default_tasks = %i[test]
 
 begin
   require 'rubocop/rake_task'
-  RuboCop::RakeTask.new
-  # insert 0, because Ruby 2.0 doesn't know Array.prepend
-  default_tasks.insert(0, 'rubocop')
+  rubocop_task = RuboCop::RakeTask.new
+  if ENV['GITHUB_ACTIONS'] == 'true'
+    rubocop_task.formatters << 'github'
+  end
```
ekohl, line 18: This doesn't appear to work. Moving back to draft.

ekohl:
No idea why the annotations aren't working, but this is otherwise working as intended.

ekohl:
I'm updating this, but now that I see it I saw a note about EL8 having Ruby 2.5 by default. I wonder if that's a concern for our CI.

PR #1608, opened 2022-11-18
use pip3.9 now that containers were updated to el8.7

evgeni:
this contains ansible-core 2.13, which uses python 3.9


evgeni:
:see_no_evil: unbreaking CI on a friday morning

PR #1610, opened 2022-12-05
enable foreman-maintain backup test on Debian for 3.4+

evgeni:


ekohl (approved):
This also means we can enable backup documentation on docs.theforeman.org. I've hidden it there on Debian since it wasn't available for a long time.

evgeni:
https://github.com/theforeman/foreman-documentation/pull/1864

PR #1611, opened 2022-12-05
verify-packages: drop pre Foreman 2.5 support

evgeni:
2.5+ is Ruby 2.7 only on EL7, so we can simplify the test quite a lot

### DIFF
--- a/bats/fb-verify-packages.bats
+++ b/bats/fb-verify-packages.bats
@@ -6,50 +6,12 @@ set -o pipefail
 load os_helper
 load foreman_helper
 
-@test "verify no rh-ruby24, tfm-ror51 or tfm-ror52 SCL packages" {
-  tSetOSVersion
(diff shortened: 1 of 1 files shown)


bats/fb-verify-packages.bats
```diff
@@ -6,50 +6,12 @@ set -o pipefail
 load os_helper
 load foreman_helper
 
-@test "verify no rh-ruby24, tfm-ror51 or tfm-ror52 SCL packages" {
-  tSetOSVersion
```
evgeni, line 10: this is not needed as it's already done in `tIsRHEL`:
https://github.com/theforeman/forklift/blob/fd43a0ad473e618151cb6aabc6e3f5b88406fff8/bats/os_helper.bash#L55-L56

evgeni: more cleanup in https://github.com/theforeman/forklift/pull/1612

ekohl (approved):

PR #1612, opened 2022-12-05
call tSetOSVersion where its actually needed

evgeni:


bats/os_helper.bash
```diff
@@ -196,6 +196,7 @@ tDirectoryExists() {
 
 tRHSubscribeAttach() {
   if tIsRHEL; then
+    tSetOSVersion
```
ekohl, line 199: I guess this is here because of `$OS_VERSION` on line 205. Still, it's odd to see `tIsRHEL` and then set the OS version.

Having said that: is there even `rhel-server-rhscl-8-rpms` or could we hardcode it to 7? 

And where is this even used? 9adcdae1abc25ccdd56caae051fbb3a4280e0d03 was just syncing the whole file from foreman-bats but I don't see it used (nor `tRHEnableEPEL` below for that matter). Should we just drop it or si this used downstream?

evgeni: I thought about dropping that code too. It is not used anywhere, no.

ekohl: Want me to submit a PR or are you working on that?

evgeni: updated

ekohl (approved):

PR #1618, opened 2022-12-08
move "check service status" to foreman

evgeni:


ekohl (approved):

evgeni:
holding off merging -- this will break downstream tests which need a round of refactoring before the deletion here will not explode :see_no_evil:

evgeni:
refactoring happened

PR #1620, opened 2022-12-08
update centos stream boxes

evgeni:
now also uses a direct cloud.centos.org link for 8-stream, as it seems noone can be bothered to update the version on
https://app.vagrantup.com/centos


evgeni:
this speeds up the pipelines by ~2:30 min (before it took 3:50 to run "yum update", now it's about 1:15)

ehelms (approved):

PR #1622, opened 2022-12-09
assert export_path and export_history_id aren't empty

evgeni:


evgeni:
FTR, something is currently broken in nightly, which results in a few things not working, but then this test just goes and does a `cp -r /* $import_path` and that fills up the disk, resulting in no results reported or anything.

ekohl (approved):
Perhaps `[ -n "${latest_export}" ]` would also be a good thing.

bats/fb-katello-content.bats
```diff
@@ -181,6 +181,11 @@ setup() {
   # /var/lib/pulp/exports/Test_Organization/Test_CV/1.0/2020-12-11T16-04-08-00-00
   import_path="/var/lib/pulp/imports/bats-test-$export_history_id"
 
+  # assert latest_export, export_path and export_history_id aren't empty
+  [ -n "${latest_export}" ]
```
ekohl, line 185: Should this happen before we call `cut` on it?

Also, on line 178 In think the comment is outdated since we pass `--fields`, so we should only get the two.

evgeni: I will update the details later, right now I just want the test to cleanly fail again

PR #1624, opened 2022-12-12
only remove sub-man on non-RHEL as otherwise we can't reinstall it

evgeni:
this "fixes" the concerns from [this comment](https://github.com/theforeman/forklift/pull/1391/files#r694969678)

### DIFF
--- a/bats/os_helper.bash
+++ b/bats/os_helper.bash
@@ -57,6 +57,10 @@ tIsEL() {
   fi
 }
 
+tIsRHEL() {
+  tIsEL "$1" && grep -q 'Red Hat Enterprise Linux' /etc/redhat-release
+}
+
 tIsDebian() {
   tIsDebianCompatible && ! tIsUbuntuCompatible
 }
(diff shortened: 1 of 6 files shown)


evgeni:
this fixed the issues I had when running GR tests on RHEL hosts

bats/os_helper.bash
```diff
@@ -57,6 +57,10 @@ tIsRHEL() {
   fi
 }
 
+tIsRHEL() {
+  tIsEL "$1" && grep -q 'Red Hat Enterprise Linux' /etc/redhat-release
```
ekohl, line 61: `ID` in `/etc/os-release` is more reliable because of false positives on Oracle Linux. On an Oracle Linux 8 container:
```console
# cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.7 (Ootpa) 
```

ekohl (approved):
Not sure how much we care about Oracle Linux (probably not), but I did want to mention it.

evgeni:
Yeah, I don't care about OEL.

PR #1625, opened 2022-12-12
force execution of REX tests in an own organization

evgeni:
This ensures any *other* proxy with the REX capability is not used, as that can lead to wrong results.


ekohl (approved):
Is this sufficient? Don't you need to set the org on the Smart Proxy too? If it works, I'm happy with this PR.

PR #1628, opened 2022-12-16
Use the FQDN for the REX Smart Proxy name

ekohl:
On Debian the hostname is usually the short name, not the FQDN like on Red Hat based systems. In Foreman the Smart Proxy is registered using the FQDN so no Smart Proxy is found.

This was found via https://community.theforeman.org/t/foreman-plugins-3-5-deb-test-pipeline-3-failed/31619 (https://ci.theforeman.org/job/foreman-plugins-3.5-deb-test-pipeline/3/) and we likely didn't find it because https://ci.theforeman.org/job/foreman-plugins-nightly-deb-test-pipeline/ hadn't ran since March. I just kicked that off, but I expect it to fail in the same way as 3.5.

Fixes: 13630ab2476fda158a55f0d395177e63c2c42306

### DIFF
--- a/bats/fb-test-foreman-rex.bats
+++ b/bats/fb-test-foreman-rex.bats
@@ -14,14 +14,16 @@ load foreman_helper
     job_template='Run Command - Script Default'
   fi
 
+  SMART_PROXY_NAME=$(hostname -f)
+
   # Force execution of the test in an own Organization
   # This ensures any *other* proxy with the REX capability is not used,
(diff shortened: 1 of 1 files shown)


evgeni (approved):

evgeni:
Technically, updating the job invocation is not necessary, as *hosts* can be found by their short name (as seen in the ansible test)

But it also doesn't hurt

PR #1633, opened 2022-12-22
call `hammer ansible roles fetch` without checking output

evgeni:
during upgrade tests, it will already be imported, and then `fetch` doesn't list it anymore

### DIFF
--- a/bats/fb-test-foreman-ansible.bats
+++ b/bats/fb-test-foreman-ansible.bats
@@ -19,13 +19,13 @@ setup() {
   echo "- command: uptime" > "/etc/ansible/roles/${ROLE_NAME}/tasks/main.yml"
 
   # check if smart_proxy_ansible finds it and can import it
-  hammer ansible roles fetch --proxy-id 1 | grep "${ROLE_NAME}"
+  hammer ansible roles fetch --proxy-id 1
   hammer ansible roles sync --proxy-id 1 --role-names "${ROLE_NAME}"
 
   # sync creates a task, so let's wait for it to complete
   tWaitForTask SyncRolesAndVariables
 
-  hammer ansible roles list | grep "${ROLE_NAME}"
+  hammer --output csv --no-headers ansible roles list --search="name=${ROLE_NAME}" | grep "${ROLE_NAME}"
 }
 
 @test "run a role" {


bats/fb-test-foreman-ansible.bats
```diff
@@ -19,13 +19,15 @@ setup() {
   echo "- command: uptime" > "/etc/ansible/roles/${ROLE_NAME}/tasks/main.yml"
 
   # check if smart_proxy_ansible finds it and can import it
-  hammer ansible roles fetch --proxy-id 1 | grep "${ROLE_NAME}"
+  if ! hammer --output csv --no-headers ansible roles list --search="name=${ROLE_NAME}" | grep "${ROLE_NAME}"; then
+    hammer ansible roles fetch --proxy-id 1 | grep "${ROLE_NAME}"
+  fi
```
ekohl, line 24: Wouldn't it be easier to just unconditionally fetch without a grep? That way we test that we can always fetch. It just doesn't have to result in importing anything.

bats/fb-test-foreman-ansible.bats
```diff
@@ -19,13 +19,15 @@ setup() {
...
   hammer ansible roles sync --proxy-id 1 --role-names "${ROLE_NAME}"
 
   # sync creates a task, so let's wait for it to complete
   tWaitForTask SyncRolesAndVariables
 
-  hammer ansible roles list | grep "${ROLE_NAME}"
+  hammer --output csv --no-headers ansible roles list --search="name=${ROLE_NAME}" | grep "${ROLE_NAME}"
```
ekohl, line 30: Not for this PR, but this feels like one of those "our bats is too old so we can't use `assert_output` cases". Should we modify https://github.com/theforeman/forklift/pull/1634 to also pull in `bats-assert`?

bats/fb-test-foreman-ansible.bats
evgeni: That would also work, yeah

bats/fb-test-foreman-ansible.bats
evgeni: You mean https://github.com/bats-core/bats-assert? Yeah, certainly something we could try pulling in.

ekohl (approved):

PR #1634, opened 2022-12-22
always install bats from git

evgeni:
we don't setup EPEL on our nodes anymore, so there is nowhere to get the packaged version. also, switch to the maintained git repo.


ekohl (approved):
I checked and https://src.fedoraproject.org/rpms/bats is just `bats-core` and there is no package for `bats-assert`. That means this is a straight 1-to-1 conversion and we're not missing anything.

Having said that: we should look at what additional parts of bats we would want.

Issue #1641, opened 2023-01-23
Documented requirements are out of date


agriffit79:
The readme specifies Ansible 2.9+ but that seems to no longer be sufficient due to multiple errors in the playbooks.

2.13.7 worked for me, but a lower version may also work?

evgeni:
I would tend to say it should still work with 2.9, what errors did you get?

evgeni:
Would you have the time to try with a "less-ancient" 2.9 version?

agriffit79:
Was just doing that... It works with 2.9.20. Not sure if that warrants an update to the readme? It cost me a whole morning trying to debug it.

evgeni:
Yeah, that sounds like a valuable update!

We started using collections at some point, and there were still bugs in that regard in *early* versions if 2.9, so it might very well be what you've been seeing

PR #1642, opened 2023-01-23
Update version references in README.md

wbclark:
Older Ansible versions are now EOL, while 2.12 is still receiving security updates until May 2023 at least. Also, the devel and nightly setups haven't used centos7 for a while now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Forklift provides tools to create Foreman/Katello environments for development, 
 ### Requirements
 
 * Vagrant - 2.2+ - Both the VirtualBox and Libvirt providers are tested
-* Ansible - 2.9+
+* Ansible - 2.12+
 * [Vagrant Libvirt provider plugin](https://github.com/vagrant-libvirt/vagrant-libvirt) (if using Libvirt)
 * Virtualization enabled in BIOS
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,7 @@ Forklift provides tools to create Foreman/Katello environments for development,
 ### Requirements
 
 * Vagrant - 2.2+ - Both the VirtualBox and Libvirt providers are tested
-* Ansible - 2.9+
+* Ansible - 2.12+
```
ekohl, line 29: In https://github.com/theforeman/forklift/issues/1641 it's suggested recent 2.9 should still work.

wbclark: Thanks, I missed the GH issue and got here via https://community.theforeman.org/t/problems-using-forklift/32003

Even if later 2.9+ still works, perhaps it is not ideal to advertise or recommend this fact considering there are multiple EOL versions between 2.9+ and the current maintenance branch.

Using Ansible with Forklift feels pretty different from most other use cases, and we have our own ansible.cfg in version control while Ansible still doesn't have support for merging multiple ansible.cfg files (see https://github.com/ansible/ansible/issues/79506 which looks written by a familiar face fwiw)

Our ansible.cfg also makes `./playbooks/galaxy_collections` the first option for `collections_path` so we're already avoiding installing them system wide when the user follows the current instructions in the README.

So, IMO, we could just add instructions to the README.md which say that, for 99% of users who are running forklift on their workstation and not part of a CI pipeline, it's very easy and reliable to make and activate a python virtual environment, `pip install ansible-galaxy`, `ansible-galaxy collection install -r requirements.yml`, and they are all set.

This entirely avoids a frequent pain point I've seen when trying to help people get up and running with Forklift.

wbclark:
Closing in favor of #1643 , thanks for the feedback on both

PR #1643, opened 2023-01-24
Improve Quickstart Guide

wbclark:
This commit is aimed at improving the organization and content of the readme, with an aim towards helping users quickly learn just enough about Forklift to move on to whatever it is that they actually care about.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,26 +7,37 @@ Forklift provides tools to create Foreman/Katello environments for development, 
  * [Using Forklift](#using-forklift)
    - [Requirements](#requirements)
    - [Quickstart](#quickstart)
+   - [Foreman Nightly Box](#foreman-nightly-box)
+   - [Katello Nightly Box](#katello-nightly-box)
+   - [Additional Documentation](#additional-documentation)
+   - [Development Environment](#development-environment)
+   - [Using Playbooks and Roles Without Vagrant](#using-playbooks-and-roles-without-vagrant)
+   - [Credentials](#credentials)
    - [Poor man's DNS a.k.a /etc/hosts](#poor-mans-dns-aka-etchosts)
    - [Adding Custom Boxes](#adding-custom-boxes)
+   - [Using SSHFS to share folders](#using-sshfs-to-share-folders)
+   - [Using NFS to share folders](#using-nfs-to-share-folders)
    - [Customize Deployment Settings](#customize-deployment-settings)
+   - [Customize Available Boxes](#customize-available-boxes)
    - [Post Install Playbooks](#post-install-playbooks)
    - [Using Local Ansible Collection](#using-local-ansible-collection)
- * [Production Environments](docs/production.md)
- * [Development Environments](docs/development.md)
- * [Stable Boxes](docs/stable_boxes.md)
- * [Testing Environments](docs/testing.md)
- * [Provisioning environment](docs/provision.md)
- * [Plugins](docs/plugins.md)
- * [Using Forklift as a Library](docs/library.md)
- * [Troubleshooting](docs/troubleshooting.md)
+ * [Production Environments](https://theforeman.github.io/forklift/production/)
+ * [Development Environments](https://theforeman.github.io/forklift/development/)
+ * [Stable Boxes](https://theforeman.github.io/forklift/stable_boxes/)
+ * [Testing Environments](https://theforeman.github.io/forklift/testing/)
+ * [Provisioning environment](https://theforeman.github.io/forklift/provision/)
+ * [Plugins](https://theforeman.github.io/forklift/plugins/)
+ * [Using Forklift as a Library](https://theforeman.github.io/forklift/library/)
+ * [Troubleshooting](https://theforeman.github.io/forklift/troubleshooting/)
+ * [Vagrant](https://theforeman.github.io/forklift/vagrant/)
+ * [Packer](https://theforeman.github.io/forklift/packer/)
 
 ## Using Forklift
 
 ### Requirements
 
 * Vagrant - 2.2+ - Both the VirtualBox and Libvirt providers are tested
-* Ansible - 2.9+
+* Ansible - 2.12+
 * [Vagrant Libvirt provider plugin](https://github.com/vagrant-libvirt/vagrant-libvirt) (if using Libvirt)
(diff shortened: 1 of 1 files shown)


ekohl (commented):
We do have https://theforeman.github.io/forklift/ which is published from this repo using mkdocs. Perhaps something we should leverage more?

wbclark:
I think that's a fine idea, but I'm not sure it's something I can take on right now. And I think a README / quickstart is still a good idea regardless, because it's usually the very first thing I look at when browsing any repo on GitHub.

Basically, this PR is based on repeat pain points that I've seen when suggesting Forklift to colleagues. The recent community support thread asking for guidance here got me thinking about this topic again.

Since the Katello team is currently onboarding a new developer who will need to start using this tool soonish, and because we'll have an intern joining us at some point in the near future as well, I went ahead and wrote out some improvements that should make this whole process a little bit more user friendly.

wbclark:
Added syntax highlighting to many code blocks where it was missing

README.md
```diff
@@ -26,34 +35,94 @@ Forklift provides tools to create Foreman/Katello environments for development,
...
+
+```sh
 ansible-galaxy collection install -r requirements.yml
-vagrant up centos7-foreman-nightly
 ```
 
-The same can be quickly done for a development environment where GITHUB_NICK is your GitHub username:
+When this is complete, simply follow the next section of this README to try out the latest nightly Foreman build using Forklift.
+
+### Foreman Nightly Box
```
ekohl, line 65: This should mostly be covered in https://github.com/theforeman/forklift/blob/master/docs/development.md already. https://theforeman.github.io/forklift/development/ should be the rendered version of that.

README.md
```diff
@@ -71,8 +140,6 @@ In an example above, ansible was instructed to use specific private key (overrid
 
 Other playbooks from playbooks/ directory can be used similarly, though some might need more variables and investigating their parameters is recommended first.
 
-More thorough guides can be found in the [docs folder](https://github.com/theforeman/forklift/tree/master/docs).
```
ekohl, line 74: Perhaps we should change this to https://theforeman.github.io/forklift/ instead and give it a better landing page.

ekohl (changes requested):
I appreciate it may be easier to add it to README, but we have documented this already. If it's not found (which is very likely), then that's the problem we should address. If there are problems in it, those should be fixed.

wbclark:
Updated to use links to https://theforeman.github.io/forklift/ throughout the README

README.md
wbclark: Changed docs/index.md to link to README.md -- how does that work for you?

ekohl: Still not a good idea, because we should encourage people to see all the documentation. In my mind that means less README. Basically none IMHO. Because if there are some instructions, people may think it's all the documentation and not look further.

I'd expect:
* A quickstart: like 5 lines of code to get started
* Required versions
* Where to find the full documentation.

wbclark: The full documentation contains a lot of information that is only going to be relevant to a smaller cross section of users, so compelling people to refer to the full documentation in the beginning poses an additional burden of sifting through all that information and determining what is relevant. This also means sifting through multiple documents, which shouldn't be required for getting started with basic usage.

On the other hand, the quickstart can provide both 1. enough information that it's realistically possible to get started using the tool, and 2. clear and prominent indication that complete documentation exists, and where to find it once it becomes necessary.

This approach is in line with the best practices outlined [here](https://swimm.io/blog/get-started-with-technical-documentation-writing-best-practices/), in particular numbers 2, 3, and 5.

I think there is information that can certainly be pruned from the README, but I'm hesitant to remove from the sections on the basic usage of the development and production boxes, which are the most important IMO. It's not super useful to know how to `vagrant up` if you don't actually know how to start the development server or access the WebUI afterwards.

What do you think?

ekohl: I don't mind having one long document, but rather than symlinking I'd suggest `mv README.md docs/index.md` and then write a minimal new README that redirects to the docs site.

wbclark: Is your primary concern to make it so that the docs site is highly discoverable to those looking at this repo in their browser?

If so, I agree that's important. Can it be adequately addressed by making the link to the docs site be the most prominent item at the very beginning of README.md, without stripping the rest of its content?

For a terminal user, locating the quickstart guide only in docs/index.md means that it's less obvious to start here, vs. the top level README.md, while a README.md which contains only redirects to the docs site is not as useful in the terminal.

From my PoV then, adding the link the docs site prominently at the top of README.md and creating docs/index.md as a symlink then combines the best features of both approaches.

That said, I don't really need to die on this hill, and I'd prefer to try merging this sooner rather than later (as I said, we're onboarding a new developer at present, and another in the near term) so if I've failed to convince you just let me know and I'll make the change.

README.md
```diff
@@ -7,53 +7,126 @@ Forklift provides tools to create Foreman/Katello environments for development,
...
+
+By default, `forklift` deploys Foreman with `admin`/`changeme` as username and password.
+
+### Katello Nightly Box
+
+Katello nightly boxes are available as well; simply change `centos8-foreman-nightly` to `centos8-katello-nightly` and the steps are otherwise exactly the same as above.
+
+### Additional Documentation
+
+The remainder of this README contains helpful notes on additional topics that are likely to be of interest to users. For the most thorough and complete guides however, please refer to the [docs page](https://theforeman.github.io/forklift).
```
wbclark, line 93: @ekohl it occurred to me that you were responding to the deleted line below, and it would have been easy to miss that I didn't actually remove it, but rather give it a separate heading and moved it closer to the top.

That said, I could move this further to the very top.

ehelms:
Nit: Based on your initial comment, a commit message that reflects the intent -- to improve the quick start for users might be more useful

ianballou (commented):
I think this is a good improvement to the README! I agree it would be nice to move this big chunk of text out of the README and into another docs file -- maybe in a second PR? I already see a nice improvement here.

wbclark:
Good morning, any update on this? Is it mergeable as is, or with minor changes, or should it be closed?

ekohl (approved):
A bit slower response since conferences took all my available time. I still think we should drive more people to the docs rather than relying on README, but for now I'll merge this and we can refactor it later.

PR #1647, opened 2023-01-26
remove checks for old foreman/katello versions

evgeni:


evgeni:
cc @pcreech @Odilhao -- would you mind checking if that affects you downstream in any way? I tried to keep things in line for 6.10+, but might have missed something.

ehelms (approved):

PR #1649, opened 2023-01-30
Refs #35985 - Reflect Puppet installer cert parameter removal

ekohl:


ehelms (approved):
I did not test, but the logic all looks right.

PR #1652, opened 2023-02-15
don't enable azure cli on Debian

evgeni:
it's not packaged there


ekohl (approved):
Looks like it's not in develop either so :+1:

PR #1653, opened 2023-02-16
don't ignore smoker errors for nightly

evgeni:


ekohl (approved):
It's finally happening. Thanks for taking it over the finish line.

PR #1654, opened 2023-02-17
Update stable devel/nightly to use centos8-stream

chris1984:


packer/centos8-katello-devel-stable.json
```diff
@@ -63,5 +64,4 @@
       "type": "ansible"
     }
   ]
-}
-
+}
```
ekohl, line 67: A line ending would be nice (true for all files)

packer/http/ks.cfg
```diff
@@ -1,46 +1,60 @@
...
-clearpart --all --initlabel
-autopart
-auth --enableshadow --passalgo=sha512 --kickstart
+# Root password
+rootpw vagrant
+# Run the Setup Agent on first boot
 firstboot --disabled
 eula --agreed
+# Do not configure the X Window System
+text
```
ekohl, line 22: I recently learned you want `cmdline` for true unattended installations. See https://pykickstart.readthedocs.io/en/latest/commands.html

packer/http/ks.cfg
```diff
@@ -1,46 +1,60 @@
...
-yum install -y sudo
+%post
+exec < /dev/tty3 > /dev/tty3
+chvt 3
+echo
+echo "################################"
+echo "# Running Post Configuration   #"
+echo "################################"
+(
+/usr/bin/yum -y install drpm sudo
```
ekohl, line 52: You should these in `%packages`. And personally I dislike `drpm` because it's much slower if you have a fast network.

packer/http/ks.cfg
```diff
@@ -1,46 +1,60 @@
...
-bootloader --location=mbr
-text
-skipx
-zerombr
-clearpart --all --initlabel
-autopart
-auth --enableshadow --passalgo=sha512 --kickstart
+# Root password
+rootpw vagrant
+# Run the Setup Agent on first boot
```
ekohl, line 18: The comment suggest you actually don't run it.

packer/http/ks.cfg
```diff
@@ -1,46 +1,60 @@
...
+%post
+exec < /dev/tty3 > /dev/tty3
+chvt 3
+echo
+echo "################################"
+echo "# Running Post Configuration   #"
+echo "################################"
+(
+/usr/bin/yum -y install drpm sudo
+useradd -m -p $(openssl passwd vagrant) vagrant
```
ekohl, line 53: It'd be better to use the `user` command: https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#id64

packer/http/ks.cfg
```diff
@@ -1,46 +1,60 @@
...
-%packages --ignoremissing --excludedocs
-@Base
-@Core
-openssh-clients
-sudo
-openssl-devel
-readline-devel
-vim
-wget
+%packages --ignoremissing
```
ekohl, line 30: Why did you drop `--excludedocs`? Is that gone in EL8? And I'd drop `--ignoremissing`.

packer/http/ks.cfg
```diff
@@ -1,46 +1,60 @@
...
+timezone UTC
 reboot
-user --name=vagrant --plaintext --password vagrant --groups=vagrant,wheel
 
-%packages --ignoremissing --excludedocs
-@Base
-@Core
-openssh-clients
-sudo
-openssl-devel
```
ekohl, line 30: Isn't the devel package used in `bundle install`?

ekohl (changes requested):

packer/centos8-katello-nightly-stable.json
```diff
@@ -27,7 +26,7 @@
                 [ "-smp", "4" ]
             ],
             "boot_command": [
-              "<tab> text ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/ks.cfg<enter><wait>"
+              "<tab> text ks=http://{{ .HTTPIP }}:{{ .HTTPPort }}/ks8.cfg<enter><wait>"
```
ekohl, line 29: Where does `ks8.cfg` come from? I don't see the file being renamed.

chris1984: That was part of my testing, when we were doing both el 7/8, will fix that up and this pr shortly. Now we only need ks.cfg since we are just dealing with el8 only

PR #1655, opened 2023-02-21
use pulpcore 3.22 for nightly

evgeni:


evgeni:
merging w/o real review so that @ianballou has something to debug when he comes online :smiling_imp:

PR #1658, opened 2023-02-23
use the FQDN as hostname in Ansible tests

evgeni:


evgeni:
this still doesn't make ansible tests pass on debian, but at least it's a start!

ekohl (approved):

PR #1660, opened 2023-03-02
add a role to use packit PRs

evgeni:


docs/development.md
```diff
@@ -206,6 +206,18 @@ The Koji role and task ID variable can be added to download and configure a repo
     - katello
 ```
 
+## Packit PR Builds
+
+Forklift supports using Packit PR builds for testing purposes. To use those one need to define the `packit_prs` variable to the list of PRs (in the form of `<org>/<repo>/<pr>`):
```
ehelms, line 211:
```suggestion
Forklift supports using Packit PR builds for testing purposes. To use those one needs to define the `packit_prs` variable to the list of PRs (in the form of `<org>/<repo>/<pr>`):
```

docs/development.md
```diff
@@ -206,6 +206,18 @@ The Koji role and task ID variable can be added to download and configure a repo
...
+
+In `99-local.yaml`:
+```yaml
+ansible:
+  variables:
+    packit_prs:
+      - theforeman/foreman_maintain/690
```
ehelms, line 218: What about supporting instead (or alongside) the actual PR URL? That will make it super easy to just copy/paste, e.g.

`theforeman/foreman_maintain/pull/689`

Bonus points for also supporting `https://github.com/theforeman/foreman_maintain/pull/689` as that may be a common thing developers instinctively also copy and paste.

evgeni: I've stolen the format from the PR support in foreman installer. I agree making it accept more inputs would be cool

ehelms: Yes, and I've always thought -- why did I implement it that way? I should have used the format from the URL itself. This was my plea to fix my mistake out of the gate with packit PRs :smile:

evgeni: I guess `item.lower().replace('https://github.com/','').replace('/pull/','/').split('/')` would work?

evgeni: This was intended as a half-joke, but it actually works.

Should I update the installer thing to also use that "magic"?

ehelms (approved):

PR #1663, opened 2023-03-20
Add task to fix .env files in foreman

poelzi:
When using a stable box, changing the hostname results in a broken installation. Foreman properly detects the hostname, but webpack is started with the hostname of the installation which is written to ~/foreman/.env Ensure the --public argument reflects the current hostname


ekohl (approved):
Thanks!

PR #1664, opened 2023-03-27
Add a test for disabling puppet

ehelms:


ehelms:
This currently fails, but is fixed by: https://github.com/theforeman/foreman_maintain/pull/702

ehelms:
Rebased this after a bit, this functionality still exists so I think it's worth having the test.

PR #1668, opened 2023-04-13
use the satellite-clone CLI, not playbook directly

evgeni:


evgeni:
this is what I used to test https://github.com/RedHatSatellite/satellite-clone/pull/432

ekohl (approved):

PR #1669, opened 2023-04-13
add file list for clone with online backup

evgeni:


evgeni:
Given satellite-clone doesn't support vanilla Foreman: no

`offline_no_pulp` is "katello without pulp data" :)

ehelms (approved):

PR #1671, opened 2023-04-26
Adding test for puppet report

Griffin-Sullivan:
I added a short test section to the puppet bats test for finding config-reports for Puppet and then checking the contents of one of the reports. I wasn't quite sure what we should expect to always find in the report, so for now I left it looking for `Resource: Puppet`.

### DIFF
--- a/bats/fb-test-puppet.bats
+++ b/bats/fb-test-puppet.bats
@@ -6,6 +6,11 @@ set -o pipefail
 load os_helper
 load foreman_helper
 
+setup() {
+  HOSTNAME="$(hostname -f)"
+  export HOSTNAME
+}
+
 if [[ -e /etc/profile.d/puppet-agent.sh ]] ; then
   . /etc/profile.d/puppet-agent.sh
 fi
@@ -92,7 +97,7 @@ fi
   [ "${count}" -gt 1 ]
 }
 
-@test "Assign puppet-environment to default taxonomies" {
+@test "assign puppet-environment to default taxonomies" {
   hammer puppet-environment update --name=production --locations "Default Location" --organizations "Default Organization"
 }
 
@@ -104,4 +109,9 @@ fi
 @test "apply class with puppet agent" {
   puppet agent -v -o --no-daemonize
   grep -i "property of the Foreman project" /etc/motd
+}
+
+@test "ensure there is a report for Puppet" {
+  REPORT_ID=$(hammer --output csv --no-headers config-report list --fields "Id" --search "host=${HOSTNAME} origin=Puppet")
+  hammer config-report info --id "${REPORT_ID%%[[:space:]]*}" | grep "Resource: Puppet"
 }


ehelms:
@ekohl This looks sane to me, but I'd want to you ensure the assertion makes sense.

bats/fb-test-puppet.bats
```diff
@@ -105,3 +105,9 @@
   puppet agent -v -o --no-daemonize
   grep -i "property of the Foreman project" /etc/motd
 }
+
+@test "ensure there is a report for Puppet" {
+  RESULT=$(hammer --output csv --no-headers config-report list --fields "Id" --search "host=${HOSTNAME} origin=Puppet")
+  REPORT_ID=$(echo "${RESULT%%[[:space:]]*}")
```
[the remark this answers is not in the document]
Griffin-Sullivan: whoops I left the echo in after changing this

bats/fb-test-puppet.bats
```diff
@@ -105,3 +105,9 @@ fi
   puppet agent -v -o --no-daemonize
   grep -i "property of the Foreman project" /etc/motd
 }
+
+@test "ensure there is a report for Puppet" {
+  RESULT=$(hammer --output csv --no-headers config-report list --fields "Id" --search "host=${HOSTNAME} origin=Puppet")
```
evgeni, line 110: depending on the OS settings, `HOSTNAME` can be short or long, so to enforce the long notation, we have the following in the Ansible test `setup()` function:
https://github.com/theforeman/forklift/blob/89f3ab574d6a5551c9b724a06c3b707d739ec102/bats/fb-test-foreman-ansible.bats#L9-L10

Griffin-Sullivan: Updated

ekohl (commented):
One idea I had for this test was to verify a new report is made. I don't know if Hammer can do this, but my idea was:
* Take a timestamp
* Ensure there are no reports after the timestamp
* Run Puppet
* Ensure a new report is made

Now an even better one that actually catches https://projects.theforeman.org/issues/35833 is:

* First make sure `/etc/motd` has a content that doesn't match Puppet (something like `echo NOT IN SYNC > /etc/motd`)
* Run Puppet with --detailed-exitcodes and assert exit code 2
* Retrieve the generated report and ensure a change is recorded
* Run Puppet with --detailed-exitcodes and assert exit code 0
* Retrieve the generated report and ensure no change is recorded

If you implement that last step today, it will fail because of https://projects.theforeman.org/issues/35833 but you can mark it as pending.

evgeni:
> One idea I had for this test was to verify a new report is made. I don't know if Hammer can do this, but my idea was:
> 
>     * Take a timestamp
>     * Ensure there are no reports after the timestamp
>     * Run Puppet
>     * Ensure a new report is made

While I am not sure the timestamp idea is *easily* implementable, we can  check for the *ID* of the report (if any) and then compare that the new report has a newer/higher ID?

Once we fleshed out how to do this, we should apply the same logic to the Ansible report test, as it suffers from the same "checks that there is *any* report, not a *new* report" bug.

That said, I think the current test is already an improvement over the status quo, so we can easily split the "check for new" into a separate one, depending on Griffins availability?

ekohl (approved):
Agreed this is is an improvement. 

@Griffin-Sullivan for the future, please keep in mind we prefer the [imperative mood](https://cbea.ms/git-commit/#imperative) for commit messages.

PR #1672, opened 2023-04-27
Add -x to bats role to log the commands being run

zjhuntin:


roles/bats/tasks/main.yml
```diff
@@ -31,7 +31,7 @@
 
 - block:
     - name: "Run bats"
-      shell: "bats --tap {{ item }} > {{ bats_output_dir }}/{{ item }}.tap"
+      shell: "bats -x --tap {{ item }} > {{ bats_output_dir }}/{{ item }}.tap"
```
evgeni, line 34:
```suggestion
      shell: "bats --trace --tap {{ item }} > {{ bats_output_dir }}/{{ item }}.tap"
```

long opts are easier to read ;)

evgeni (approved):
`--ack-pending-long-opt-change`

evgeni (approved):

PR #1674, opened 2023-05-07
Fixed syntax glitch in README

sayan3296:
Few minor tweaks in README i.e. 
* There was an *`* at the end of  /etc/sudoers. 
* Without the `-e` , the `\n` will not be honored by `echo`
* And rather than adding stuff directly in `/etc/sudoers`, we should create user-specific file inside `sudoers.d`


ekohl (approved):
Thanks!

PR #1675, opened 2023-05-09
set bats_test_name_prefix for install and upgrade pipes

evgeni:


evgeni:
this should make it easier to differentiate results from different runs inside jenkins

ehelms (approved):

ekohl (approved):

PR #1693, opened 2023-08-11
Update README for change to centos8-stream

quartsize:
The quickstart doesn't currently work verbatim, I imagine since 95fc1ff (Drop CentOS 8 boxes where there's stream, 2022-02-08).

```
> vagrant up centos8-foreman-nightly
The machine with the name 'centos8-foreman-nightly' was not found configured for
this Vagrant environment.


ehelms (approved):
Thanks for the find!

PR #1696, opened 2023-08-21
Use the latest Vagrant centos8-stream image

ekohl:
Recently CentOS has started to publish Stream 8 boxes to Vagrant Cloud again, so it's no longer needed to specify the URL directly. This has the benefit that vagrant box update can work.

### DIFF
--- a/vagrant/boxes.d/00-centos.yaml
+++ b/vagrant/boxes.d/00-centos.yaml
@@ -31,7 +31,6 @@ boxes:
   centos8-stream:
     box_name: 'centos/stream8'
     disk_size: 40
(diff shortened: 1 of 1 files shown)


evgeni:
Given the recent comments in https://issues.redhat.com/browse/CS-1186 I am not sure we want to merge this right now?

ekohl:
Good question. I'm not sure it should block it. Yes, the current C9 box is broken, but C8 does work so I feel that the benefit of being able to use `vagrant box update` is worth it. Even if we need to pin it later if it does break.

evgeni:
Fair.

evgeni (approved):

PR #1706, opened 2023-09-21
Update Incremental export size check on katello-content.bats

Odilhao:
Pulpcore changed its compression logic, the value that bats expects should be update.

See https://github.com/pulp/pulpcore/issues/4434 and https://github.com/pulp/pulpcore/issues/4411 for more details


Odilhao:
This should also unbreak Katallo Nightly

parthaa (approved):
ACK

PR #1712, opened 2023-10-06
Add Candlepin pipeline to run puppet-candlepin tests

ehelms:
Starter test pipeline, this will require some changes to puppet-candlepin to allow configuring the repository to test. My goal is to provide this for https://github.com/theforeman/foreman-infra/issues/1937#issuecomment-1749500409

### DIFF
--- a/pipelines/candlepin/02-install.yml
+++ b/pipelines/candlepin/02-install.yml
@@ -1,0 +1,11 @@
+---
+- name: Setup git repo
+  become: True
+  hosts:
+   - "{{ forklift_name }}"
+  vars_files:
+    - ../vars/forklift_candlepin.yml
+  vars:
+    beaker_puppet_module: "puppet-candlepin"
+  roles:
+    - beaker
--- a/pipelines/candlepin/03-tests.yml
+++ b/pipelines/candlepin/03-tests.yml
@@ -1,0 +1,17 @@
+---
+- name: run tests
+  become: True
+  hosts:
+   - "{{ forklift_name }}"
+  vars_files:
+    - ../vars/forklift_candlepin.yml
+  tasks:
+    - name: Run acceptance tests
+      ansible.builtin.include_role:
+        name: beaker
+        tasks_from: test
+      vars:
+        beaker_puppet_module: "puppet-candlepin"
+        beaker_os: "{{ pipeline_os.replace('-stream', '') }}"
+        beaker_environment:
+          BEAKER_FACTER_CANDLEPIN_BASEURL: "https://stagingyum.theforeman.org/candlepin/{{ pipeline_version }}/el{{ ansible_distribution_major_version }}/x86_64"
--- a/roles/beaker/tasks/main.yml
+++ b/roles/beaker/tasks/main.yml
@@ -1,0 +1,50 @@
+---
+- name: Install podman-docker
+  package:
+    name: podman-docker
+    state: installed
+
+- name: Start podman
+  service:
+    name: podman
+    state: started
+
+- name: Install git
+  package:
+    name: git
+    state: installed
+
+- name: Enable ruby 2.7 module
+  command: dnf module enable -y ruby:2.7
+  when: ansible_distribution_major_version == "8"
+
+- name: enable powertools
+  command: dnf config-manager --set-enabled powertools
+  when: ansible_distribution_major_version == "8"
+
+- name: enable CRB
+  command: dnf config-manager --set-enabled crb
+  when: ansible_distribution_major_version == "9"
+
+- name: Install Ruby
+  package:
+    state: installed
+    name:
+      - ruby
+      - ruby-devel
+      - rubygem-bundler
+      - gcc-c++
+      - make
+      - redhat-rpm-config
+      - libyaml-devel
+
+- name: Clone puppet module
+  ansible.builtin.git:
+    repo: "https://github.com/theforeman/{{ beaker_puppet_module }}.git"
+    dest: "{{ beaker_puppet_module_path }}"
+    version: "{{ beaker_puppet_module_version | default(omit) }}"
+
+- name: Bundle install
+  command: bundle install
+  args:
+    chdir: "{{ beaker_puppet_module_path }}"
--- a/pipelines/candlepin.yml
+++ b/pipelines/candlepin.yml
@@ -1,0 +1,3 @@
+- import_playbook: candlepin/01-boxes.yml
+- import_playbook: candlepin/02-install.yml
+- import_playbook: candlepin/03-tests.yml
--- a/pipelines/candlepin/01-boxes.yml
+++ b/pipelines/candlepin/01-boxes.yml
@@ -1,0 +1,8 @@
+---
+- name: create vagrant boxes
+  hosts: localhost
+  become: False
+  vars_files:
+    - ../vars/forklift_candlepin.yml
+  roles:
+    - forklift
--- a/pipelines/vars/forklift_candlepin.yml
+++ b/pipelines/vars/forklift_candlepin.yml
@@ -1,0 +1,8 @@
+forklift_name: "pipe-candlepin-{{ pipeline_version }}-{{ pipeline_os }}"
+
+server_box:
+  box: "{{ pipeline_os }}"
+  memory: 4680
+
+forklift_boxes:
+  "{{ {forklift_name: server_box} }}"
--- a/roles/beaker/defaults/main.yml
+++ b/roles/beaker/defaults/main.yml
@@ -1,0 +1,7 @@
+beaker_base_environment:
+  BEAKER_HYPERVISOR: "docker"
+  BEAKER_provision: "yes"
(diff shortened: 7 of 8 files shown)


ehelms:
@ekohl How would you make https://github.com/theforeman/puppet-candlepin/blob/master/spec/setup_acceptance_node.pp#L20 configurable for running the beaker tests dynamically? I can imagine wanting to set:

 * Candlepin version for a known repository location
 * A different repository URL for testing against different locations (e.g. stage, prod)

ekohl:
I'm (mostly) happy with how puppet-pulpcore does it.

* [pulpcore::repo](https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/repo.pp) has a parameter
* That `version` is set in [setup_acceptance_node.pp](https://github.com/theforeman/puppet-pulpcore/blob/41d6f1f7b79b8e4320509b38f0b342aa2adc3877/spec/setup_acceptance_node.pp#L12-L14) using [environment variables](https://github.com/voxpupuli/voxpupuli-acceptance#environment-variables-to-facts)
* These [versions](https://github.com/theforeman/puppet-pulpcore/blob/41d6f1f7b79b8e4320509b38f0b342aa2adc3877/.github/workflows/ci.yml#L70-L73) are [set in the matrix build](https://github.com/theforeman/puppet-pulpcore/blob/41d6f1f7b79b8e4320509b38f0b342aa2adc3877/.github/workflows/ci.yml#L91)

I'm unhappy with the last part: [gha-puppet](https://github.com/voxpupuli/gha-puppet) should have a way to do this without having to copy the whole workflow.

My plan to solve that was to first get https://github.com/voxpupuli/puppet_metadata/pull/35 done. That makes it possible to solve [this](https://github.com/voxpupuli/gha-puppet/blob/4743dd4bd18d1c11f70889b0d7cac6fe77e8e758/.github/workflows/beaker.yml#L135-L136) TODO. Then we can consider ways to input those versions to CI.

pipelines/candlepin/02-install.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Install Ruby
+      package:
+        name:
+          - ruby
+          - ruby-devel
+          - rubygem-bundler
+          - gcc-c++
+          - make
+          - redhat-rpm-config
+        state: installed
```
evgeni, line 36: given you're planning to re-use this for pulpcore, I think this could go to a `beaker` role?

ehelms: Fair.

pipelines/candlepin/03-tests.yml
```diff
@@ -0,0 +1,18 @@
...
+    - ../vars/forklift_candlepin.yml
+  tasks:
+    - name: Run acceptance tests
+      command: bundle exec rake beaker
+      args:
+        chdir: /src/puppet-candlepin
+      environment:
+        BEAKER_HYPERVISOR: "docker"
+        BEAKER_provision: "yes"
+        BEAKER_setfile: "centos8-64{hostname=centos8-64.example.com}"
```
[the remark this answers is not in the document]
ehelms: Good catch.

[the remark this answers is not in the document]
ehelms: Implemented.

pipelines/candlepin/02-install.yml
ehelms: Done - let me know what you think.

pipelines/candlepin/03-tests.yml
```diff
@@ -0,0 +1,18 @@
...
+    - name: Run acceptance tests
+      command: bundle exec rake beaker
+      args:
+        chdir: /src/puppet-candlepin
+      environment:
+        BEAKER_HYPERVISOR: "docker"
+        BEAKER_provision: "yes"
+        BEAKER_setfile: "{{ pipeline_os | regex_replace('-stream', '') }}-64{hostname={{ pipeline_os | regex_replace('-stream', '') }}-64.example.com}"
+        BEAKER_destroy: "no"
+        BEAKER_FACTER_CANDLEPIN_VERSION: "{{ pipeline_version }}"
```
evgeni, line 18:
```suggestion
      environment:
        BEAKER_HYPERVISOR: "docker"
        BEAKER_provision: "yes"
        BEAKER_setfile: "{{ beaker_os }}-64{hostname={{ beaker_os }}-64.example.com}"
        BEAKER_destroy: "no"
        BEAKER_FACTER_CANDLEPIN_VERSION: "{{ pipeline_version }}"
      vars:
        beaker_os: "{{ pipeline_os | regex_replace('-stream', '') }}"
```

does that work?

roles/beaker/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+      - gcc-c++
+      - make
+      - redhat-rpm-config
+      - libyaml-devel
+
+- name: Clone puppet module
+  ansible.builtin.git:
+    repo: "https://github.com/theforeman/{{ beaker_puppet_module }}.git"
+    dest: "/src/{{ beaker_puppet_module }}"
+    version: "{{ beaker_puppet_module_version | default('master') }}"
```
evgeni, line 41:
```suggestion
    version: "{{ beaker_puppet_module_version | default(omit) }}"
```

will that take the default branch, whichever it is?

pipelines/candlepin/02-install.yml
evgeni: one comment, but I like!

evgeni:
@ehelms I moved a few things around, to make it even more re-useable, and overall I like it :)

pipelines/candlepin/03-tests.yml
```diff
@@ -0,0 +1,17 @@
...
+  vars_files:
+    - ../vars/forklift_candlepin.yml
+  tasks:
+    - name: Run acceptance tests
+      ansible.builtin.include_role:
+        name: beaker
+        tasks_from: test
+      vars:
+        beaker_puppet_module: "puppet-candlepin"
+        beaker_os: "{{ pipeline_os | regex_replace('-stream', '') }}"
```
evgeni, line 15: this even works with `pipeline_os=almalinux8`, nice!

pipelines/candlepin/03-tests.yml
```diff
@@ -0,0 +1,17 @@
...
+        beaker_environment:
+          BEAKER_FACTER_CANDLEPIN_VERSION: "{{ pipeline_version }}"
```
evgeni, line 17: one thing that caught me is that *today* this is the Katello version, not the Candlepin version, so when I ran this with `4.3` it installed Candlepin from Katello 4.3, which was 4.1.something and that didn't like Java 17 at all XD

works fine with `nightly`.

ehelms: That's cause the other piece of the puzzle is being able to set the baseurl on the puppet-module -- https://github.com/theforeman/puppet-candlepin/pull/249

evgeni (approved):

ehelms:
This now relies on https://github.com/theforeman/puppet-candlepin/pull/252 and relies upon two additional variables:

```
ansible-playbook pipelines/candlepin.yml -e pipeline_os=centos8-stream -e pipeline_version=nightly -e pipeline_beaker_os=centos9 -e pipeline_beaker_os_version=9
```

evgeni:
Can't we deduct the additional variables from the `pipeline_os`? We have (today) no good way for passing in additional vars via Jenkins. We could have them in some vars file… but it feels odd given that IMHO all the data should already be present.

pipelines/candlepin/03-tests.yml
```diff
@@ -0,0 +1,17 @@
...
+  tasks:
+    - name: Run acceptance tests
+      ansible.builtin.include_role:
+        name: beaker
+        tasks_from: test
+      vars:
+        beaker_puppet_module: "puppet-candlepin"
+        beaker_os: "{{ pipeline_beaker_os }}"
+        beaker_environment:
+          BEAKER_FACTER_CANDLEPIN_BASEURL: "https://stagingyum.theforeman.org/candlepin/{{ pipeline_version }}/el{{ pipeline_beaker_os_version }}/x86_64"
```
evgeni, line 17: technically, given this runs *on* the vagrant box, this could be
```suggestion
          BEAKER_FACTER_CANDLEPIN_BASEURL: "https://stagingyum.theforeman.org/candlepin/{{ pipeline_version }}/el{{ ansible_distribution_major_version }}/x86_64"
```

ehelms: That won't work for testing EL8 and EL9  when the host itself is only EL8. That is why I felt these additional variables had to be introduced.

evgeni: Agreed. I was working under the assumption we're keeping the current mode where we spin up a VM that matches the target OS.

I think I'm missing some bigger piece how this should be executed by automation.

ehelms: Fair, while not efficient, that is how this workflow works and we could keep that connection and assumption.

PR #1713, opened 2023-10-19
add fapolicyd role

evgeni:


evgeni:
Is there a way to achieve that besides overriding `ExecStart=` of the systemd unit?

ehelms:
> Is there a way to achieve that besides overriding `ExecStart=` of the systemd unit?

Manually starting it and running it in the background:

```
fapolicyd --debug 2> fapolicy.output &
```

And/or running it in permissive mode as part of this. We can use `noup` or `daemonize` to do it. I can understand if this feels ugly or invasive. Figured I'd ask since running in debug is common practice for debugging.

evgeni:
Yeah, that indeed feels ugly.

(RFE against `fapolicyd.rpm` to have an `EnvironmentFile` in the unit and use `$FAPOLICYD_OPTIONS` in `ExecStart`…)

ekohl (approved):

PR #1714, opened 2023-10-20
update ansible-lint to 6.x

evgeni:


roles/foreman/meta/main.yml
```diff
@@ -2,6 +2,7 @@
 dependencies:
   - role: umask
   - role: selinux
+  - role: fapolicyd
```
ekohl, line 5: This looks unrelated

evgeni: err, indeed. how did that land here.

ekohl (approved):

PR #1715, opened 2023-10-25
Fix ansible-error: unsupported locale setting

m-bucher:
Fixes the following error on Debian12 bookworm with locale `de_DE.UTF-8`.
Not fully sure, which aspect is to blame :innocent: 

```
==> centos8-katello-devel: Running provisioner: disk_resize (ansible)...
Vagrant gathered an unknown Ansible version:


and falls back on the compatibility mode '1.8'.

Alternatively, the compatibility mode can be specified in your Vagrantfile:
https://www.vagrantup.com/docs/provisioning/ansible_common.html#compatibility_mode

    centos8-katello-devel: Running ansible-playbook...
ERROR: Ansible could not initialize the preferred locale: unsupported locale setting
Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.

### CODE REVIEW

**Comment by ekohl** [comment]
> Not fully sure, which aspect is to blame 😇

You probably have `de_DE.UTF-8` configured on your host and the Debian 12 image doesn't have that locale. We also have code that ensures the locale should be available, but it's written in Ansible ...

https://github.com/theforeman/forklift/blob/17014e2b11a5a9e645edac05c6569431eb7653a6/roles/foreman_installer/tasks/locales.yml#L11-L14

**Comment by m-bucher** [comment]
@ekohl so I guess `locale_gen` should be the first provisioning-task we do on the new machine?

**Comment by ekohl** [comment]
Yes, I think it would make sense to extract it to its own role and run it as the very first step. That is, if you can even do that. Not sure where it happens.

> Vagrant gathered an unknown Ansible version:
> 
> 
> and falls back on the compatibility mode '1.8'.

Perhaps this is actually the root cause. Looking at the source, this is how it attempts to detect it:
https://github.com/hashicorp/vagrant/blob/f33ac0bc98bdbcb421331d62fba42ba39d1aff85/plugins/provisioners/ansible/provisioner/host.rb#L113-L136

Perhaps this fails in your setup?

**Comment by m-bucher** [comment]
I think I understand now:
My system has:
```
$ locale -a
C
C.utf8
de_DE.utf8
POSIX
```
forklift's `Vagrantfile` sets `en_US.UTF-8` [here](https://github.com/theforeman/forklift/blob/17014e2b11a5a9e645edac05c6569431eb7653a6/Vagrantfile#L5).

So either I have to do `locale-gen en_US.UTF-8` on my system or `locale-gen de_DE.utf8` on the guest-system.
From a user's perspective I'd say do the latter.

ekohl:
But how do you have Ansible installed?

m-bucher:
> But how do you have Ansible installed?

https://packages.debian.org/bookworm/ansible-core

PR #1725, opened 2023-11-15
Update Pulpcore to 3.39

ianballou:
Related: 

- [x] https://github.com/Katello/katello/pull/10791
- [x] https://github.com/theforeman/foreman-packaging/pull/9958
- [x] https://github.com/theforeman/foreman-packaging/pull/9957
- [ ] https://github.com/theforeman/foreman-packaging/pull/9989

### DIFF
--- a/roles/pulpcore_repositories/tasks/main.yml
+++ b/roles/pulpcore_repositories/tasks/main.yml
@@ -24,4 +24,13 @@
     creates: /etc/dnf/modules.d/pulpcore.module
   when:
     - pulpcore_repositories_version is version('3.16', '>=')
+    - pulpcore_repositories_version is version('3.39', '<')
+    - ansible_distribution_major_version == '8'
+
+- name: 'Disable pulpcore:el8 module'
+  file:
+    path: /etc/dnf/modules.d/pulpcore.module
+    state: absent
+  when:
+    - pulpcore_repositories_version is version('3.39', '>=')
     - ansible_distribution_major_version == '8'
--- a/roles/pulpcore_repositories/defaults/main.yml
+++ b/roles/pulpcore_repositories/defaults/main.yml
@@ -1,2 +1,2 @@
 ---
-pulpcore_repositories_version: '3.28'
+pulpcore_repositories_version: '3.39'
--- a/vagrant/config/versions.yaml
+++ b/vagrant/config/versions.yaml
@@ -65,7 +65,7 @@ installers:
 
   - foreman: 'nightly'
     katello: 'nightly'
-    pulpcore: '3.28'
+    pulpcore: '3.39'
     puppet: 7
     boxes:
       - 'almalinux8'


ianballou:
I've also made it so the pulpcore module is not installed unless the pulpcore version is between 3.16 and 3.39.  I'm not 100% certain if I should add the second conditional like I did.

pipelines/pulpcore/02-install.yml
```diff
@@ -14,7 +14,7 @@
       state: present
     when:
      - pipeline_version is defined
-     - pipeline_version == 'nightly' or pipeline_version is version('3.28', '>=')
+     - pipeline_version == 'nightly' or pipeline_version is version('3.39', '>=')
```
ekohl, line 17: This will break the 3.28 pipeline since it'll no longer install PG 12 there. 3.39 should match >= 3.28 anyway:
```suggestion
     - pipeline_version == 'nightly' or pipeline_version is version('3.28', '>=')
```

roles/pulpcore_repositories/tasks/main.yml
```diff
@@ -24,4 +24,5 @@
     creates: /etc/dnf/modules.d/pulpcore.module
   when:
     - pulpcore_repositories_version is version('3.16', '>=')
+    - pulpcore_repositories_version is version('3.39', '<')
```
ekohl, line 27: I think this is correct.

On a related note: I wonder if we need to disable the module on `>= 3.39`.

ekohl (changes requested):

roles/pulpcore_repositories/tasks/main.yml
```diff
@@ -24,4 +24,16 @@
...
+    - ansible_distribution_major_version == '8'
+
+- name: 'Disable pulpcore:el8 module'
+  command: dnf module disable -y pulpcore:el8
+  # can't use the `dnf` module for modules without a default stream
+  # https://github.com/ansible/ansible/issues/56504
+  # https://github.com/ansible/ansible/issues/64852
+  args:
+    creates: /etc/dnf/modules.d/pulpcore.module
```
ekohl, line 36: I don't think this is right. The file will already be there if it is enabled and instead it will actually skip this step if it's enabled (which is true for upgrades).

Just for background:
```console
# dnf module enable pulpcore:el8
# ls /etc/dnf/modules.d
pulpcore.module  python39.module
# dnf module disable pulpcore
# ls /etc/dnf/modules.d
pulpcore.module  python39.module
# cat /etc/dnf/modules.d/pulpcore.module
[pulpcore]
name=pulpcore
stream=
profiles=
state=disabled
```

Note that the enable part also enabled python39 but disable doesn't disable python. It looks like ansible-core has rebased to python3.11 (al least in CentOS Stream 8), but worth considering in the release notes.

ekohl: @evgeni any thoughts on using this for idempotency?

```yaml
- name: 'Disable pulpcore:el8 module'
  command: dnf module disable -y pulpcore:el8
  args:
    removes: /etc/dnf/modules.d/pulpcore.module

- name: Delete pulpcore module
  file:
    path: /etc/dnf/modules.d/pulpcore.module
    state: absent
```

evgeni: This looks reasonable. The `removes` on the disable command is a "lie", but then you fix it up and things should be fine.

OTOH, we could also *just* remove the file, that's the only thing that carries the info about the module state.

evgeni: I went with remove only and updated the PR

evgeni:
https://github.com/theforeman/foreman-operations-collection/blob/develop/roles/foreman_repositories/tasks/redhat-modules.yml#L23 will need updating too

ekohl (approved):

evgeni:
https://github.com/theforeman/foreman-operations-collection/pull/142

PR #1729, opened 2023-11-15
Switch Pulpcore testing to using puppet-pulpcore test suite

ehelms:
Given we install Pulpcore using puppet-pulpcore, the idea here is to vet packaging changes against the puppet module rather than the Pulp Ansible installer.

This is in draft mode because first the puppet module needs to support configuring the baseurl in order to point the testing at `stagingyum.theforeman.org`.

### DIFF
--- a/pipelines/pulpcore/02-install.yml
+++ b/pipelines/pulpcore/02-install.yml
@@ -22,7 +22,7 @@
   become: True
   vars_files:
     - ../vars/install_base.yml
-    - ../vars/forklift_{{ pipeline_type }}.yml
+    - ../vars/forklift_pulpcore.yml
   environment:
     DJANGO_SETTINGS_MODULE: pulpcore.app.settings
   pre_tasks:
--- a/pipelines/pulpcore/02-install.yml
+++ b/pipelines/pulpcore/02-install.yml
@@ -1,15 +1,20 @@
 ---
-- name: Setup git repo
+- name: Enable Postgresql 12 module if necessary
   become: True
   hosts:
    - "{{ forklift_name }}"
   vars_files:
     - ../vars/forklift_pulpcore.yml
   tasks:
-    - name: Install podman-docker
-      package:
-        name: podman-docker
-        state: installed
+  - name: Enable postgresql 12 module
+    ansible.builtin.dnf:
+      name: '@postgresql:12'
+      state: present
+    when:
+     - pipeline_version is defined
+     - pipeline_version != 'nightly' or pipeline_version is version('3.40', '<')
+     - pipeline_os is defined
+     - pipeline_os is search("centos8-stream")
 
 - name: install pulpcore
   hosts:
@@ -21,13 +26,33 @@
   environment:
     DJANGO_SETTINGS_MODULE: pulpcore.app.settings
   pre_tasks:
-    - when: pipeline_version != 'nightly' and pipeline_version is version('3.28', '<=' )
+    - when: pipeline_version != 'nightly' and pipeline_version is version('3.28', '==' )
       set_fact:
         pulp_pkg_repo: "http://koji.katello.org/releases/yum/pulpcore-{{ pipeline_version }}/el{{ ansible_distribution_major_version }}/$basearch/"
-    - when: pipeline_version == 'nightly' or pipeline_version is version('3.39', '>=')
+    - when: pipeline_version != 'nightly' and pipeline_version is version('3.39', '==')
       set_fact:
         pulp_pkg_repo: "https://stagingyum.theforeman.org/pulpcore/{{ pipeline_version }}/el{{ ansible_distribution_major_version }}/$basearch/"
         pulp_pkg_name_prefix: "python3.11-"
   roles:
-    - epel_repositories
-    - pulp.pulp_installer.pulp_all_services
+    - role: epel_repositories
+      when:
+        - pipeline_version is defined
+        - pipeline_version != 'nightly' and pipeline_version is version('3.28', '==' )
+    - role: pulp.pulp_installer.pulp_all_services
+      when:
+        - pipeline_version is defined
+        - pipeline_version != 'nightly' or pipeline_version is version('3.39', '==')
+
+- name: Setup git repo
+  become: True
+  hosts:
+   - "{{ forklift_name }}"
+  vars_files:
+    - ../vars/forklift_pulpcore.yml
+  vars:
+    beaker_puppet_module: "puppet-pulpcore"
+  roles:
+    - role: beaker
+      when:
+        - pipeline_version is defined
+        - pipeline_version == 'nightly' or pipeline_version is version('3.40', '>=')
--- a/pipelines/pulpcore/02-install.yml
+++ b/pipelines/pulpcore/02-install.yml
@@ -1,22 +1,15 @@
 ---
-- name: Enable Postgresql 12 module if necessary
+- name: Setup git repo
   become: True
   hosts:
-   - "{{ forklift_server_name }}"
+   - "{{ forklift_name }}"
   vars_files:
-    - ../vars/install_base.yml
-    - ../vars/forklift_{{ pipeline_type }}.yml
-
+    - ../vars/forklift_pulpcore.yml
   tasks:
-  - name: Enable postgresql 12 module
-    ansible.builtin.dnf:
-      name: '@postgresql:12'
-      state: present
-    when:
-     - pipeline_version is defined
-     - pipeline_version == 'nightly' or pipeline_version is version('3.28', '>=')
-     - pipeline_os is defined
-     - pipeline_os is search("centos8-stream")
+    - name: Install podman-docker
+      package:
+        name: podman-docker
+        state: installed
 
 - name: install pulpcore
   hosts:
--- a/pipelines/pulpcore/03-tests.yml
+++ b/pipelines/pulpcore/03-tests.yml
@@ -1,5 +1,5 @@
 ---
-- name: run tests
+- name: run tests beaker
   become: True
   hosts:
    - "{{ forklift_name }}"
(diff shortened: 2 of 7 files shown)


pipelines/pulpcore/03-tests.yml
```diff
@@ -1,10 +1,18 @@
...
+  tasks:
+    - name: Run acceptance tests
+      command: bundle exec rake beaker
+      args:
+        chdir: /src/puppet-pulpcore
+      environment:
+        BEAKER_HYPERVISOR: "docker"
+        BEAKER_provision: "yes"
+        BEAKER_setfile: "centos8-64{hostname=centos8-64.example.com}"
+        BEAKER_destroy: "no"
```
evgeni, line 17: why not destroying them?

pipelines/vars/forklift_pulpcore.yml
```diff
@@ -1,9 +1,11 @@
+forklift_name: "pipe-pulp-{{ pipeline_version }}-{{ pipeline_os }}"
```
evgeni, line 1: why did you have to define that? it should come from `install_base.yml`

pipelines/pulpcore/03-tests.yml
```diff
@@ -1,10 +1,18 @@
...
-  roles:
-    - pulp.pulp_installer.pulp_health_check
+    - ../vars/forklift_pulpcore.yml
+  tasks:
+    - name: Run acceptance tests
+      command: bundle exec rake beaker
+      args:
+        chdir: /src/puppet-pulpcore
+      environment:
+        BEAKER_HYPERVISOR: "docker"
```
evgeni, line 14: I don't particularly like that we spin up a centos box just to run a centos container inside, especially as our primary way of deployment is not a container.

Shouldn't beaker default to using the machine it's running on when we do not set BEAKER_HYPERVIRSOR?

pipelines/pulpcore/03-tests.yml
ehelms: Makes debugging easier if it sticks around. We end up destroying the box anyway so no harm in keeping it.

pipelines/vars/forklift_pulpcore.yml
ehelms: The *should* did not pan out

pipelines/pulpcore/03-tests.yml
ehelms: I would argue it's not *just* to run a container inside -- we also install dependencies needed for the puppet module to perform testing: https://github.com/theforeman/forklift/pull/1729/files#diff-dec5007e2a651ace599f5154ceb5187aa75fb8373133bac84a45fd15840fbe9dR19-R41

Especially for our CI, that means we have a clean environment every time (yes, I know CentOS CI **starts** with a clean box and thus we could shorten this). Locally, this also means we have a clean environment every time. And, it's consistent.

evgeni:
Overall I think this will work, and is down the line the more meaningful test (as it uses the same codebase as we use for prod deployments later on).

What is the plan wrt older Pulpcore releases? Those do not reside on stagingyum and often can't be tested with `puppet-pulpcore` `master`. Should we keep the old pipe as "pulpcore_old" or something, and utilize that for releases before 3.39?

pipelines/pulpcore/03-tests.yml
evgeni: Fair.

I was also thinking about `BEAKER_HYPERVISOR: vagrant` which would fit nicely if we wouldn't run this all from inside a vagrant box already.

It's hypervisors all the way down.

pipelines/vars/forklift_pulpcore.yml
evgeni: Hmm. Weird. (I mean, you stopped including it, but still)

pipelines/pulpcore/03-tests.yml
ehelms: We could possibly do container-inception as well, however, I was not trying to re-invent the wheel here. Rather my goal was to switch away from the Ansible based Pulp installer and to using our own test suite that we vet a working Pulp against.

ehelms:
> What is the plan wrt older Pulpcore releases? Those do not reside on stagingyum and often can't be tested with `puppet-pulpcore` `master`. Should we keep the old pipe as "pulpcore_old" or something, and utilize that for releases before 3.39?

That is part of the reason this is in draft :) So I appreciate you having thought it through and thrown out an idea. I think keeping the status quo for prior releases is the easiest option.

ehelms:
BTW, I was intending to vet this model on new ground first by doing it with Candlepin if you want to review that first -- https://github.com/theforeman/forklift/pull/1712

evgeni:
> > What is the plan wrt older Pulpcore releases? Those do not reside on stagingyum and often can't be tested with `puppet-pulpcore` `master`. Should we keep the old pipe as "pulpcore_old" or something, and utilize that for releases before 3.39?
> 
> That is part of the reason this is in draft :) So I appreciate you having thought it through and thrown out an idea. I think keeping the status quo for prior releases is the easiest option.

We could also be adventurous.

According to https://github.com/theforeman/forklift/blob/master/vagrant/config/versions.yaml we have only to care for 3.22+, so really 3.22, 3.28 and 3.39 (after this whole change).
Today `puppet-pulpcore` does support those old versions and given the whole repo-url stuff is merged, we could go and use the current state (before merging the 3.39 changes) as the version to test old releases, and `master` for 3.39+

pipelines/pulpcore/02-install.yml
```diff
@@ -1,40 +1,52 @@
...
-  - name: Enable postgresql 12 module
-    ansible.builtin.dnf:
-      name: '@postgresql:12'
-      state: present
-    when:
-     - pipeline_version is defined
-     - pipeline_version == 'nightly' or pipeline_version is version('3.28', '>=')
-     - pipeline_os is defined
-     - pipeline_os is search("centos8-stream")
+    - name: Install podman-docker
```
ekohl, line 9: I've used beaker-docker a few times with plain podman. How I do it on my Fedora 38 box:
```console
$ systemctl start --user podman.socket
$ export DOCKER_HOST=unix:///run/user/$UID/podman/podman.sock
```

Does it make sense to introduce a podman role for this?

pipelines/pulpcore/02-install.yml
```diff
@@ -1,40 +1,52 @@
...
+          - ruby
+          - ruby-devel
+          - rubygem-bundler
+          - gcc-c++
+          - make
+          - redhat-rpm-config
+          - libyaml-devel
+        state: installed
+
+    - name: Clone puppet-pulpcore
```
ekohl, line 44: We'll want to add a parameter to select the branch. In git master we may drop support for a version that still has pipelines. If we can then pin it to some stable branch it should still be good.

ekohl (commented):
We may end up with a circular dependency here. puppet-pulpcore tests its CI against nightly, but you need puppet-pulpcore to pass to promote nightly.

pipelines/pulpcore/03-tests.yml
ekohl: Today we spin up very expensive bare metal boxes so we can use virtualization. If we don't, it's much cheaper for CentOS CI if we request a VM. So I'd say we either go with `BEAKER_HYPERVISOR=vagrant_libvirt` so we can also test SELinux (which we can't inside containers) or enhance our Duffy integration so it can request VMs via a parameter.

Today that's hardcoded here:
https://github.com/theforeman/jenkins-jobs/blob/d78f0f3b828444b109105fa5d27927be3f158c31/centos.org/ansible/provision_duffy.yml#L6

Perhaps also a good excuse to address https://github.com/theforeman/foreman-infra/issues/1828.

evgeni:
> We may end up with a circular dependency here. puppet-pulpcore tests its CI against nightly, but you need puppet-pulpcore to pass to promote nightly.

We certainly will (even tho in many cases in the past adding a new version was more something optional, as it enabled new features or something), and I guess we won't get away w/o running releases with an unmerged puppet-pulpcore branch for a moment in the case where it's truly needed (like now with 3.39), but this seems "ok"?

(I mean, right now, we point nightly/3.39 at a fork of the old installer pcreech hacked together, so same same)

ehelms:
I was thinking that even though it may be circular, it would help identify if new Pulp versions / dependencies arriving via the nightly pipeline introduce a breakage we need to be aware of in puppet-pulpcore as early as possible.

ekohl:
We can work around it, by pointing a PR (temporarily) at staging (https://github.com/theforeman/puppet-pulpcore/commit/fa2bd4f722a09030a55988b22b02be2efb79febf makes that possible). My main issue is that we should acknowledge this by documenting is somewhere.

pipelines/pulpcore/02-install.yml
wbclark: You could include the role from https://github.com/theforeman/forklift/pull/1580 and then set the source repository and any number of desired remotes and branches in `pipelines/vars/forklift_pulpcore.yml`

That handles the bundle install as well if you add `install_gems: true`

Odilhao:
Now that https://github.com/theforeman/puppet-pulpcore/pull/333 was merged, we can look at this PR again.

playbooks/setup_forklift.yml
```diff
@@ -95,13 +95,13 @@
...
       command:
-        cmd: ansible-galaxy collection install -r requirements-pulp-322.yml
+        cmd: ansible-galaxy collection install -r requirements-pulp.yml
         chdir: "{{ forklift_dest }}"
       when:
         - forklift_install_pulp_from_galaxy
         - pipeline_version is defined
         - pipeline_version != 'nightly'
-        - pipeline_version is version('3.18', '>=')
+        - pipeline_version is version('3.28', '>=')
```
ekohl, line 104: We've dropped < 3.28
```suggestion
```

pipelines/pulpcore/02-install.yml
```diff
@@ -2,19 +2,17 @@
...
+    - ../vars/forklift_pulpcore.yml
   tasks:
   - name: Enable postgresql 12 module
     ansible.builtin.dnf:
       name: '@postgresql:12'
       state: present
     when:
      - pipeline_version is defined
-     - pipeline_version == 'nightly' or pipeline_version is version('3.28', '>=')
+     - pipeline_version != 'nightly' and pipeline_version is version('3.28', '>=')
```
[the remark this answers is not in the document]
Odilhao: We will branch a new version soon, should we change to 
`pipeline_version != 'nightly' or pipeline_version is version('3.40', '<')`?

Odilhao (approved):

Odilhao (approved):

PR #1732, opened 2023-11-21
Add Foreman 3.9 version config

zjhuntin:


vagrant/config/versions.yaml
```diff
@@ -63,6 +63,16 @@ installers:
       - 'debian11'
       - 'ubuntu2004'
 
+  - foreman: '3.9'
+    katello: '4.11'
+    pulpcore: '3.39'
```
ekohl, line 68: This is assuming we'll complete 3.39, which we're not certain of yet.

ekohl (approved):
Looks good, but when we merge this our upgrade pipelines will include it so let's wait until we actually have branched.

PR #1735, opened 2023-11-27
Skip ostree tests if pulpcore 3.39

pcreech:


ianballou:
I think the ostree ones were actually passing but failing silently. The export tests seemed to be the actual failure points https://ci.theforeman.org/blue/rest/organizations/jenkins/pipelines/katello-nightly-rpm-pipeline/runs/1838/log/?start=0

pcreech:
ack, moving around.  I can probably go ahead and handle the .gz failures as well.  what's the new extension?

dralley:
It's just .tar, rather than .tar.gz

pcreech:
Updated, added second commit switching .gz to .tar

evgeni:
> Updated, added second commit switching .gz to .tar

This will break for all releases using versions before 3.39, can we use `*.tar*`, assuming the old ones are .tar.gz?

evgeni:
Also, doesn't this disable*all* Export/Import tests?

pcreech:
It does disable all, it appeared the issue was in a few.  Is the suggestion to keep incremental in?

evgeni:
Ah, I think I've read the original comment wrongly. The idea is to disable all import/export tests for 3.39 until that one Pulpcore big is fixed?

bats/fb-katello-content.bats
```diff
@@ -122,11 +122,13 @@ setup() {
...
+  tSkipIfPulp339
+
   hammer content-export complete version --organization="${ORGANIZATION}" \
     --content-view="${CONTENT_VIEW}" --version="1.0"
   export_version_id=$(hammer --output csv --no-headers content-view version show --version="1.0" --content-view="${CONTENT_VIEW}" --organization="${ORGANIZATION}" \
     --fields=id)
-  actual_size=$(du -k "$(hammer --output csv --no-headers content-export list --content-view-version-id=$export_version_id --fields="path")"/*.gz | cut -f 1)
+  actual_size=$(du -k "$(hammer --output csv --no-headers content-export list --content-view-version-id=$export_version_id --fields="path")"/*.tar | cut -f 1)
 
  [ $actual_size -ge 40 ]
```
ekohl, line 133: Not for this PR, but can we actually list the contents of the tar file and check that? It would also validate the resulting file(s) are actually valid tar files.

As for the rename: don't we need to keep the gzip bits for older Pulp versions?

bats/foreman_helper.bash
```diff
@@ -28,6 +28,12 @@ tKatelloVersion() {
   ) | cut -d. -f1-2
 }
 
+tSkipIfPulp39() {
```
ekohl, line 31:
```suggestion
tSkipIfPulp339() {
```

ekohl (changes requested):
At least the function name needs to be fixed.

bats/foreman_helper.bash
pcreech: Thanks Visual Studio Code!  Or, more specifically, the need to *save* files in editors.... :D

bats/fb-katello-content.bats
pcreech: Based on evgeni's suggestion, and verifying with dalley that it is so, I've switched it to do ".tar*' to encapsulate gz without needing to resort to adding complexity with if/else logic

bats/foreman_helper.bash
pcreech: This _should_ be fixed now. Ctl+s to the rescue!

pcreech:
@ekohl @evgeni UPdated

evgeni (approved):
It looks as nice as such a hack can be :)

PR #1743, opened 2023-12-07
Add candlepin repositories for Katello staging

ekohl:


Odilhao:
I think we also need to change the https://github.com/theforeman/forklift/blob/b7ef5bb829b508eca65e4996f530aa6052982c0a/roles/katello_repositories/tasks/_stagingyum_staging_repos.yml#L11 `katello_repositories_version` here to `candlepin_repositories_version`

roles/candlepin_repositories/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+pulpcore_repositories_version: '4.3'
```
evgeni, line 2: nope

roles/candlepin_repositories/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+    gpgcheck: no
+    enabled: yes
+
+- name: 'Enable pki-core module'
+  command: dnf module enable -y pki-core
+  # can't use the `dnf` module for modules without a default stream
+  # https://github.com/ansible/ansible/issues/56504
+  # https://github.com/ansible/ansible/issues/64852
+  args:
+    creates: /etc/dnf/modules.d/pki-core.module
```
evgeni, line 16: *technically* this is not required as people will enabled `katello:el8` which pulls that in?

roles/candlepin_repositories/defaults/main.yml
Odilhao: This should be `candlepin_repositories_version`

roles/forklift_versions/library/forklift_versions.py
```diff
@@ -86,17 +86,15 @@ def main():
...
+                if 'candlepin' in version:
+                    forklift_vars['candlepin_repositories_version'] = version['candlepin']
                 ret = forklift_vars
                 break
     else:
         possible_versions = set()
         for version in reversed(versions['installers']):
             if not scenario_os in version['boxes']:
                 continue
-            # this is a hack, to be removed once all Katello versions support EL8 in the same manner Foreman does
```
evgeni, line 96: :laughing:

evgeni (changes requested):

roles/candlepin_repositories/tasks/main.yml
ekohl: For now I will leave it there so you can use candlepin standalone.

Odilhao (approved):

PR #1744, opened 2023-12-07
Add condition for foreman-client-repo based on bats version

Odilhao:
Aims to fix https://jenkins-foreman.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/foreman-pipeline-katello-rpm-4.11/detail/foreman-pipeline-katello-rpm-4.11/4/pipeline/94


roles/foreman_client_repositories/tasks/release_repo.yml
```diff
@@ -4,3 +4,6 @@
     name: https://yum.theforeman.org/client/{{ foreman_client_repositories_version }}/{{ foreman_client_repositories_dists[ansible_os_family] }}{{ ansible_distribution_major_version }}/x86_64/foreman-client-release.rpm
     disable_gpg_check: True
     state: present
+  when:
+    - katello_repositories_version == "nightly" or foreman_client_repositories_version is version('3.9', '<')
```
evgeni, line 8: `katello_repositories_version` is not necessarily defined in the client context.

roles/foreman_client_repositories/tasks/release_repo.yml
```diff
@@ -4,3 +4,6 @@
     name: https://yum.theforeman.org/client/{{ foreman_client_repositories_version }}/{{ foreman_client_repositories_dists[ansible_os_family] }}{{ ansible_distribution_major_version }}/x86_64/foreman-client-release.rpm
     disable_gpg_check: True
     state: present
+  when:
+    - foreman_client_repositories_version is version('3.9', '<')
+    - ansible_distribution_major_version == '8'
```
evgeni, line 9: why this limitation?

evgeni:
I think this is wrong, we rather should set the staging_source correctly: https://github.com/theforeman/forklift/pull/1745

PR #1747, opened 2023-12-11
do not try to upgrade PostgreSQL

evgeni:
this role was dropped from FOC and we don't need it anymore


ehelms (approved):
There is no try.

PR #1752, opened 2024-01-04
allow overriding the server/proxy OS

evgeni:
this enables mixed-OS testing


evgeni:
the changes to the definitions that do not have a proxy are just for consistency (and in the hope they will get proxies at some point)

ehelms (approved):

PR #1757, opened 2024-01-26
clean up old Ubuntu initrds

evgeni:
for some reason /boot contains both initrds with and without .img even if the ones with are not used, however, as /boot space is limited they prevent upgrades of packages, failing our pipelines

delete them

we don't really care about new kernels, or rebooting those VMs, but we care about updated userland, and this prevents it

### DIFF
--- a/roles/vagrant_workarounds/tasks/main.yml
+++ b/roles/vagrant_workarounds/tasks/main.yml
@@ -7,3 +7,9 @@
     vtype: boolean
   when:
     - ansible_os_family == 'Debian'
+
+# workaround for https://github.com/lavabit/robox/issues/294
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I think this is an issue in https://app.vagrantup.com/generic/boxes/ubuntu2004/versions/4.3.12 and wonder if we should instead pin to 4.3.10 for now.

If that is indeed a regression, then [create an issue](https://github.com/lavabit/robox/issues)?

evgeni:
Opened https://github.com/lavabit/robox/issues/294 and referenced it in the role and the commit message

ekohl (approved):

PR #1758, opened 2024-02-06
Fixing some plays/conf/readme in order to deploy from MAC OS

waldirio:


playbooks/foreman.yml
```diff
@@ -1,4 +1,4 @@
-- hosts: all
+- hosts: " {{ group_name | default('all') }}"
```
waldirio, line 1: Hello,

I'm thinking of updating all the plays, like this. with this change, we should be able to set the hosts during the runtime, and if we keep this empty, no problem, it will be as is at this moment.

I can submit this request for all the plays in another PR, if you believe it to be interesting.

Thank you!
Waldirio

README.md
```diff
@@ -131,9 +131,26 @@ In case using Vagrant is not desired, ansible playbooks and roles from this repo
...
+```
+
+Now, you can share the ssh key with the server
+```sh
+ssh-copy-id vagrant@test.example.com
+```
+
 in forklift checkout
 ```sh
 echo -e "[devel]\ntest.example.com" > inventories/local_inventory
```
waldirio, line 156: Hi,

Here, there is no devel.yml, I'm not sure if this one was around before or not, but I didn't change it. I'll do a double check on all the plays on this folder. Just FYI.

Thank you!
Waldirio

README.md
```diff
@@ -131,9 +131,26 @@ In case using Vagrant is not desired, ansible playbooks and roles from this repo
 on test.example.com machine, where the dev env should be deployed
 ```sh
 useradd vagrant
+echo vagrant:redhat | chpasswd
 echo "vagrant	ALL=(ALL)	NOPASSWD: ALL" >> /etc/sudoers.d/vagrant
 ```
 
+Here we can see the initial configuration, when creating the python virtual environment and preparing the setup (MACOS Users or Regular Linux with no ansible)
```
ehelms, line 138: What if we introduce this virtual env method as it's own sub-section within this section to help users understand that a virtual env is an option?

playbooks/foreman.yml
ehelms: I don't follow entirely... could you provide an example command as a follow up comment?

waldirio: Hello @ehelms 

Sure, in the forklift example, we can see the command below
```
echo -e "[devel]\ntest.example.com" > inventories/local_inventory
```

This will include the `test.example.com` in the local_inventory, and in the `ansible.cfg`, we are reading pointing inventory to the directory `inventory = $PWD/inventories/`, which means, this will read all the files that we have created.

Assuming that I have already my devel env pointing to `test.example.com` and I would like to create a second server, now, using katello, for instance. I should be able to create a second entry on the `inventories/local_inventory` file and point the installation just for that server.

```
echo -e "[katello]\nnewserver.example.com" >> inventories/local_inventory
```

Now, the new group will be `katello`, but with the current configuration, once I call the play, it will be executed against all the servers on this inventory file/folder. With this change, we can run against all the servers, if we omit the extra vars, or we can act against that specific group, in this case, `katello`.

Basically, if we pass the group_name as `-e group_name=katello`, this will call only that group, if we leave it, this will use the standard wich will be all servers in the inventory.

Please, let me know if it's clear now.

Thank you!
Waldirio

README.md
waldirio: Sounds like a plan, let me do it.

Thank you!
Waldirio

waldirio: @ehelms please, let me know what you think.

Thank you!
Waldirio

playbooks/foreman.yml
ehelms: I thought the same could be achieved through native ansible method:

```
ansible-playbook playbooks/foreman.yml -l host1
ansible-playbook playbooks/foreman.yml -l katello
```

https://docs.ansible.com/ansible/latest/inventory_guide/intro_patterns.html#patterns-and-ansible-playbook-flags

README.md
```diff
@@ -131,9 +132,34 @@ In case using Vagrant is not desired, ansible playbooks and roles from this repo
 on test.example.com machine, where the dev env should be deployed
 ```sh
 useradd vagrant
+echo vagrant:redhat | chpasswd
```
ehelms, line 135: Why is this needed?

README.md
```diff
@@ -131,9 +132,34 @@ In case using Vagrant is not desired, ansible playbooks and roles from this repo
...
+#### Python Venv for MacOS or Non Ansible Linux
+
+Here we can see the initial configuration, when creating the python virtual environment and preparing the setup (MACOS Users or Regular Linux with no ansible)
+```sh
+/usr/local/bin/python3-intel64 -m venv ~/.venv/forklift
+source ~/.venv/forklift/bin/activate
+pip install --upgrade pip
+pip install ansible
+```
+Note. Above, you can see `python3-intel64`. This is very specific for MacOS M1 or M2 with arm processors, just to keep the compatibility with x86. If you have a regular x86, feel free to use `python3` and it should be enough.
```
ehelms, line 148: I would invert this relationship so the example is the standard Linux example and then include here `Note for MacOS M1 or M2 users: Replace `python3` with `python3-intel64`.

README.md
```diff
@@ -131,9 +132,34 @@ In case using Vagrant is not desired, ansible playbooks and roles from this repo
...
+git clone https://github.com/theforeman/forklift.git
+cd forklift
+ansible-galaxy collection install -r requirements.yml
+```
+
+Now, you can share the ssh key with the server
+```sh
+ssh-copy-id vagrant@test.example.com
+```
+
```
ehelms, line 162: Is this a repeat of previous steps in our documentation that is unspecific to the virtual env method?

playbooks/foreman.yml
waldirio: @ehelms that's perfect! Removing this additional work!

Thank you!

README.md
waldirio: I believe vagrant will deploy some key or information on the VM, in my case, I'm building a machine externally, so, once there are a bunch of plays using vagrant, I believe that keeping the user and setting the auth should be great, once I need to copy the ssh key to it. Anyway, I'm still testing the forklift with this env, passing --user root seems to work fine, but I'm not sure if we really need the vagrant user ... I believe yes.

Ps.: The ansible.cfg is using vagrant user as remote, so, not sure if we will connect as vagrant or if the play will connect as root and after that, will switch to the vagrant user.

Please, let me know if you are familiar with this process on forklift.

Thank you!
Waldirio

README.md
waldirio: updated.

README.md
waldirio: ok, let's check how we can do this better. In the requirements, we can see
```
Vagrant - 2.2+ - Both the VirtualBox and Libvirt providers are tested
Ansible - 2.12+
[Vagrant Libvirt provider plugin](https://github.com/vagrant-libvirt/vagrant-libvirt) (if using Libvirt)
Virtualization enabled in BIOS
```

For a regular fedora, we can install the packages and it will work with no issues. As a mac user, trying to test this environment and set the best practices, it's not standard ansible via package, as well vagrant. Once virtualization it's not something that will works in a short term for this kind of processor, what I'm doing is, using the forklift to prepare the basic installation, using the plays that are already around, and installing the minimal as possible.

To achieve that, we can use virtual env, python3 is already standard on mac, that specific binary will keep the compatibility with x86 python modules, then internally we can just install ansible anv voila, we can proceed using the plays.

All the plays that we have atm will work with no problem? No idea, but I'll test! 😉 

With that said, I'm recommending the venv just for the section below, for everyone else with fedora/rhel, it should not be a big deal, they could install the rpms and everything should be fine.
---
#### Python Venv for MacOS or Non Ansible Linux
---

Let me know what you think.
Waldirio

README.md
```diff
@@ -131,9 +132,34 @@ In case using Vagrant is not desired, ansible playbooks and roles from this repo
...
+#### Python Venv for MacOS or Non Ansible Linux
+
+Here we can see the initial configuration, when creating the python virtual environment and preparing the setup (MACOS Users or Regular Linux with no ansible)
+```sh
+/usr/local/bin/python3 -m venv ~/.venv/forklift
+source ~/.venv/forklift/bin/activate
+pip install --upgrade pip
+pip install ansible
+```
+Note. Above, you can see `python3`, this is the standard for any x86 server running Linux. If you are using MacOS M1 or M2 with arm processors, then you should use `/usr/local/bin/python3-intel64` instead, just for the sake of compatibility with x86.
```
ekohl, line 148: Do we care if it's compatible with x86? Does Ansible rely on any native libraries that aren't available on ARM?

ekohl (commented):
Like @ehelms, I think the structure isn't the best. You put it between two parts that are about using non-vagrant machines. IMHO they are two separate concerns. I can imagine that on a mac you need to install ansible using a venv, but still use Vagrant (with VirtualBox).

waldirio:
Hello all,

I'm already working on it, @ianballou is helping a lot, and soon we will have some news. I believe we can keep this PR on hold, and then we can return to it.

and @ekohl answering your point, in my case, no vagrant at all, just forklift and centos8 stream server, and moving on to setup the whole environment.

I'll keep you all posted.

Thank you!
Waldirio

PR #1760, opened 2024-02-08
Use ansible_user_dir in customize_home role

wbclark:
This was only working when run under the vagrant user


roles/customize_home/tasks/main.yml
```diff
@@ -20,6 +20,9 @@
   when: customize_home_git_repo is defined
 
 - include_tasks: fix_env.yml
+  args:
+    apply:
+      become: false
```
wbclark, line 25: Hmmm, when I tried it without `become: false` it tried to use `/root/` but with `become: false` it fails due to permission denied.

I will have to try `/home/{{ ansible_user }}/foreman/.env` instead

roles/customize_home/tasks/fix_env.yml
```diff
@@ -1,11 +1,11 @@
 - name: Check that the .env exists
   stat:
-    path: /home/vagrant/foreman/.env
+    path: "/home/{{ ansible_user }}/foreman/.env"
```
evgeni, line 3: How about 
```suggestion
    path: "{{ ansible_user_dir }}/foreman/.env"
```
(from https://github.com/ansible/ansible/blob/stable-2.16/lib/ansible/module_utils/facts/system/user.py)

This should be present in any run that has facts enabled, but has the benefit of not guessing the home dir as `/home/$USERNAME` but actually looking it up from the user database.

wbclark: Thanks, I'll give it a test

wbclark: This only works if I manually add ansible_user_dir to the variables for the ansible provisioner in the box yaml config... otherwise (I guess because `become: true` in `playbooks/katello_devel.yml`) it still tries to use /root

I also realized that I was already overriding the value for ansible_user in my box, so that is likely the only reason my other approach was working (and I must have encountered something like this before already)

wbclark: I suppose I could just add a role variable here with a default value that works for most setups to avoid this annoying behavior with `become`

evgeni: ah, yeah… `become` certainly kills that.

role variable sounds good to me then!

ekohl (approved):

roles/customize_home/tasks/fix_env.yml
evgeni: Another idea, can we force `become: false` when running this role?

ekohl (changes requested):
Actually, the commit message no longer matches the content of the commit.

wbclark:
> Actually, the commit message no longer matches the content of the commit.

I haven't gotten back implementing the role variable yet either. I was out sick, but will get to that today

roles/customize_home/tasks/fix_env.yml
wbclark: > Another idea, can we force `become: false` when running this role?

See the thread above this one that I marked as "resolved" :(

> but with `become: false` it fails due to permission denied.

wbclark: > Another idea, can we force `become: false` when running this role?

although I was using `become: false` for just the one task and not the whole role, and the user should have permission to modify the contents of their own home directory....

wbclark: OK, the issue appears to be that we have `become: true` in `playbooks/katello_devel.yml` and I wasn't able to override that

The good news is that it looks like `become` works for role dependencies in `meta/main.yml`, so I can remove it from `playbooks/katello_devel.yml` and add it back as needed per role dependency in `roles/katello_devel/meta/main.yml`

roles/katello_devel/meta/main.yml
```diff
@@ -2,6 +2,7 @@
 dependencies:
   - role: etc_hosts
```
evgeni, line 3: I'd expect this to also need become true?

evgeni (commented):
We should drop the scl roles, they would need become true also, but are unused.

roles/katello_devel/meta/main.yml
wbclark: I had the same thought, but `become: yes` is already specified at the task level within that role.

wbclark:
> We should drop the scl roles, they would need become true also, but are unused.

I pushed a change that removes the SCL roles from katello_devel

Two of these were now also unused in other roles and playbooks, so I removed those roles entirely.

roles/katello_devel/meta/main.yml
```diff
@@ -2,24 +2,20 @@
...
   - role: foreman_installer_devel_scenario
+    become: true
   - role: foreman_installer
+    become: true
     foreman_installer_scenario: katello-devel
     foreman_installer_additional_packages:
       - foreman-installer-katello
     foreman_installer_disable_system_checks: true
     foreman_installer_options_internal_use_only:
-      - "{{ '--katello-devel-scl-ruby=' +  ruby_scl_version if ansible_distribution_major_version == '7' else '' }}"
```
wbclark, line 22: (this was already removed in puppet-katello_devel)

wbclark:
Also, if you prefer the removal of unused SCL roles and change to centos8-stream in katello-katello devel pipelines in separate PRs, let me know and I'll make the change

evgeni (approved):
There are separate commits, so this looks good to me!

wbclark:
@ekohl would you be willing to remove the out of date "changes requested" flag please?

wbclark:
What sorcery is that! Didn't realize it was possible.

Also, I cannot merge in this repository -- would you please, @evgeni ?

evgeni:
I think this sorcery is limited to people with commit rights ;-)

(done)

wbclark:
TYVM!

Issue #1748, opened 2023-12-12
Point repositories to copr instead of koji


ianballou:
New dev boxes still point to koji which is out of date at this point.

PR #1761, opened 2024-02-09
Use copr repositories for katello devel boxes

wbclark:
Closes https://github.com/theforeman/forklift/issues/1748

### DIFF
--- a/roles/katello_repositories/tasks/staging_repos.yml
+++ b/roles/katello_repositories/tasks/staging_repos.yml
@@ -25,5 +25,4 @@
 - name: 'Add Candlepin repository'
   include_role:
     name: candlepin_repositories
-  when:
-    - candlepin_repositories_version is defined
+  when: candlepin_repositories_version is defined or katello_repositories_staging_source == "stagingyum"
--- a/roles/katello_devel/meta/main.yml
+++ b/roles/katello_devel/meta/main.yml
@@ -3,7 +3,9 @@ dependencies:
   - role: etc_hosts
   - role: foreman_server_repositories
     foreman_server_repositories_katello: true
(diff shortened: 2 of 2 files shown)


roles/katello_repositories/tasks/staging_repos.yml
```diff
@@ -25,5 +25,3 @@
 - name: 'Add Candlepin repository'
   include_role:
     name: candlepin_repositories
-  when:
-    - candlepin_repositories_version is defined
```
ehelms, line 29: I think this change would break pipelines that are still based on Koji.

wbclark: It seems odd that we have this check for candlepin repos but not for pulpcore repos. What am I missing?

I'd also like to avoid creating another value for candlepin_repositories_version that has to be maintained separately, if it can be avoided.

I'll attempt modifying this condition for now.

wbclark: I changed it to `when: candlepin_repositories_version is defined or katello_repositories_staging_source == "stagingyum"` to avoid including the separate candlepin repositories role when using Koji

ehelms: For devel, do you want the default or do you want nightly? e.g. This right now will enable the default which is '4.3'.

wbclark: Summarizing the offline conversation for the record:

We'll go with 4.3 for as default for now.

We'll plan to switch to nightly candlepin as default in the future once the katello integration tests (see: https://theforeman.org/plugins/katello/developers.html#upgrading-candlepin ) can be added to the pipeline.

ehelms (approved):

PR #1765, opened 2024-02-12
Fix stagingyum repositories URLs

wbclark:
https://stagingyum.theforeman.org/foreman/staging/el8/x86_64/ returns 404, it should be https://stagingyum.theforeman.org/foreman/nightly/el8/x86_64/

This fixes the issue for those who don't have in their box yaml: katello_repositories_version: nightly

### DIFF
--- a/roles/foreman_repositories/tasks/_stagingyum_staging_repos.yml
+++ b/roles/foreman_repositories/tasks/_stagingyum_staging_repos.yml
@@ -1,18 +1,18 @@
 ---
-- name: 'Foreman {{ foreman_repositories_version }} Staging repository'
+- name: 'Foreman Nightly Staging repository'
   yum_repository:
     name: foreman-staging
-    description: "Foreman {{ foreman_repositories_version }} Staging Repository"
-    baseurl: "https://stagingyum.theforeman.org/foreman/{{ foreman_repositories_version }}/el{{ ansible_distribution_major_version }}/x86_64/"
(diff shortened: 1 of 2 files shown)


evgeni (changes requested):
No, you gotta set the right version. "Staging" is not a version.

wbclark:
> No, you gotta set the right version. "Staging" is not a version.

it's not clear to me where that's coming from.

I had `katello_repositories_version: nightly` in my box yaml already so it just worked for me.

But that's already the default for the katello_repositories role: https://github.com/theforeman/forklift/blob/master/roles/katello_repositories/defaults/main.yml#L2

ianballou:
> CC: @ianballou this looks like it should have been working for you already using the defaults -- were you passing some extra variables to your box that were breaking it?

Very likely yes! However I got it to work with:

```
centos8-katello-devel-2:
  primary: true
  box: centos8-stream
  cpus: 2
  memory: 15000
  ansible:
    playbook: 'playbooks/katello_devel.yml'
    group: 'devel'
    variables:
      ssh_forward_agent: true
      foreman_installer_options:
        - "--foreman-proxy-content-enable-ostree=true"
      katello_repositories_environment: staging
```

evgeni:
Every release pipeline uses those repos to validate them before pushing them live ;)

wbclark:
Closing as this was not ultimately needed.

PR #1767, opened 2024-02-13
create an archive of the smoker report

evgeni:
fetching individual files is super slow, so let's transfer just one file!


evgeni:
jenkins change in https://github.com/theforeman/jenkins-jobs/pull/414

PR #1771, opened 2024-02-14
Drop test skipping on Pulp 3.39

ekohl:
The bug was fixed in Pulp 3.39.4. pulpcore-packaging currently contains that since [early january](https://github.com/theforeman/pulpcore-packaging/commit/cffbf381b2c862f589b36c32462ea6921d7fd315).

PR #1772, opened 2024-02-14
Un-skip Pulpcore 3.39 Katello bats tests

ianballou:
Tested on Katello 4.11.

I suppose I could remove the `tSkipIfPulp339` definition, but  maybe it'll be useful later.


ehelms:
Great minds think a like? https://github.com/theforeman/forklift/pull/1771

PR #1774, opened 2024-02-16
test Discovery on EL9 again

evgeni:


evgeni:
tested locally, works fine \o/

PR #1777, opened 2024-02-21
Use firefox with Smoker

ekohl:
In Chrome 121 there's a memory leak when using chromedriver. This is fixed with 122, but that's not available yet.

Untested right now.


ekohl:
Trying it out with https://github.com/tmiller02/ansible-role-geckodriver I see:
```
WARNING: Failed to gather log types: Message: HTTP method not allowed
```

I thought that was now implemented in Firefox.

ekohl:
For now I'm closing this since Chromium 122 landed. I'd still like to use Firefox by default, but that requires some work.

PR #1780, opened 2024-03-04
never install puppet 7.29.0

evgeni:


roles/vagrant_workarounds/tasks/main.yml
```diff
@@ -13,3 +13,24 @@
...
+      Package: puppet-agent
+      Pin: version 7.29.0*
+      Pin-Priority: -10
+  when:
+    - ansible_os_family == 'Debian'
+
+# workaround for https://github.com/puppetlabs/puppet/issues/9268
+- name: never install Puppet 7.29.0 on EL
+  community.general.ini_file:
+    path: /etc/dnf/dnf.conf
```
evgeni, line 31: we place it into the main `dnf.conf` as when this role runs `/etc/yum.repos.d/puppet7-release.repo` doesn't exist yet, plus, even if it'd exist, there is a chance that it gets overwritten during one of our "upgrade packages" steps and we really do not want that.

ekohl (approved):

PR #1783, opened 2024-03-05
Add RH Cloud and virt-who Configure to Katello Devel Role

chris1984:


roles/katello_devel/meta/main.yml
```diff
@@ -1,4 +1,4 @@
----
+ca---
```
ianballou, line 1: :canada: 

Typo? :)

chris1984: Fat fingers, let me fix it

chris1984: @ianballou fixed

chris1984:
@ehelms can this get a merge, simple one :)

PR #1784, opened 2024-03-06
Use fixtures.pulpproject.org for bats tests

ehelms:
The fixtures at fedorapeople.org cannot be accessed when running RHEL 9 in FIPS mode. This also has the nice affect of using the fixtures Pulp maintains rather than a variety of artifacts from multiple places that are not owned by a group.


ekohl:
@parthaa @ianballou mind taking a look?

@ehelms I noticed our webserver also doesn't support TLS 1.3 (because it's EL7) so do we not test syncing client repos or is Ruby/Python code ignoring FIPS?

ehelms:
> @ehelms I noticed our webserver also doesn't support TLS 1.3 (because it's EL7) so do we not test syncing client repos or is Ruby/Python code ignoring FIPS?

To my knowledge they ignore it, which is somewhat by design as we are not guaranteeing the application layer be FIPS compliant.

ianballou:
@ehelms if  you've tested this change against BATS I'm cooling acking it.

ehelms:
> @ehelms if you've tested this change against BATS I'm cool acking it.

I ran it locally, give me the green check mark and we'll see what our pipelines think.

ekohl (approved):
Good enough for me.

PR #1785, opened 2024-03-07
Add EL9 to FIPS role

ehelms:


evgeni (approved):
This is identical to el8, so we could also drop 7 and merge everything in main yml again

ehelms:
> This is identical to el8, so we could also drop 7 and merge everything in main yml again

Done

evgeni (approved):

PR #1788, opened 2024-03-12
workaround broken Tomcat on EL8

evgeni:


Odilhao (approved):
Never say Never! :notes:

PR #1795, opened 2024-03-19
pass the pipeline version to beaker

evgeni:
our nightly repos are unsigned and if you pass version=nightly to pulpcore::repo it will disable gpg checking, otherwise things don't install


Odilhao (approved):

ekohl (approved):
We'll also need to fix the GPG key. This will work for nightly because of https://github.com/theforeman/puppet-pulpcore/blob/7f48ca8566429f9b0e3db448e163be69e0bd0fb1/manifests/repo.pp#L21 but it'll break when we branch. https://github.com/theforeman/puppet-pulpcore/blob/7f48ca8566429f9b0e3db448e163be69e0bd0fb1/spec/setup_acceptance_node.pp#L1-L4 doesn't have a parameter for the GPG key, so that first needs a patch.

PR #1796, opened 2024-03-21
Use Pulpcore Nightly with Katello nightly pipelines

Odilhao:


ekohl (approved):

vagrant/config/versions.yaml
```diff
@@ -90,7 +90,7 @@ installers:
   - foreman: 'nightly'
     katello: 'nightly'
     candlepin: '4.3'
```
ekohl, line 92: @ehelms you want to also change this to nightly?

ehelms: I do not believe we arrived at a decision on the workflow, I think we were leaning towards yes but we needed to document it first.

PR #1799, opened 2024-03-22
Drop pulp-installer pipelines and switch Pulp 3.39 to puppet-pulpcore

Odilhao:


pipelines/pulpcore/02-install.yml
```diff
@@ -62,4 +11,4 @@
     - role: beaker
       when:
         - pipeline_version is defined
-        - pipeline_version == 'nightly' or pipeline_version is version('3.40', '>=')
+        - pipeline_version == 'nightly' or pipeline_version is version('3.39', '>=')
```
ekohl, line 14: Perhaps drop the whole `when` block?

pipelines/pulpcore/03-tests.yml
```diff
@@ -12,26 +12,10 @@
         tasks_from: test
       when:
         - pipeline_version is defined
-        - pipeline_version == 'nightly' or pipeline_version is version('3.40', '>=' )
+        - pipeline_version == 'nightly' or pipeline_version is version('3.39', '>=' )
```
ekohl, line 15: Same here

ehelms (approved):

ekohl (changes requested):

Odilhao:
> I wanted to also ask to clean up https://github.com/theforeman/forklift/blob/master/pipelines/vars/forklift_pulpcore.yml because it's probably largely unused now.

Just created https://github.com/theforeman/forklift/pull/1801 to address the `when` that we don't need anymore and the unused variables.

PR #1811, opened 2024-04-12
Update pulpcore to nightlies

sjha4:


ekohl (approved):
This matches https://github.com/theforeman/forklift/blob/94562187fb985b30063ea44b01d937e421f154f6/vagrant/config/versions.yaml#L93

PR #1816, opened 2024-04-29
Wait for hammer ping to work in Puppet tests

evgeni:


bats/fb-test-puppet.bats
```diff
@@ -33,7 +33,7 @@ fi
 
   foreman-installer --enable-foreman-plugin-puppet --enable-foreman-cli-puppet --foreman-proxy-puppet true --foreman-proxy-puppetca true --enable-puppet --puppet-server true "${cert_options[@]}"
   # Force hammer to reload the apidoc cache - https://projects.theforeman.org/issues/28283
```
evgeni, line 35: This is fixed in 3.3, but I'd like to keep compatibility with that for a while.

ekohl: Should the comment then be moved to `tHammerPing` where it actually sends `--reload-cache`?

evgeni: Maybe. But the reload is only really *needed* here, so idk.

ekohl: Perhaps `tHammerPing` should accept a `$1` then?

evgeni: Hmm, e247256378e640dd44d1ca425733b58222f087d4 says it was to aid upgrades in core (regardless of puppet).
But then again https://github.com/theforeman/forklift/pull/1390 doesn't really tell much either.

ekohl: I vaguely remember it being some Hammer bug where Hammer didn't notice it needed to download a new cache after Foreman was upgraded from n-1 to n. https://projects.theforeman.org/issues/28283 (as linked in the comment) also describes that.

ekohl (approved):

PR #1819, opened 2024-05-08
add virtualbox links for centos

evgeni:


Odilhao (approved):
Virtualbox :thinking:

PR #1820, opened 2024-05-14
Add a config example for centos9-stream katello devel

ianballou:
Since centos 8 stream is going EOL.

### DIFF
--- a/vagrant/boxes.d/99-local.yaml.example
+++ b/vagrant/boxes.d/99-local.yaml.example
@@ -1,7 +1,23 @@
 ---
-centos8-katello-devel:
+centos9-katello-devel:
   primary: true
-  box: centos8-stream
+  box: centos9-stream
+  ansible:
+    playbook: 'playbooks/katello_devel.yml'
+    group: 'devel'
+    variables:
+      ssh_forward_agent: true
+      foreman_devel_github_push_ssh: True
+      katello_devel_github_username: <REPLACE ME>
+      foreman_repositories_environment: staging
+      katello_repositories_environment: staging
+      foreman_installer_options:
+        - "--foreman-proxy-content-enable-ostree=true"
+        - "--katello-devel-modulestream-nodejs=18"
+
+almalinux8-katello-devel:
+  primary: true
+  box: almalinux8
   ansible:
     playbook: 'playbooks/katello_devel.yml'
     group: 'devel'
--- a/vagrant/boxes.d/99-local.yaml.example
+++ b/vagrant/boxes.d/99-local.yaml.example
@@ -11,6 +11,8 @@ centos9-katello-devel:
       katello_devel_github_username: <REPLACE ME>
       foreman_repositories_environment: staging
       katello_repositories_environment: staging
+      foreman_repositories_version: nightly
+      katello_repositories_version: nightly
       foreman_installer_options:
         - "--foreman-proxy-content-enable-ostree=true"
         - "--katello-devel-modulestream-nodejs=18"
@@ -31,27 +33,27 @@ almalinux8-katello-devel:
         - "--foreman-proxy-content-enable-ostree=true"
         - "--katello-devel-modulestream-nodejs=12"
 
-centos8-luna-demo:
-  box: centos8-stream
+almalinux8-luna-demo:
+  box: almalinux8
   memory: 10240
   ansible:
     playbook: playbooks/luna_demo_environment.yml
 
-centos8-hammer-devel:
-  box: centos8-stream
+almalinux8-hammer-devel:
+  box: almalinux8
   memory: 512
   ansible:
     playbook: 'playbooks/hammer_devel.yml'
     group: 'hammer-devel'
     variables: # defaults in roles/hammer_devel/defaults/main.yml
-      #hammer_devel_server: centos8-katello-devel.custom-domain.example.com
+      #hammer_devel_server: almalinux8-katello-devel.custom-domain.example.com
 
-centos8-proxy-devel:
-    box: centos8-stream
+almalinux8-proxy-devel:
+    box: almalinux8
     ansible:
       playbook: 'playbooks/foreman_proxy_content_dev.yml'
       group: 'foreman-proxy-content'
-      server: 'centos8-katello-devel'
+      server: 'almalinux8-katello-devel'
 
 centos7-katello-client:
   box: centos7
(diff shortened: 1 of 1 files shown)


ianballou:
@parthaa does this match with your strategy?

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -1,4 +1,20 @@
...
+      ssh_forward_agent: true
+      foreman_devel_github_push_ssh: True
+      katello_devel_github_username: <REPLACE ME>
+      foreman_repositories_environment: staging
+      katello_repositories_environment: staging
+      foreman_installer_options:
+        - "--foreman-proxy-content-enable-ostree=true"
+        - "--katello-devel-modulestream-nodejs=18"
+
 centos8-katello-devel:
```
evgeni, line 18: I think we should delete this box, or move it to Alma8, as in the current state it won't work anymore.

ianballou: I moved it to Alma.

ekohl (commented):
I've asked @archanaserver to look at this as well. The file is filled with centos 8 references that are broken.

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -1,7 +1,23 @@
...
+    group: 'devel'
+    variables:
+      ssh_forward_agent: true
+      foreman_devel_github_push_ssh: True
+      katello_devel_github_username: <REPLACE ME>
+      foreman_repositories_environment: staging
+      katello_repositories_environment: staging
+      foreman_installer_options:
+        - "--foreman-proxy-content-enable-ostree=true"
+        - "--katello-devel-modulestream-nodejs=18"
```
wbclark, line 16: Musing: we _could_ remove this with a merge of https://github.com/theforeman/puppet-katello_devel/pull/300 , which (among other fixes) makes NodeJS 18 the default, but I lean towards keeping it here anyway for the purpose of discoverability.

wbclark:
So far, it seems that I am the only person who has run into this issue, but just in case anyone else faces ssh authentication problems due to EL9 images not supporting RSA keypairs:

First, try forcing a re-download of the latest CentOS 9 Stream base image; if that fails for whatever reason, patching your Vagrant with the following commits will fix it:

- https://github.com/hashicorp/vagrant/commit/b40f6e5fdae1113518e9c92f96dfcd364b646ff9
- https://github.com/hashicorp/vagrant/commit/67562588c97fb6a29e017a36aaec3240e4ada85a
- https://github.com/hashicorp/vagrant/commit/cebfb7a63b0d7306bd65585cefbcb61730385f31
- https://github.com/hashicorp/vagrant/commit/380afe5fac713d4ad930c147940beb9237c18379
- https://github.com/hashicorp/vagrant/commit/e0dbbcc04ce626c851eeba122ca45c66bfbf1342
- https://github.com/hashicorp/vagrant/commit/c8fc8b3ad2054110ea1f06d678abf58a98ff887f

ianballou:
I added a commit that removes the rest of the centos8-stream references.

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -15,50 +31,50 @@ centos8-katello-devel:
...
-centos8-katello-client:
-  box: centos8-stream
+almalinux8-katello-client:
+  box: almalinux8
   ansible:
     playbook: 'playbooks/katello_client.yml'
     group: 'client'
     variables:
-      katello_client_server: 'centos8-katello-devel'
+      katello_client_server: 'almalinux8-katello-devel'
```
jeremylenz, line 74: We could remove this. As far as I know this playbook installs katello-agent on a host.

wbclark: That playbook includes, among other things, this role: https://github.com/theforeman/forklift/blob/master/roles/katello_client/tasks/main.yml

It does a bit more than just installing katello-agent so probably best to leave it alone

jeremylenz: I think it's outside the scope here probably, but I think that box and role/playbook should be removed. But we can have that discussion separately..

wbclark (approved):

jeremylenz (approved):

wbclark:
@parthaa @jeremylenz the relevant puppet-katello_devel PRs are
- https://github.com/theforeman/puppet-katello_devel/pull/300
- https://github.com/theforeman/puppet-katello_devel/pull/302

ianballou:
I'll wait a little longer for @archanaserver due to @ekohl 's message above.

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -69,24 +87,24 @@ centos7-foreman-bats-ci:
     playbook: 'playbooks/bats_pipeline_foreman_nightly.yml'
     group: 'bats'
 
-centos8-dynflow-devel:
-  box: centos8-stream
+almalinux8-dynflow-devel:
+  box: almalinux8
   memory: 512
```
ekohl, line 92: I really doubt an EL8 box will function with 512 MB RAM, but probably outside of this PR's scope

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -15,50 +33,50 @@ centos8-katello-devel:
...
   ansible:
     playbook: 'playbooks/hammer_devel.yml'
     group: 'hammer-devel'
     variables: # defaults in roles/hammer_devel/defaults/main.yml
-      #hammer_devel_server: centos8-katello-devel.custom-domain.example.com
+      #hammer_devel_server: almalinux8-katello-devel.custom-domain.example.com
 
 katello-remote-execution:
```
ekohl, line 51: I don't think this box makes any sense anymore. katello has a hard dependency on REX now

vagrant/boxes.d/99-local.yaml.example
ekohl: Perhaps comment it out as an example?

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -15,50 +33,50 @@ centos8-katello-devel:
...
   ansible:
     playbook: 'playbooks/katello_client.yml'
     group: 'client'
     variables:
-      katello_client_server: 'centos8-katello-devel'
+      katello_client_server: 'almalinux8-katello-devel'
 
-centos8-katello-bats-ci:
-  box: centos8-stream
+almalinux8-katello-bats-ci:
```
ekohl, line 78: Related, but not blocking for this PR: below is the CentOS 7 bats box that can't work anymore.

ekohl (changes requested):
I opened https://github.com/theforeman/forklift/pull/1827 to drop redundant boxes. It will give a merge conflict, but I'd prefer to first clean them up rather than fix them up and then remove them.

ianballou:
@ekohl I think I addressed all your comments

vagrant/boxes.d/99-local.yaml.example
```diff
@@ -1,7 +1,25 @@
...
+      ssh_forward_agent: true
+      foreman_devel_github_push_ssh: True
+      katello_devel_github_username: <REPLACE ME>
+      foreman_repositories_environment: staging
+      katello_repositories_environment: staging
+      foreman_repositories_version: nightly
+      katello_repositories_version: nightly
+      foreman_installer_options:
+        - "--foreman-proxy-content-enable-ostree=true"
+#        - "--katello-devel-modulestream-nodejs=18"
```
ekohl, line 18: I think commenting in this style is easier to read. Is that just me?
```suggestion
        #- "--katello-devel-modulestream-nodejs=18"
```

ianballou: I feel like your way is nicer to look at, but having it on the right side is quicker to delete in a terminal editor (only have to go down instead of down and right). 

The method I chose was pretty much the result of a coin flip :)

jeremylenz: VSCode has an easy keyboard shortcut to uncomment the current line, just saying..

ekohl: I never thought about it because in vim `^` navigates to the first non-whitespace character of a line so `^x` deletes the comment.

ianballou: That's a good tip. I moved the comment.

ianballou:
Reviews have gone quiet and I got some acks, so merging now. The ansible-lint issue there appears to be unrelated.

PR #1800, opened 2024-03-22
Drop koji support

ekohl:
Now that Foreman 3.8 is EOL, nothing is built with Koji anymore and it can be retired. This means dropping the koji staging repositories, but also including certain scratch builds.

It also cleans up some pulp stuff and includes https://github.com/theforeman/forklift/pull/1797

PR #1821, opened 2024-05-15
remove *_repositories_staging_source variables from katello_devel

wbclark:
These are no longer needed since https://github.com/theforeman/forklift/pull/1800
See also: https://github.com/theforeman/forklift/pull/1761/files#r1484290004


evgeni (approved):

wbclark:
Thanks for the quick review @evgeni 

I don't have merge access in this repository -- would you do the honors?

Issue #1824, opened 2024-05-29
Update Forklift to Alma


ehelms:
- [x] Devel boxes
- [x] Remove CentOS 8 Stream boxes
- [x] Update smoker box to use CS9

ekohl:
https://github.com/theforeman/forklift/pull/1820

archanaserver:
- [ ] https://github.com/theforeman/forklift/pull/1832
- [ ] https://github.com/theforeman/forklift/pull/1826

archanaserver:
We can close this issue now and mark all tasks completed (I don't have permission to do so).

PR #1825, opened 2024-06-05
run Smoker on CS9

evgeni:


evgeni:
Refs, but not closes, https://github.com/theforeman/forklift/issues/1824

ekohl (approved):

PR #1832, opened 2024-07-12
centos7 and centos8 name update

archanaserver:
Ref: https://github.com/theforeman/forklift/pull/1826#pullrequestreview-2145729290

PR #1826, opened 2024-06-06
Drop CentOS 8 and CentOS 7 boxes

archanaserver:
https://github.com/theforeman/forklift/issues/1824

- [x] https://github.com/theforeman/forklift/pull/1841
- [x] https://github.com/theforeman/forklift/pull/1842
- [x] https://github.com/theforeman/forklift/pull/1843
- [x] https://github.com/theforeman/forklift/pull/1844
- [x] https://github.com/theforeman/forklift/pull/1832

PR #1834, opened 2024-07-19
Switch stable box from alma to centos9 in 99-local

chris1984:
We are using Centos9 for the stable boxes in vagrant cloud, so using alma would cause it to break. 

Switching the name, we also are only publishing el9 to save resources and overhead.

### DIFF
--- a/vagrant/boxes.d/99-local.yaml.example
+++ b/vagrant/boxes.d/99-local.yaml.example
@@ -92,15 +92,15 @@ almalinux8-dynflow-devel:
     variables:
       dynflow_devel_github_fork_remote_name: 'origin'
 
-almalinux8-katello-nightly-stable:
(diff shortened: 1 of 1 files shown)


ekohl (commented):
In https://github.com/theforeman/forklift/pull/1826 and https://github.com/theforeman/forklift/pull/1832 @archanaserver is also updating docs. Perhaps you could collaborate on making that consistent.

nadjaheitmann:
Tested your PR and box starts up perfectly fine :)

ekohl:
@wbclark that would be fine by me because I think we should get into a habit of updating the docs as part of the code change.

wbclark:
I agree with @ekohl, this practice makes it easier to navigate the git history as well. What do you say, @chris1984 ?

PR #1835, opened 2024-07-20
Clone test gems to katello.local.rb to allow local katello testing

qcjames53:
This PR addresses [SAT-26017](https://issues.redhat.com/browse/SAT-26017), adding katello test gems to foreman's katello.local.rb on vagrant box provision if certain conditions are met.

**Considerations:**
As of now, `customize_home` is only called for katello-related boxes. I added checks for both the source dir and destination file existing just in case. They shouldn't kick in unless we decide to `customize_home` for non-katello boxes as well, but I figured it's better to be safe.

**Testing Steps:**

1. In the forklift dir, create a new katello-devel box and provision it. On completion, check the contents of `~/foreman/bundler.d/katello.local.rb`. It should a 'test' group similar to the following:
```
group :test do
  gem "vcr", "~> 6.1"
  gem 'theforeman-rubocop', '~> 0.0.6', require: false
end
```
2. Re-run the provisioning on this box without destroying. `~/foreman/bundler.d/katello.local.rb` should contain only one 'test' group.
3. Create a katello-devel-stable box. `~/foreman/bundler.d/katello.local.rb` should still contain a 'test' group.

### DIFF
--- a/roles/customize_home/tasks/main.yml
+++ b/roles/customize_home/tasks/main.yml
@@ -11,6 +11,43 @@
     dest: "{{ ansible_env.HOME }}"
   when: custom_home.stat.isdir is defined and custom_home.stat.isdir
 
+- name: Check if ~/katello/gemfile.d directory exists
+  ansible.builtin.stat:
+    path: "{{ ansible_user_dir }}/katello/gemfile.d"
+  register: katello_gemfile_dir
+
+- name: Check if ~/foreman/bundler.d/katello.local.rb file exists
+  ansible.builtin.stat:
+    path: "{{ ansible_user_dir }}/foreman/bundler.d/katello.local.rb"
+  register: katello_local_rb_file
+
+- name: Find all .rb filepaths
+  ansible.builtin.find:
+    paths: "{{ ansible_user_dir }}/katello/gemfile.d"
+    patterns: '*.rb'
+  register: found_files
+  when: katello_gemfile_dir.stat.exists and katello_local_rb_file.stat.exists
+
+- name: Read the contents of all .rb files
+  ansible.builtin.slurp:
+    src: "{{ item.path }}"
+  loop: "{{ found_files.files }}"
+  register: found_file_contents
+  when: katello_gemfile_dir.stat.exists and katello_local_rb_file.stat.exists
+
+- name: Ensure encoding is base 64
+  ansible.builtin.fail:
+    msg: "File {{ item.item.path }} is not base64 encoded"
+  loop: "{{ found_file_contents.results }}"
+  when: katello_gemfile_dir.stat.exists and katello_local_rb_file.stat.exists and item.encoding != "base64"
+
+- name: Append .rb file contents to katello.local.rb so we can run local tests
+  ansible.builtin.blockinfile:
+    path: "{{ ansible_user_dir }}/foreman/bundler.d/katello.local.rb"
+    block: "{{ item.content | b64decode }}"
+  loop: "{{ found_file_contents.results }}"
+  when: katello_gemfile_dir.stat.exists and katello_local_rb_file.stat.exists
+
 - name: Remove .gitkeep file copied over from custom directory
   file:
     path: "{{ ansible_env.HOME }}/.gitkeep/"


roles/customize_home/tasks/main.yml
```diff
@@ -11,6 +11,21 @@
...
   when: custom_home.stat.isdir is defined and custom_home.stat.isdir
 
+- name: Check if ~/katello/gemfile.d directory exists
+  ansible.builtin.stat:
+    path: ~/katello/gemfile.d
+  register: katello_gemfile_dir
+
+- name: Check if ~/foreman/bundler.d/katello.local.rb file exists
+  ansible.builtin.stat:
+    path: ~/foreman/bundler.d/katello.local.rb
```
wbclark, line 21: (Same idea as above)

wbclark (commented):
Hey @qcjames53 thanks for getting the ball rolling on this!

This is a good first attempt, but it's not idempotent, right? This is one of the reasons we normally try to avoid the shell module except as a last resort.

Doing this the "proper ansible way" seems surprisingly hard, I guess because it's a less common use case than getting some file content from our (control) machine and copying it to the target machine, but I think it should be possible using:

- https://docs.ansible.com/ansible/latest/collections/ansible/builtin/find_module.html
- https://docs.ansible.com/ansible/latest/collections/ansible/builtin/slurp_module.html
- https://docs.ansible.com/ansible/latest/collections/ansible/builtin/blockinfile_module.html

Could you give it a shot this way? I don't want you to get bogged down with it for too long, so let me know if I can help unblock you on any of the specifics. :)

qcjames53:
Hi William, thanks for the review! I rewrote the tasks following your suggestions; it's certainly a lot cleaner now. I didn't end up testing the stable box for this specific push but I can confirm it functions ok for creating regular katello-devel boxes.

roles/customize_home/tasks/main.yml
```diff
@@ -11,6 +11,21 @@
     dest: "{{ ansible_env.HOME }}"
   when: custom_home.stat.isdir is defined and custom_home.stat.isdir
 
+- name: Check if ~/katello/gemfile.d directory exists
+  ansible.builtin.stat:
+    path: ~/katello/gemfile.d
```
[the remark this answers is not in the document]
qcjames53: Implemented, thank you :)

roles/customize_home/tasks/main.yml
qcjames53: Thanks :)

ekohl (changes requested):
Just another thought: you can also read the files on the fly. I suspect you can drop in this file to `bundler.d/katello-test.local.rb`:
```ruby
Dir['/path/to/katello/gemfile.d/*.rb'].each do |file|
  eval_gemfile(file)
end
```

I think that results in way easier Ansible code while also making sure the files are always in sync (because there is no copying) even when you do `git pull`.

This is effectively what already happens:
https://github.com/theforeman/foreman/blob/75ac6568187719d114335381bfaf32d3dd266572/Gemfile#L57-L59
https://github.com/theforeman/smart-proxy/blob/d287e3cc01698852c9f6ff85eac246263d01a5b3/Gemfile#L10-L12

You can note there's a difference. The hope was that `eval_gemfile` was supported by dependabot, but it's not when used with a variable.

wbclark:
Hey @qcjames53 , your new implementation is indeed what I had in mind, but I think @ekohl has a good point -- it would be much better still to have ruby read the files on the fly instead of doing it once during provisioning, so even if there are changes we can just run a `bundle install` without having to re-run the whole provisioning.

It's a good reminder to both of us to think carefully about the problem before accepting the proposed solution in a ticket.

wbclark:
Another thought is that the customize_home role exists to customize the user's home directory and is probably not the best fitting place for this functionality to live. What to do about that?

One idea is that while the katello_devel role is currently a meta role that functions only through its dependencies:

```
$ tree roles/katello_devel/
roles/katello_devel/
└── meta
    └── main.yml

2 directories, 1 file
```

It doesn't have to be, and we _could_ give it this structure:

```
$ tree roles/katello_devel/
roles/katello_devel/
├── files
│   └── katello-test.local.rb
├── meta
│   └── main.yml
└── tasks
    └── main.yml

4 directories, 3 files
```

But, I think doing it that way is still not quite right, because it would require doing a second `bundle install` when we already do that at an earlier stage here: https://github.com/theforeman/puppet-katello_devel/blob/master/manifests/setup.pp#L27-L29 (which executes while the installer is running)

So probably the very best approach would be the same general idea as above, but doing it within the puppet module directly instead of in forklift, for example by creating a new file resource immediately after we create the katello repository here: https://github.com/theforeman/puppet-katello_devel/blob/master/manifests/config.pp#L47-L51

Before I send you on any more wild goose chases, let's ask if this plan sounds good to @ekohl

ekohl:
We could enhance the generated bundler.d file to always include the test dependencies. That mirrors our CI so is probably the right thing anyway. And we should then do it for every plugin

ekohl:
To elaborate on that: https://github.com/theforeman/puppet-katello_devel/blob/master/templates/plugin.local.rb.erb can be extended to also look for `gemfile.d` in the same path as the plugin itself.

wbclark:
@qcjames53 is planning to open a new PR in puppet-katello_devel adding support for test gems in the plugin defined type. I'll close this one in the meantime. Thanks to you both!

PR #1838, opened 2024-08-16
Add backup tests for foreman-proxy

ehelms:


bats/fb-test-backup.bats
```diff
@@ -58,6 +58,18 @@ BACKUP_ONLINE_DIR="${BACKUP_BASE_DIR}/online"
   tFileExists "${BACKUP_OFFLINE_DIR}/pulp_data.tar"
 }
 
+@test "check offline backup contents for Foreman Proxy" {
```
ekohl, line 61: This test only works on content proxies. I'm not entirely sure if this is the right way:
```suggestion
@test "check offline backup contents for Foreman Proxy Content" {
```

Then it should also test for `foreman-installer-katello` to be present or otherwise skip it.

bats/fb-test-backup.bats
```diff
@@ -83,3 +97,15 @@ BACKUP_ONLINE_DIR="${BACKUP_BASE_DIR}/online"
...
+@test "check online backup contents for Foreman Proxy" {
+  tForemanMaintainAvailable
+
+  if tPackageExists foreman; then
+    skip "Stand-alone foreman-proxy test"
+  fi
+
+  tFileExists "${BACKUP_ONLINE_DIR}/pulp_data.tar"
+  tFileExists "${BACKUP_ONLINE_DIR}/pulpcore.dump"
+  tar -tvf "${BACKUP_ONLINE_DIR}/config_files.tar.gz" | grep "${HOSTNAME}.tar.gz"
```
jameerpathan111, line 110:
```suggestion
  tar -tvf "${BACKUP_ONLINE_DIR}/config_files.tar.gz" | grep "${HOSTNAME}-certs.tar"
```
It should be like this, or at least that's how the file on my machine is.

jameerpathan111 (changes requested):

bats/fb-test-backup.bats
ehelms: This depends on what you pass to the certs generate command. And our role in forklift uses: `foreman_proxy_content_certs_tar: "/root/{{ ansible_nodename }}.tar.gz"`

ekohl (approved):
@evgeni knows more about backups so I'd prefer if he also weighed in.

bats/fb-test-backup.bats
```diff
@@ -58,6 +58,18 @@ BACKUP_ONLINE_DIR="${BACKUP_BASE_DIR}/online"
   tFileExists "${BACKUP_OFFLINE_DIR}/pulp_data.tar"
 }
 
+@test "check offline backup contents for Foreman Proxy Content" {
+  tForemanMaintainAvailable
+
+  if ! tPackageExists foreman && tPackageExists foreman-installer-katello; then
+    tFileExists "${BACKUP_OFFLINE_DIR}/pulp_data.tar"
+    tFileExists "${BACKUP_OFFLINE_DIR}/pgsql_data.tar.gz"
```
evgeni, line 66: We don't do pgsql_data.tar.gz for offline backups anymore ;)
See lines 43++ how to check that properly based on Foreman version (well, you probably need to check installer version, there).

evgeni (changes requested):

bats/fb-test-backup.bats
ehelms: Makes sense, I had tested on Katello 4.13.

evgeni (approved):

PR #1839, opened 2024-08-16
Foreman 3.12 branching

pcreech:


vagrant/config/versions.yaml
```diff
@@ -101,6 +101,20 @@ installers:
...
 
+  - foreman: '3.12'
+    katello: '4.14'
+    candlepin: '4.4'
+    pulpcore: '3.49'
+    puppet: 8
+    boxes:
+      - 'almalinux8'
+      - 'almalinux9'
+      - 'centos8-stream'
```
ekohl, line 112: We should really drop it everywhere, but perhaps just fix 3.12 at least?
```suggestion
```

ekohl (approved):
This shouldn't be merged until we have RC1 in progress because it blocks nightly until then.

vagrant/config/versions.yaml
pcreech: removed

PR #1840, opened 2024-08-18
Allow foreman_proxy_content_dev to run as non-vagrant user

wbclark:


ekohl (approved):
Looks correct to me and I certainly like this formatting more.

PR #1842, opened 2024-08-20
Update configuration file

archanaserver:


ekohl (changes requested):
This PR is not very descriptive. When reading it I don't understand why this is being changed. The summary is also not very helpful.

When I start to look at it, I see the directory also has `pipelines/vars/common/foreman_nightly_centos7.yml` and `pipelines/vars/common/katello_nightly_centos7.yml` which makes me wonder: why do we keep those? And further, why do we even have these files? We've supported CentOS 9 for some time and the pipelines were passing.

I have some vague recollection @ehelms used these when manually testing and b62c9d9b63b38abd89c647f01c6dd3612018d480 (the commit that introduced it) doesn't show any code that actually uses it.

My advice would be to at least also remove CentOS 7 (which is also EOL) and come up with a better summary.

ekohl (approved):

PR #1843, opened 2024-08-20
Use centos9-stream as pipeline_os in ansible-lint

archanaserver:


ekohl (changes requested):
I'd summarize this as `Use centos9-stream as pipeline_os in ansible-lint`. Other than that I think this is OK.

ekohl (approved):
@evgeni mind taking a look as well?

evgeni (approved):

PR #1844, opened 2024-08-20
Drop Foreman < 3.8 versions support

archanaserver:
https://github.com/theforeman/forklift/pull/1826


ekohl (changes requested):
I think this current PR should be summarized as "Update forklift_versions role tests to use current versions" or similar.

But taking a step back, I'd combine this with another change: drop older boxes. We have [versions.yaml](https://github.com/theforeman/forklift/blob/master/vagrant/config/versions.yaml) and we can drop older versions there.

What we at least need to keep is currently supported versions (3.10, 3.11 & nightly) and what's needed to release them. To release 3.10, we need n-2 so 3.8. That means you can drop 3.7 and older.

To do that, you need to update the tests, but that's usually not the objective. So the PR summary should then be something like "Drop Foreman < 3.8 versions".

When dropping older version support, you can take a look if there are some version specific hacks that can also be dropped, but there is some downstream concern. Forklift is also used in Satellite so you may need to keep 3.5 support for now.

You can consider dropping the versions 1 by 1, each in a separate commit. Start with the oldest and work your way to the newest.

archanaserver:
> When dropping older version support, you can take a look if there are some version specific hacks that can also be dropped, but there is some downstream concern. Forklift is also used in Satellite so you may need to keep 3.5 support for now.

@ekohl I've been using this file to check the versions compatibility but how can I review the dependency for `candlepin`, `pulpcore` and `puppet` for the selected `foreman` and `katello` version?

roles/forklift_versions/molecule/default/verify.yml
```diff
@@ -59,27 +45,27 @@
...
+        scenario_version: "4.13"
         upgrade: True
         upgrade_step: 2
     - name: Ensure upgrade steps have been determined correctly
       assert:
         that:
-          - forklift_upgrade_version_start == '4.5'
-          - forklift_upgrade_version_intermediate == '4.7'
-          - forklift_upgrade_version_final == '4.9'
+          - forklift_upgrade_version_start == '4.9'
```
archanaserver, line 69: @evgeni how am i suppose to manage `4.9` version here when we no longer support this?

archanaserver: i believe the failure is coming form here

evgeni: You are correct, this fails now because you deleted 4.9. The way this test works (Upgrade from N-4 to N-2 to N), it means we need to keep N-4 in versions.yaml, even if it's unsupported by us anymore. Given there is 4.14 already, we could update `scenario_version` to 4.14 and then this test uses 4.10 - 4.12 - 4.14, which is enough to delete 4.9.
But then we have to keep 4.10 -- but I think you intended that anyway based on the title of this PR?

bats/foreman_helper.bash
```diff
@@ -46,10 +46,10 @@ tSkipIfOlderThan43() {
   fi
 }
 
-tSkipIfNewerThan45() {
```
evgeni, line 49: This is still used in https://github.com/theforeman/forklift/blob/d8db449a566e41b0476b1bbfdb737f48af611e32/bats/fb-katello-content.bats#L255-L256

However, we don't really have to care, as 4.5+ is the only thing we care about these days, so we can just delete the whole "add subscription to activation key" test and then also delete this `SkipIfNewer` definition.

bats/fb-test-foreman-rex.bats
```diff
@@ -8,8 +8,8 @@ load foreman_helper
 
 @test "run 'uptime' via Remote Execution" {
   FOREMAN_VERSION=$(tForemanVersion)
-  if ! tIsVersionNewer "${FOREMAN_VERSION}" 3.3; then
-    job_template='Run Command - SSH Default'
+  if ! tIsVersionNewer "${FOREMAN_VERSION}" 3.7; then
+    skip "Skipping test for versions older than 3.8"
```
evgeni, line 12: Why skipping it here? We only care for versions newer than 3.3, so we can do `job_template='Run Command - Script Default'` unconditionally and drop the whole `if/else` block

vagrant/config/versions.yaml
```diff
@@ -1,55 +1,11 @@
...
-    katello: '4.6'
-    pulpcore: '3.18'
-    puppet: 7
-    boxes:
-      - 'almalinux8'
-      - 'centos8-stream'
-      - 'debian11'
-      - 'ubuntu2004'
 
   - foreman: '3.5'
```
evgeni, line 3: This can be removed too -- While we want to keep support for supported Satellite releases (aka: Foreman 3.5) we do not need the definition here at that files is not used downstream.

evgeni (approved):

vagrant/boxes.d/00-centos.yaml
ekohl, line 1: Was dropping centos 7 & 8 intended to be part of this PR?

evgeni: I thought it's fine as a cleanup? :)

ekohl: I'd have preferred it to be more explicit and at least as a separate commit.

PR #1845, opened 2024-08-20
Drop centos8 and centos8-stream

archanaserver:


archanaserver:
Closed this because squash it here https://github.com/theforeman/forklift/pull/1844

PR #1847, opened 2024-08-20
add rocky linux box configuration

evgeni:


vagrant/boxes.d/00-rocky.yaml
```diff
@@ -0,0 +1,17 @@
+boxes:
+  rocky8:
+    box_name: 'rockylinux/8'
+    libvirt: https://dl.rockylinux.org/pub/rocky/8/images/x86_64/Rocky-8-Vagrant-Libvirt.latest.x86_64.box
```
evgeni, line 4: bleh, at least this one doesn't boot properly

ekohl: Why do we need to specify the URL? Doesn't https://app.vagrantup.com/rockylinux work (like CentOS)?

evgeni: It works like centos, yes, it yields a 404.

evgeni: https://github.com/theforeman/forklift/blob/2e259fe13406fc5dbec516cd63f9324660d86731/vagrant/boxes.d/00-centos.yaml#L34

ekohl: Is that what they mean with bug-for-bug compatible?

evgeni: :fire:

ekohl (approved):

PR #1849, opened 2024-09-02
workaround for broken libsemanage on CentOS Stream 9

evgeni:


ekohl (approved):
With the planned releases I'd like to have passing pipelines sooner rather than later so thanks!

PR #1850, opened 2024-09-02
force downgrade to libsemanage-3.6-1.el9.x86_64

evgeni:


evgeni:
#1849 wasn't enough, we need to explicitly downgrade the package

ekohl (approved):

PR #1853, opened 2024-09-04
add test for container push on Katello 4.14+

evgeni:


bats/fb-katello-container.bats
```diff
@@ -13,3 +13,13 @@ load fixtures/content
   CONTAINER_PULL_LABEL=$(echo "${ORGANIZATION_LABEL}-${PRODUCT_LABEL}-${CONTAINER_REPOSITORY_LABEL}"| tr '[:upper:]' '[:lower:]')
   podman pull "${HOSTNAME}/${CONTAINER_PULL_LABEL}"
 }
+
+@test "push container to katello" {
+  tContainerPushSupported
+  tPackageExists podman || tPackageInstall podman
+  podman login "${HOSTNAME}" -u admin -p changeme
+  CONTAINER_PULL_LABEL=$(echo "${ORGANIZATION_LABEL}-${PRODUCT_LABEL}-${CONTAINER_REPOSITORY_LABEL}"| tr '[:upper:]' '[:lower:]')
+  CONTAINER_PUSH_LABEL=$(echo "${ORGANIZATION_LABEL}/${PRODUCT_LABEL}/${CONTAINER_REPOSITORY_LABEL}-bats-$(date -u '+%s')"| tr '[:upper:]' '[:lower:]')
```
evgeni, line 22: Yes, you need to pull with `ORG-PRODUCT-REPO` but push with `ORG/PRODUCT/REPO`.

@ianballou this hurts :crying_cat_face:

bats/foreman_helper.bash
```diff
@@ -53,6 +53,13 @@ tSkipIfNewerThan45() {
   fi
 }
 
+tContainerPushSupported() {
+  KATELLO_VERSION=$(tKatelloVersion)
+  if ! tIsVersionNewer "${KATELLO_VERSION}" 4.13; then
```
evgeni, line 58: That's based on https://github.com/Katello/katello/pull/10952 -- is that a correct assumption?

bats/fb-katello-container.bats
ehelms: Aye it does - @ianballou is looking into whether we could switch to a `/` based naming scheme which is more common within the container world.

evgeni: interestingly, pushed stuff is then also only pull-able with `/` -- so `-` only applies to synced stuff? confusing!

ianballou (commented):
This looks fine at a glance, was it tested on a live system?

bats/foreman_helper.bash
ianballou: That PR is just the beginning, I need to check but I think 4.14 would be better.

bats/fb-katello-container.bats
evgeni: nah, let's keep as close to the default config as possible

ianballou (approved):
Looks like there was just enough of the container push feature implemented for this test to work in 4.13, nice. Ack!

bats/foreman_helper.bash
ianballou: @evgeni I confirmed that Katello 4.14 and higher should be used here.

ianballou (changes requested):

bats/foreman_helper.bash
evgeni: done!

ianballou (approved):
Looks good!

PR #1856, opened 2024-09-10
Add CentOS Stream 9 provisioning using theforeman.foreman modules

archanaserver:
This PR rewrite the CentOS Stream 9 provisioning tasks by replacing
Hammer CLI commands with [`theforeman.foreman`](https://github.com/theforeman/foreman-ansible-modules) modules.

For reference:  https://github.com/theforeman/forklift/pull/1832#discussion_r1680774786

### DIFF
--- a/roles/foreman_provisioning/tasks/configure_centos_9.yml
+++ b/roles/foreman_provisioning/tasks/configure_centos_9.yml
@@ -1,0 +1,21 @@
+- name: "Ensure CentOS Stream 9" # noqa: args[module]
+  theforeman.foreman.operatingsystem:
+    name: CentOS_Stream
+    family: Redhat
+    major: 9
+    architectures:
+      - x86_64
+    media:
+      - CentOS Stream 9 mirror
+    provisioning_templates:
+      - Kickstart default
+    ptables:
+      - Kickstart default
+    state: present
+
+- name: "Set default template for CentOS Stream 9" # noqa: args[module]
+  theforeman.foreman.os_default_template:
+    operatingsystem: "CentOS_Stream 9"
+    template_kind: "provision"
+    provisioning_template: "Kickstart default"
+    state: present
--- a/playbooks/foreman_provisioning.yml
+++ b/playbooks/foreman_provisioning.yml
@@ -2,6 +2,12 @@
   become: true
   vars:
     libvirt_tftp: true
+  module_defaults:
+    group/theforeman.foreman.foreman:
+      username: admin
+      password: changeme
+      server_url: "https://localhost/"
+      validate_certs: false
   roles:
     - foreman
     - libvirt
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,4 +1,7 @@
 collections:
+  - name: https://github.com/theforeman/foreman-ansible-modules.git
+    type: git
+  
   - name: https://github.com/theforeman/foreman-operations-collection
     type: git
 
--- a/roles/foreman_provisioning/tasks/main.yml
+++ b/roles/foreman_provisioning/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
-- name: 'Setup CentOS 7 provisioning'
-  import_tasks: configure_centos_7.yml
+- name: 'Setup CentOS 9 provisioning'
+  import_tasks: configure_centos_9.yml
 
 - name: 'Setup Fedora 27 provisioning'
   import_tasks: configure_fedora_27.yml


archanaserver:
@evgeni https://github.com/theforeman/forklift/pull/1856 I have raised the changes here, but can't run the task due to weird syntax error, how can i test this? I don't see any syntax error

`The error appears to be in '/home/akumari/Desktop/theforeman/forklift/roles/foreman_provisioning/tasks/configure_centos_9.yml': line 1, column 3, but may
be elsewhere in the file depending on the exact syntax problem.`

roles/foreman_provisioning/tasks/configure_centos_9.yml
```diff
@@ -0,0 +1,27 @@
...
+      - Kickstart default
+    state: present
+
+- name: "Set default template for CentOS Stream 9"
+  theforeman.foreman.os_default_template:
+    username: "{{ foreman_provisioning_username }}"
+    password: "{{ foreman_provisioning_password }}"
+    server_url: "{{ foreman_provisioning_server_url }}"
+    operatingsystem: "CentOS Stream 9"
+    template_kind: "kickstart"
```
evgeni, line 25:
```suggestion
    template_kind: "provision"
```

The linter is right there is no "kickstart" kind, just "provision"

roles/foreman_provisioning/tasks/configure_centos_9.yml
```diff
@@ -0,0 +1,27 @@
+- name: "Ensure CentOS Stream 9"
+  theforeman.foreman.operatingsystem:
+    username: "{{ foreman_provisioning_username }}"
+    password: "{{ foreman_provisioning_password }}"
+    server_url: "{{ foreman_provisioning_server_url }}"
+    name: CentOS Stream
```
evgeni, line 6: Names can't contain spaces. So it needs to be "CentOS" or "CentOSStream" or "CentOS_Stream".

Looking at Foreman core, we use `CentOS_Stream` there, so I'd keep that.

roles/foreman_provisioning/tasks/configure_centos_9.yml
```diff
@@ -0,0 +1,27 @@
+- name: "Ensure CentOS Stream 9"
+  theforeman.foreman.operatingsystem:
+    username: "{{ foreman_provisioning_username }}"
+    password: "{{ foreman_provisioning_password }}"
+    server_url: "{{ foreman_provisioning_server_url }}"
```
evgeni, line 5: Instead of having to pass username/password/url to every module, Ansible supports so-called module-defaults.

However, you can only define those in playbooks, not in roles (don't ask), which would mean we need something like

```
- hosts: all
  become: true
  vars:
    libvirt_tftp: true
  module_defaults:
    group/theforeman.foreman.foreman:
      username: admin
      password: changeme
      server_url: "https://localhost/"
      validate_certs: false
  roles:
    - foreman
    - libvirt
    - foreman_provisioning
```
for the `playbooks/foreman_provisioning.yml` playbook

roles/foreman_provisioning/tasks/configure_centos_9.yml
```diff
@@ -0,0 +1,27 @@
+- name: "Ensure CentOS Stream 9"
+  theforeman.foreman.operatingsystem:
```
evgeni, line 2: When you try to run this, it will complain you did not define `password_hash`.

Congratulations, you just found a bug in the Ansible modules! :)

I'll go fix it, but in the meantime you can add `password_hash: SHA512`

roles/foreman_provisioning/tasks/configure_centos_9.yml
```diff
@@ -0,0 +1,27 @@
...
+    ptables:
+      - Kickstart default
+    state: present
+
+- name: "Set default template for CentOS Stream 9"
+  theforeman.foreman.os_default_template:
+    username: "{{ foreman_provisioning_username }}"
+    password: "{{ foreman_provisioning_password }}"
+    server_url: "{{ foreman_provisioning_server_url }}"
+    operatingsystem: "CentOS Stream 9"
```
evgeni, line 24: If you change the name above to `CentOS_Stream`, this will become:
```suggestion
    operatingsystem: "CentOS_Stream 9"
```

roles/foreman_provisioning/tasks/configure_centos_9.yml
evgeni: https://github.com/theforeman/foreman-ansible-modules/pull/1779

roles/foreman_provisioning/tasks/configure_centos_9.yml
ekohl: Looks like the linter doesn't understand that.

ekohl:
This would be a bit easier to review if the `centos7 and centos8 boxes name update` commit wasn't in here.

roles/foreman_provisioning/tasks/configure_centos_9.yml
archanaserver: @evgeni @ekohl to fix this what is the best approach to resolve this? Should I just add this in `.ansible-lint` skip_list?

roles/foreman_provisioning/tasks/configure_centos_9.yml
evgeni, line 1: This file is currently not added to `tasks/main.yml` and thus won't be executed.

evgeni (commented):
I think this PR/commit should be called "add CentOS Stream 9 provisioning using theforeman.foreman modules" now -- you're not refactoring anything as such.

(and then the other PR drops the 7/8 stuff)

evgeni (approved):

PR #1857, opened 2024-09-13
use db dumps for offline backups too in clone pipeline

evgeni:


ekohl (approved):
I recall this changed at some point. Do we need to care about versions?

evgeni:
> I recall this changed at some point. Do we need to care about versions?

We do not. This pipeline (clone) is not executed anywhere automatically, I just use it for sat-clone tests myself.

PR #1864, opened 2024-10-04
Attempt to fix container search bats issue with a delay

ianballou:
5 seconds should be adequate. If it's not the case, then we likely have a podman push bug somewhere.


ekohl (commented):
I'm not a fan of using sleep as a synchronization mechanism. Is there a reason `podman push` returns before it becomes visible? Does the spec for pushes say anything about this?

Note I don't object to merging this, but I'd at least like to understand it.

ianballou:
> I'm not a fan of using sleep as a synchronization mechanism. Is there a reason `podman push` returns before it becomes visible? Does the spec for pushes say anything about this?
> 
> Note I don't object to merging this, but I'd at least like to understand it.

I don't like the sleep either,  but so far it's only reproducible on the test upgrade pipeline for centos9-stream in CI, not even locally. There shouldn't be any delay.  Since the push appears to be working, this sleep should at least tell is if its timing, or if there's a rare bug going on where the container repo isn't getting pushed properly. I suppose either way it's a rare bug :)

ianballou:
My thought is to remove the sleep as soon as possible since I'm not sure how else to debug the CI pipeline.

ekohl:
Did you run the pipeline locally?

ianballou:
> Did you run the pipeline locally?

Yeah this was tested last week:
```
ansible-playbook pipelines/upgrade_pipeline.yml -e pipeline_os=centos9-stream -e pipeline_type=katello -e pipeline_version=nightly
```

ianballou:
Just popping in to say that the pipeline passed, so the issue is indeed intermittent.

ekohl (approved):
That does sound like a timing issue. Mind adding a comment to explain? Perhaps just link to this PR

PR #1865, opened 2024-10-15
update foreman-maintain before stopping services

evgeni:
This ensures any possible fixes for process handling are available.

### DIFF
--- a/roles/foreman_installer/tasks/upgrade.yml
+++ b/roles/foreman_installer/tasks/upgrade.yml
@@ -1,4 +1,12 @@
 ---
+- name: 'Update foreman-maintain'
+  package:
+    name: rubygem-foreman_maintain
+    state: latest
+  when:
+    - foreman_installer_scenario != 'foreman'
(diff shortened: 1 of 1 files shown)


evgeni:
I am unsure I want to keep this forever (as it's how a Satellite upgrade works under the hood) and update our docs, or just use it to test the fix for https://community.theforeman.org/t/katello-nightly-rpm-pipeline-2155-failed/39693/ and backport the actual fix to the 1.7.z branch afterwards.

ehelms (approved):

ekohl (approved):
I'd like to avoid keeping this forever.

In fact, what I'd prefer is that we find a solution where we don't need foreman-maintain at all to stop the services. I've always wonder if we can define a `foreman.target` and somehow get `systemd` to do the hard work.

evgeni:
https://github.com/theforeman/foreman_maintain/pull/942

evgeni:
https://github.com/theforeman/forklift/pull/1867

PR #1872, opened 2024-10-25
Make rpm_packaging playbook runnable against RHEL9 targets

adamruzicka:


ehelms:
BTW, we have a container in the packaging repository now that aims to do this but in a way that works with the repository:

https://github.com/theforeman/foreman-packaging/blob/rpm/develop/Containerfile
https://github.com/theforeman/foreman-packaging/blob/rpm/develop/build-container
https://github.com/theforeman/foreman-packaging/blob/rpm/develop/enter-container

ehelms:
> I'm aware of that. I'd say that either the rpm_packaging playbook should work where we need it to or that we shouldn't carry it at all, if we want to force people to go the container way.

I am happy to accept these changes, I just wanted to make sure you were aware of it as we try to keep that one up to date. If this works better for you, by all means, use it.

ehelms (approved):

PR #1854, opened 2024-09-04
use ansible_default_ipv4 if possible when writing /etc/hosts

evgeni:

PR #1875, opened 2024-10-29
use ansible_default_ipv4 fact if possible when writing /etc/hosts

evgeni:


evgeni:
This is a fixed version of https://github.com/theforeman/forklift/pull/1854 that got reverted (https://github.com/theforeman/forklift/pull/1861) because it broke pipelines.

ehelms (approved):

PR #1884, opened 2024-11-26
Check for Katello version properly for container tests

ianballou:
`$(tKatelloVersion)` replaces the undefined KATELLO_VERSION used in the tests previously.

Also updates the container tests in the client test file.


ianballou:
This PR should fix nightlies. I did some testing on a 4.15 box.

ekohl (approved):
Thanks!

PR #1885, opened 2024-11-29
Associate more templates with CentOS Stream

adamruzicka:
Followup on https://github.com/theforeman/forklift/pull/1856

### DIFF
--- a/roles/foreman_provisioning/tasks/configure_centos_9.yml
+++ b/roles/foreman_provisioning/tasks/configure_centos_9.yml
@@ -1,22 +1,8 @@
-- name: 'Set templates fact'
-  set_fact:
-    templates:
-      - kind: 'host_init_config'
-        name: 'Linux host_init_config default'
-      - kind: 'provision'
-        name: 'Kickstart default'
-      - kind: 'PXEGrub'
-        name: 'Kickstart default PXEGrub'
-      - kind: 'PXEGrub2'
-        name: 'Kickstart default PXEGrub2'
-      - kind: 'PXELinux'
-        name: 'Kickstart default PXELinux'
-      - kind: 'finish'
-        name: 'Kickstart default finish'
-      - kind: 'iPXE'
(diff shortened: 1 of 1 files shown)


evgeni (approved):
I haven't tested this myself, but the logic reads correctly to me

ekohl (approved):

roles/foreman_provisioning/tasks/configure_centos_9.yml
```diff
@@ -7,15 +13,16 @@
       - x86_64
     media:
       - CentOS Stream 9 mirror
-    provisioning_templates:
-      - Kickstart default
+    provisioning_templates: "{{ result.resources | map(attribute='name') }}"
     ptables:
       - Kickstart default
     state: present
+    password_hash: "SHA256"
```
ekohl, line 20: We made this the default, but I don't mind making it explicit now because we didn't add a migration to change existing OSes.

evgeni: Ah yes, probably Adam didn't have https://github.com/theforeman/foreman-ansible-modules/pull/1779 in his setup :)

adamruzicka: I indeed did not have that

evgeni: Heh, I recall fixing that after Archana run into the same thing when initially writing this play.

ekohl: Funny, I thought it was related to https://github.com/theforeman/foreman/commit/e2dee7d6f7ffb505d317ca1b205f302ccd622063 but now that I read the PR itself it should really be the more secure SHA512.
```suggestion
    password_hash: "SHA512"
```

adamruzicka: https://github.com/theforeman/forklift/pull/1886

PR #1886, opened 2024-12-03
Change password hash to sha512 to centos stream 9

adamruzicka:


ekohl (approved):
I don't mind being explicit here because we can't really update our requirements to depend on a certain version:
https://github.com/theforeman/forklift/blob/4f7e13e0e0f899ce6eab9f7f05167dc15e5d09c6/requirements.yml#L2-L3

PR #1866, opened 2024-10-17
Fix foreman_provisioning role

adamruzicka:
The role expects that the user has `community.libvirt` role installed. With these changes, the role should at least pass.

PR #1888, opened 2024-12-03
Rely on FAM for much of the foreman_provisioning setup

ekohl:
This isn't tested now and needs some proper review. I'll highlight some questions inline.

### DIFF
--- a/roles/foreman_provisioning_infrastructure/tasks/compute_profiles.yml
+++ b/roles/foreman_provisioning_infrastructure/tasks/compute_profiles.yml
@@ -1,22 +1,21 @@
 ---
-- name: 'find compute profile'
-  shell: >
-    {{ foreman_provisioning_hammer }} compute-profile info --name "libvirt-profile"
-  register: foreman_provisioning_compute_profile
-  ignore_errors: True
-
-- name: 'create compute profile'
-  shell: >
-    {{ foreman_provisioning_hammer }} compute-profile create
-    --name "libvirt-profile"
-  when: "'Error' in foreman_provisioning_compute_profile.stderr"
-
 - name: 'create compute attributes'
-  shell: >
-    {{ foreman_provisioning_hammer }} compute-profile values create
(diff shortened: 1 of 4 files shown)


roles/foreman_provisioning_infrastructure/tasks/compute_profiles.yml
```diff
@@ -1,22 +1,21 @@
...
+      - compute_resource: libvirt
+        vm_attrs:
+          cpus: 1
+          memory: 2400000000
+          start: 1
+          nics_attributes:
+            0:
+              type: network
+              network: provision
+              model: virtio
```
ekohl, line 16: @evgeni this is taken from the examples in https://theforeman.github.io/foreman-ansible-modules/develop/plugins/compute_profile_module.html#ansible-collections-theforeman-foreman-compute-profile-module but the description says it's a list, so should it be
```suggestion
            - type: network
              network: provision
              model: virtio
```

roles/foreman_provisioning_infrastructure/tasks/main.yml
```diff
@@ -1,126 +1,64 @@
 ---
 # Make sure admin does not default to any taxonomy
 - name: 'disable default context for admin'
+  # TODO: how to pass org and location IDs 0 to FAM?
```
ekohl, line 4: I couldn't find this on https://theforeman.github.io/foreman-ansible-modules/develop/plugins/user_module.html#ansible-collections-theforeman-foreman-user-module and I'm wonder how this even works.

ekohl (commented):
This together with https://github.com/theforeman/forklift/pull/1866 can mean we no longer need Hammer and we can drop this bit:
https://github.com/theforeman/forklift/blob/4f7e13e0e0f899ce6eab9f7f05167dc15e5d09c6/roles/foreman_provisioning/meta/main.yml#L3

I also wonder if we should move this into `foreman_provisioning` and drop the `foreman_provisioning_infrastructure` role.

PR #1890, opened 2024-12-17
use a login shell when using become for bats

evgeni:
otherwise the puppet path (/opt/puppetlabs/bin) is not loaded and tests
fail to call puppetserver

### DIFF
--- a/roles/bats/tasks/main.yml
+++ b/roles/bats/tasks/main.yml
@@ -43,6 +43,7 @@
       register: "test_output"
       ignore_errors: true
       environment: "{{ bats_environment }}"
+      # Adds -i to Ansible's default
       become_flags: '-H -S -n -i'
 
     - name: "Read results"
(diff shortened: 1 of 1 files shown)


roles/bats/tasks/main.yml
```diff
@@ -43,6 +43,7 @@
       register: "test_output"
       ignore_errors: true
       environment: "{{ bats_environment }}"
+      become_flags: '-H -S -n -i'
```
evgeni, line 46: HSn is Ansible's default, we're adding `-i` to it.

ekohl: I always like long options
```suggestion
      become_flags: '--set-home --stdin --non-interactive --login'
```

ekohl (approved):

roles/bats/tasks/main.yml
evgeni: I thought about that, but then erred on the side of a smaller diff (ansible uses short opts internally).

We could do `-H -S -n --login`?

PR #1892, opened 2025-01-09
Update haproxy role and make it more generic

ehelms:


roles/haproxy/templates/haproxy.cfg.j2
```diff
@@ -61,8 +61,8 @@ frontend https
 backend f-proxy-https
    option tcp-check
    balance source
-{% for host in foreman_proxies %}
-   server f-proxy-https-{{loop.index}} {{ hostvars[host].ansible_default_ipv4.address }}:443 check
+{% for host in haproxy_targets %}
+   server f-proxy-https-{{loop.index}} {{ host }}:443 check
```
ekohl, line 65: Should the port also be in the variable? Perhaps even make the var a dict: service -> backends where each backend is just hostname:port

ehelms: As in go fully generic? I was intending to keep this specific to our load-balancer use case, but remove the fact that it was tightly coupled to how Vagrant works within Forklift to have an inventory.

ekohl: I'm debating it. On the one hand, it's good but if we go that route perhaps we can use some community collection instead. You could design an intermediate layer too. On the other hand, it may also be overkill.

I'm fine merging this now (provided tests pass) and leave it to a future improvement. Just want to make sure we at least consider it

PR #1893, opened 2025-01-13
force Vagrant to use the "new" API URL

evgeni:
It seems Vagrant has introduced a new API and their proprietary client
does use that by default now, so they turned off the old one (or the
redirect, I really have no idea what was there last week).

Just setting the new URL (which I found out by using `vagrant --debug`)
as the server URL works for now, so let's do that, and at least unblock
nightlies.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,6 +4,7 @@
 
 ENV['LANG'] = 'en_US.UTF-8'
 ENV['LC_ALL'] = 'en_US.UTF-8'
+ENV['VAGRANT_SERVER_URL'] ||= 'https://vagrantcloud.com/api/v2/vagrant'
 VAGRANTFILE_DIR = File.dirname(__FILE__)
 
 require "#{VAGRANTFILE_DIR}/vagrant/lib/forklift"


evgeni:
I have filed https://bugzilla.redhat.com/show_bug.cgi?id=2337302 but do not expect much movement, given the state of Vagrant.

vagrant/boxes.d/00-centos.yaml
```diff
@@ -1,4 +1,14 @@
 boxes:
+  centos8-stream:
```
ekohl, line 2: Any reason you're adding centos8-stream back?

ekohl (commented):
Initially @ianballou also found this in https://github.com/theforeman/forklift/pull/1859 but I see that part has now been reverted again.

vagrant/boxes.d/00-centos.yaml
evgeni: yes, me failing at git, why else?

ekohl (commented):
Can we force this in our wrapper instead so it benefits local uses as well? We already have some:
https://github.com/theforeman/forklift/blob/e7a0f211b6d817fdd972d0a89879538bf475ad7e/Vagrantfile#L5-L6

So I'd suggest
```ruby
ENV['VAGRANT_SERVER_URL'] ||= https://vagrantcloud.com/api/v2/vagrant
```

Or would you prefer to avoid that for people who have moved to the new client?

evgeni:
I haven't tested if that works, but have no objection doing so there.
Nobody should be using the non-free client ;)

evgeni:
Tested. This is actually better, as it also fixes `vagrant up …` usage w/o the pipes!

ekohl (approved):

PR #1894, opened 2025-01-13
Fix freeipa_server role

adamruzicka:
Server installation should now pass. Client installation is only run when katello_server_origin is defined.


ekohl (approved):
I see there's also `--disable-system-checks` which we dropped quite some time ago:
https://github.com/theforeman/forklift/blob/55674b755d439cf0cebbc226e04ff235d7194e62/roles/freeipa_server/tasks/install_freeipa_client.yml#L68

From seeing this I assume you didn't use the client role and only the server.

adamruzicka:
> From seeing this I assume you didn't use the client role and only the server.

Yes, I'm taking it one step at a time

PR #1897, opened 2025-02-04
Add inventory for working with broker

ehelms:
This adds a basic inventory for the [broker project](https://github.com/SatelliteQE/broker)

### DIFF
--- a/inventories/broker.py
+++ b/inventories/broker.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import yaml
+import os
+import subprocess
+import sys
+from io import StringIO
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Broker inventory script")
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument('--list', action='store_true')
+    group.add_argument('--host')
+    return parser.parse_args()
+
+
+def get_running_hosts():
+    cmd = ["broker", "inventory", "--details"]
+
+    try:
+        output = subprocess.check_output(cmd, universal_newlines=True).rstrip()
+    except FileNotFoundError:
+        return
+
+    hosts = yaml.safe_load(output)
+    return hosts.values()
+
+
+def list_running_hosts():
+    hosts = get_running_hosts()
+    variables = dict(get_configs(hosts))
+
+    return {
+        "_meta": {
+            "hostvars": variables
+        },
+        "all": {
+            "hosts": list(variables.keys())
+        },
+    }
+
+
+def get_configs(hosts):
+    if not hosts:
+        return
+
+    for host in hosts:
+        details = {
+            'ansible_host': host['ip'],
+            'ansible_port': '22',
+            'ansible_user': 'root'
+        }
+        yield host['hostname'], details
+
+
+def main():
+    args = parse_args()
+    hosts = list_running_hosts()
+
+    if args.list:
+        json.dump(hosts, sys.stdout)
+    elif args.host:
+        details = hosts['_meta']['hostvars']
+        json.dump(details[args.host], sys.stdout)
+
+
+if __name__ == '__main__':
+    main()


ehelms:
I see the ansible errors in other PRs so I think they are unrelated to this inventory being added.

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
...
+        },
+        "all": {
+            "hosts": [host['hostname'] for host in hosts],
+        },
+    }
+
+
+def get_configs(hosts):
+    if not hosts:
+        return
```
ekohl, line 61: Does this really help? I suppose for the case where `hosts` is `None` instead of an empty list.

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
...
+def list_running_hosts():
+    hosts = get_running_hosts()
+    variables = dict(get_configs(hosts))
+
+    return {
+        "_meta": {
+            "hostvars": variables,
+        },
+        "all": {
+            "hosts": [host['hostname'] for host in hosts],
```
ekohl, line 54: Isn't this equivalent:
```suggestion
            "hosts": variables.keys(),
```

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python
```
ekohl, line 1: I'd make this Python 3 only
```suggestion
#!/usr/bin/env python3
```

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
...
+    from StringIO import StringIO  # pyright: reportMissingImports=false
+except ImportError:
+    from io import StringIO  # pyright: reportMissingImports=false
+
+from collections import defaultdict
+
+
+try:
+    DEVNULL = subprocess.DEVNULL
+except AttributeError:
```
ekohl, line 20: It was added in Python 3.3 so if you make the script Python 3 only then I think you're guaranteed that it's present. See https://docs.python.org/3/library/subprocess.html#subprocess.DEVNULL

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
...
+import os
+import subprocess
+import sys
+
+try:
+    from StringIO import StringIO  # pyright: reportMissingImports=false
+except ImportError:
+    from io import StringIO  # pyright: reportMissingImports=false
+
+from collections import defaultdict
```
ekohl, line 15: This is unused

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
...
+    group.add_argument('--list', action='store_true')
+    group.add_argument('--host')
+    return parser.parse_args()
+
+
+def get_running_hosts():
+    try:
+        subprocess.check_call(["which", "broker"], stdout=DEVNULL)
+    except subprocess.CalledProcessError:
+        return
```
ekohl, line 36: I'd rely on catching `FileNotFoundError` coming from `subprocess.check_output` instead of calling `which`

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
...
+
+    cmd = "broker inventory --details"
+    output = subprocess.check_output(cmd.split(), universal_newlines=True).rstrip()
```
ekohl, line 39: I'd write out the list instead of splitting a string, even if in practice it probably doesn't matter
```suggestion
    cmd = ["broker", "inventory", "--details"]
    output = subprocess.check_output(cmd, universal_newlines=True).rstrip()
```

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
...
+import json
+import yaml
+import os
+import subprocess
+import sys
+
+try:
+    from StringIO import StringIO  # pyright: reportMissingImports=false
+except ImportError:
+    from io import StringIO  # pyright: reportMissingImports=false
```
ekohl, line 13: If you make it Python 3-only you can also simplify this to just `from io import StringIO`.

inventories/broker.py
```diff
@@ -0,0 +1,84 @@
...
+
+def main():
+    args = parse_args()
+    hosts = list_running_hosts()
+
+    if args.list:
+        json.dump(hosts, sys.stdout)
+    elif args.host:
+        details = hosts['_meta']['hostvars']
+        json.dump(details[args.host], sys.stdout)
```
ekohl, line 80: Do you want to catch `KeyError` or do you assume the user just did the right thing? :)

ekohl (commented):
Perhaps turning it around: would it make sense for broker to provide an Ansible inventory script?

inventories/broker.py
ehelms: Yep.

ehelms:
> Perhaps turning it around: would it make sense for broker to provide an Ansible inventory script?

I thought about that, then it wasn't clear how to point Forklift at that reliably. And then this was easy, and made my life better :)

inventories/broker.py
ehelms: No idea, but I'll try it.

inventories/broker.py
ehelms: Yea, I stole this from the vagrant one, so probably time to do it there too.

inventories/broker.py
```diff
@@ -0,0 +1,71 @@
...
+    return parser.parse_args()
+
+
+def get_running_hosts():
+    cmd = ["broker", "inventory", "--details"]
+
+    try:
+        output = subprocess.check_output(cmd, universal_newlines=True).rstrip()
+    except FileNotFoundError:
+        return
```
ekohl, line 26: Perhaps raise a `SystemExit('Broker command not found')` instead of silently accepting the error?

ekohl (approved):

inventories/broker.py
ehelms: Since this inventory is optional, I think it's better to "fail" silently. I realize that doesn't help if you intentionally want broker but screwed something up. If this raises an error, you end up with this output everytime:

```
[WARNING]:  * Failed to parse /home/ehelms/workspace/upstream/forklift/inventories/broker.py with script plugin: Inventory script (/home/ehelms/workspace/upstream/forklift/inventories/broker.py) had an execution error: Broker not
found
[WARNING]:  * Failed to parse /home/ehelms/workspace/upstream/forklift/inventories/broker.py with ini plugin: /home/ehelms/workspace/upstream/forklift/inventories/broker.py:3: Expected key=value host variable assignment, got:
argparse
[WARNING]: Unable to parse /home/ehelms/workspace/upstream/forklift/inventories/broker.py as an inventory source
```

We do the same thing for the vagrant.py plugin, for the same reason (you can use forklift without Vagrant). https://github.com/theforeman/forklift/blob/master/inventories/vagrant.py#L36-L39

ekohl: If Ansible always loads all inventories in `inventories` then I think you're right.

PR #1902, opened 2025-03-18
make luna demo env work on el9

evgeni:


ekohl:
https://github.com/theforeman/forklift/pull/1866 & https://github.com/theforeman/forklift/pull/1888 work towards fixing the `foreman_provisioning_infrastructure` role and making it easier to maintain.

PR #1906, opened 2025-05-21
enable testing of HDM in the plugins pipeline

evgeni:


evgeni:
FWIW, I did test that on nightly with both Debian and CentOS boxen.

PR #1908, opened 2025-06-05
use "sos report", not "sosreport" to generate a sos report

evgeni:
On EL9, calling `sosreport` just prints
```
sosreport binary is deprecated, use 'sos report' instead
```
And doesn't generate any report.

Using `sos report` as suggested works, and that also works on Debian
(where calling `sosreport` still works -- with a deprecation warning).

### DIFF
--- a/roles/sos_report/tasks/main.yml
+++ b/roles/sos_report/tasks/main.yml
@@ -8,7 +8,7 @@
     state: present
 
 - name: 'Generate sosreport'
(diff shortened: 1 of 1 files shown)


pmoravec (approved):
Absolutely ACK from a sos developer, the binary name change has been promoted for long.

`sosreport` binary on EL should at least return nonzero, sos maintainers will improve this independently.

PR #1911, opened 2025-06-12
Remove oVirt installer option

stejskalleos:


stejskalleos:
> there is no ovirt-related data in the DB

Yes, the oVirt integration hasn't been working properly for quite some time, and there is probably nothing to migrate anyway.

evgeni (approved):

PR #1913, opened 2025-08-08
Foreman 3.16 branching

ogajduse:


vagrant/config/versions.yaml
```diff
@@ -93,6 +93,17 @@ installers:
...
+    puppet: 8
+    boxes:
+      - 'almalinux9'
+      - 'centos9-stream'
+      - 'debian12'
+      - 'ubuntu2204'
+
   - foreman: 'nightly'
     katello: 'nightly'
     candlepin: '4.4'
```
ianballou, line 109: A bit outside the scope I know, but do you mind bumping this to 4.6 as well?

ianballou (commented):
Looks good, just one side comment

vagrant/config/versions.yaml
ogajduse: updated

PR #1947, opened 2026-05-29
Fixes #39330 - Update bats test for removed CV/LCE activation key params

jeremylenz:
## Summary

- Replace deprecated `--content-view` and `--lifecycle-environment` flags in `hammer activation-key create` with the new `--content-view-environments="LCE/CV"` format
- Follows the Katello change that removed backward-compat params for separate CV/LCE fields (Katello/katello#11753)

## Test plan

- [ ] Run the `fb-katello-content` bats pipeline and confirm the "create activation key" test passes

### DIFF
--- a/bats/fb-katello-content.bats
+++ b/bats/fb-katello-content.bats
@@ -253,7 +253,7 @@ setup() {
 
 @test "create activation key" {
   hammer activation-key create --organization="${ORGANIZATION}" \
-    --name="${ACTIVATION_KEY}" --content-view="${CONTENT_VIEW}" --lifecycle-environment="${LIFECYCLE_ENVIRONMENT}" \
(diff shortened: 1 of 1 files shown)


jturel (approved):

jeremylenz:
@jturel thanks for the ack :) If you can merge, feel free (I don't seem to have access.)

jturel:
Sadly I can't :(

chris1984:
@ekohl can this get a merge?

ekohl (approved):
I'm looking at https://projects.theforeman.org/issues/39330 and it only refers to a Jira issue when it was deprecated. Not on my work machine so I can't view it.

So my question is whether will fail on still supported Katello versions. If so, this really deserves a version check.

jeremylenz:
The new `content_view_environments` params have been available for quite some time now, and I'm pretty sure they would work on all supported versions. Is that enough to not require a version check?

PR #1948, opened 2026-06-01
Refs #39330 - Use label for content-view-environments activation key param

jeremylenz:
## Summary

- Use `CONTENT_VIEW_LABEL` (`Test_CV`) instead of `CONTENT_VIEW` (`Test CV`) for the `--content-view-environments` flag, which expects labels not names

## Test plan

- [ ] Run the `fb-katello-content` bats pipeline and confirm the "create activation key" test passes


jturel (approved):

jeremylenz:
Also switched label definitions from `${VAR/ /_}` to `${VAR// /_}` — single slash only replaces the first space, double slash replaces all of them.

jeremylenz:
The ShellCheck SC2034 ("appears unused") warnings are false positives — this is a fixtures file that gets sourced by the bats tests, so the variables are used externally. Added a file-level disable.

zjhuntin:
Confirmed on a dev box this is working.

zjhuntin (approved):

PR #1951, opened 2026-06-15
use OpenVox by default

evgeni:


evgeni:
Draft, as I haven't tested any of this yet

evgeni:
Tested successfully:
- Fresh install of Foreman nightly on CentOS Stream 9
- Fresh install of Foreman nightly on Ubuntu 22.04
- Fresh install of Foreman 3.18 on CentOS Stream 9
- Fresh install of Foreman 3.18 on Ubuntu 22.04
- Upgrade to Foreman nightly (from 3.18) on CentOS Stream 9
- Upgrade to Foreman nightly (from 3.18) on Ubuntu 22.04
- Upgrade to Foreman 3.19 (from 3.17) on CentOS Stream 9
- Upgrade to Foreman 3.19 (from 3.17) on Ubuntu 22.04

Tested, but failing:
- Upgrade to Foreman 3.18 (from 3.16) on CentOS Stream 9
  - Fails to install 3.16 as only 3.17+ has support for OpenVox packages: https://github.com/theforeman/puppet-puppet/blob/master/CHANGELOG.md#2200-2025-11-06
- Upgrade to Foreman 3.18 (from 3.16) on Ubuntu 22.04
  - Fails the same way as CentOS Stream

As the failure is on 3.16, it doesn't really affect real world deployments (that version is EOL), but does affect our pipelines as we're using that as a baseline for upgrade tests.

@ogajduse I'd prefer not to add some complicated "if 3.16 for upgrades, do puppet, otherwise openvox" logic, and just say we don't care about 3.16 (and thus upgrades to 3.18) for the odd case we will have another 3.18 release. WDYT

ogajduse:
> As the failure is on 3.16, it doesn't really affect real world deployments (that version is EOL), but does affect our pipelines as we're using that as a baseline for upgrade tests.
> 
> @ogajduse I'd prefer not to add some complicated "if 3.16 for upgrades, do puppet, otherwise openvox" logic, and just say we don't care about 3.16 (and thus upgrades to 3.18) for the odd case we will have another 3.18 release. WDYT

I say we don't care about upgrades from 3.16 _too much_, as even 3.17 is EOL. 

If we do another 3.18 release, what is the patch that we should include in it?

evgeni:
> > As the failure is on 3.16, it doesn't really affect real world deployments (that version is EOL), but does affect our pipelines as we're using that as a baseline for upgrade tests.
> > @ogajduse I'd prefer not to add some complicated "if 3.16 for upgrades, do puppet, otherwise openvox" logic, and just say we don't care about 3.16 (and thus upgrades to 3.18) for the odd case we will have another 3.18 release. WDYT
>
> If we do another 3.18 release, what is the patch that we should include in it?

We don't need any changes in 3.18 to support Vox (you did the last packaging bits last week). The only "problem" is that our pipelines to release 3.18 include the upgrade test from 3.16, and *that* will fail. We can skip it (or see that we reduce it to 3.17→3.18 upgrade)

ogajduse:
> We can skip it (or see that we reduce it to 3.17→3.18 upgrade)

I see. LGTM then. Thank you, Evgeni!

ogajduse (approved):

PR #1952, opened 2026-06-16
Use ansible_facts dict instead of top level facts

evgeni:
This fixes the massive amount of warnings Ansible has been raising.


ogajduse (approved):
Adjusting to the change seems more appropriate then trying to bring the "old" behavior back.

```ini
[defaults]
inject_facts_as_vars = True
```

PR #1964, opened 2026-07-08
Use dnf allow_downgrade for openssl EL9 workaround

zjhuntin:
ansible.builtin.package with state: present won't downgrade when a newer version is already installed. Switch to ansible.builtin.dnf with allow_downgrade: true so the pinned NVRs actually replace the broken 3.5.7-1 packages.

### DIFF
--- a/roles/vagrant_workarounds/tasks/main.yml
+++ b/roles/vagrant_workarounds/tasks/main.yml
@@ -15,23 +15,21 @@
     - ansible_facts['os_family'] == 'Debian'
 
 # workaround for https://issues.redhat.com/browse/RHEL-192424
+- name: force downgrade to openssl-3.5.5-3.el9
+  ansible.builtin.dnf:
+    name: openssl-3.5.5-3.el9
+    state: present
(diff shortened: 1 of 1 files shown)


Odilhao (approved):

ehelms:
Did you consider that the dnf module has a provision for downgrade? https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/dnf_module.html#parameter-allow_downgrade

zjhuntin:
@ehelms I did and tried it locally with that, but it failed :frowning:

ehelms:
This worked for me:

```
  tasks:
    - name: Downgrade openssl to an older version
      ansible.builtin.dnf:
        name: openssl-3.5.5-3.el9
        state: present
        allow_downgrade: yes                       
```

zjhuntin:
@ehelms my bad, it does work with that. I had tried it with all three packages and it did not work, with just the one that works, thanks!

ehelms (approved):

PR #1965, opened 2026-07-08
Ensure that the workaround for openssl is only executed on Centos Stream 9

Odilhao:


ogajduse (approved):

roles/vagrant_workarounds/tasks/main.yml
```diff
@@ -23,6 +23,7 @@
   when:
     - ansible_facts['os_family'] == 'RedHat'
```
evgeni, line 24: os_name is CentOS and could have been used here

Odilhao: I think the problem for me was that I did run 

`ansible localhost -i "localhost," -c local -m setup  | grep os_name` and it returned blank, probably because I did that inside of one container.

I used this fact because: of this:

```
ansible localhost -i "localhost," -c local -m setup -a "filter=ansible_distribution*"
localhost | SUCCESS => {
    "ansible_facts": {
        "ansible_distribution": "CentOS",
        "ansible_distribution_file_parsed": true,
        "ansible_distribution_file_path": "/etc/centos-release",
        "ansible_distribution_file_variety": "CentOS",
        "ansible_distribution_major_version": "9",
        "ansible_distribution_release": "Stream",
        "ansible_distribution_version": "9",
        "discovered_interpreter_python": "/usr/bin/python3"
    },
    "changed": false
}

```

evgeni: Or distribution, that works too.

Never remember the right name between facter/puppet and ansible

PR #1972, opened 2026-07-23
Fix env stat check

sbernhard:
Before this commit, the task 'Check that the .env exists' failed with:

TASK [customize_home : Ensure .env file contains current hostname] *************
[ERROR]: Task failed: Conditional result was "{'changed': False, 'stat': [...]" of type 'dict', which evaluates to True. Conditionals must have a boolean result.


sbernhard:
@evgeni can you review this, please?

evgeni:
Why me? It looks sane, but I never touched that role (knowingly)

sbernhard:
> Why me? It looks sane, but I never touched that role (knowingly)

https://github.com/theforeman/forklift/commit/c240d11dd9486429beb23f149ccf167021b3d212

evgeni:
:woman_shrugging:

PR #1976, opened 2026-08-03
Support EL10 in beaker role

Odilhao:
## Summary

Ports the EL10 acceptance testing capability from theforeman/puppet-pulpcore#411 / https://github.com/theforeman/puppet-pulpcore/pull/412 (now on `master`) into forklift's real pulpcore pipeline (`pipelines/pulpcore/{01-boxes,02-install,03-tests}.yml`), which is what runs in Jenkins at `pulpcore-nightly-rpm-pipeline`.

- `pipelines/pulpcore/03-tests.yml`: on EL10, override `BEAKER_PUPPET_COLLECTION` to `openvox8`. Puppet doesn't publish an EL10 release RPM yet (`puppet8-release-el-10.noarch.rpm` 404s), but OpenVox does (`openvox8-release-el-10.noarch.rpm`). EL9 and other OSes are unaffected.
- `roles/beaker/tasks/main.yml`:
  - Extended the existing EL9 "enable CRB" task to also cover EL10 (needed for build deps like `libyaml-devel`). No ruby-module-enable step was needed for EL10 since it ships Ruby ≥3.2 by default and dropped DNF modularity.
  - Added an EL10-only fix for a podman/beaker-docker incompatibility discovered while validating this locally: podman's newer Docker-compat API (confirmed on podman 6.0.2, shipped by CentOS Stream 10) reports `HostConfig.NetworkMode` as the generic `"bridge"`, while the actual default network it attaches containers to is still named `"podman"`. beaker-docker 3.1.2 looks up `NetworkSettings.Networks[NetworkMode]` to get the container IP, so that key mismatch crashes with `undefined method '[]' for nil`. Renaming podman's default network to `"bridge"` (via a `containers.conf.d` drop-in) makes the two agree. Confirmed EL9/AlmaLinux9 production runs don't hit this (older podman doesn't have the mismatch), so it's gated to `distribution_major_version == "10"` only.

No changes were needed to `01-boxes.yml` — the `centos10-stream` box (`centos/stream10`) already exists in `vagrant/boxes.d/00-centos.yaml`, and `03-tests.yml`'s existing `beaker_os: "{{ pipeline_os.replace('-stream', '') }}"` correctly produces `centos10`, matching beaker-hostgenerator's `centos10-64` platform.

No change was needed for the Redis→Valkey test assertion — that's already fixed in puppet-pulpcore (`spec/acceptance/basic_spec.rb`, OS-conditional on valkey vs redis, from https://github.com/theforeman/puppet-pulpcore/pull/411/ https://github.com/theforeman/puppet-pulpcore/pull/412 and picked up automatically since `roles/beaker/tasks/main.yml` clones the module fresh from `theforeman/puppet-pulpcore` at `HEAD` by default.

Nightly is tested against the real `stagingyum.theforeman.org` staging repo for EL10 (no Copr staging override needed, unlike puppet-pulpcore's CI, since forklift's pipeline isn't gated the same way GitHub Actions is).

## Test plan

- [x] Ran `ansible-playbook pipelines/pulpcore.yml -e pipeline_version=nightly -e pipeline_os=centos10-stream -e pipeline_type=pulpcore -e pipeline_action=pulpcore -e expected_version=''` locally end-to-end (boxes → install → acceptance tests) — passed with `failed=0`.
- [x] Confirmed EL9 (`almalinux9`) production Jenkins runs are unaffected — the podman network fix and `BEAKER_PUPPET_COLLECTION` override are both gated to `distribution_major_version == "10"`.

### DIFF
--- a/roles/beaker/tasks/main.yml
+++ b/roles/beaker/tasks/main.yml
@@ -4,6 +4,22 @@
     name: podman-docker
     state: installed
 
+- name: Create containers.conf.d directory
+  file:
+    path: /etc/containers/containers.conf.d
+    state: directory
+    mode: "0755"
+  when: ansible_facts['distribution_major_version'] == "10"
+
+- name: Align podman's default network name with its reported NetworkMode
+  copy:
+    dest: /etc/containers/containers.conf.d/99-default-network.conf
+    content: |
+      [network]
+      default_network = "bridge"
+    mode: "0644"
+  when: ansible_facts['distribution_major_version'] == "10"
+
 - name: Start podman
   service:
     name: podman
@@ -28,7 +44,7 @@
 
 - name: enable CRB
   command: dnf config-manager --set-enabled crb
-  when: ansible_facts['distribution_major_version'] == "9"
+  when: ansible_facts['distribution_major_version'] in ["9", "10"]
 
 - name: Install Ruby
   package:
--- a/roles/beaker/defaults/main.yml
+++ b/roles/beaker/defaults/main.yml
@@ -3,6 +3,7 @@ beaker_base_environment:
   BEAKER_provision: "yes"
   BEAKER_setfile: "{{ beaker_os }}-64{hostname={{ beaker_os }}-64.example.com}"
   BEAKER_destroy: "no"
+  BEAKER_PUPPET_COLLECTION: "openvox8"
   PATH: "{{ ansible_facts['env'].PATH }}:/usr/local/bin"
 beaker_environment: {}
 beaker_puppet_module_path: "/src/{{ beaker_puppet_module }}"


pipelines/pulpcore/03-tests.yml
```diff
@@ -13,6 +13,9 @@
       vars:
         beaker_puppet_module: "puppet-pulpcore"
         beaker_os: "{{ pipeline_os.replace('-stream', '') }}"
-        beaker_environment:
+        beaker_environment: "{{ beaker_environment_base if ansible_facts['distribution_major_version'] != '10' else beaker_environment_base | combine(beaker_environment_el10) }}"
+        beaker_environment_base:
           BEAKER_FACTER_PULPCORE_BASEURL: "https://stagingyum.theforeman.org/pulpcore/{{ pipeline_version }}/el{{ ansible_facts['distribution_major_version'] }}/x86_64"
           BEAKER_FACTER_PULPCORE_VERSION: "{{ pipeline_version }}"
+        beaker_environment_el10:
+          BEAKER_PUPPET_COLLECTION: "openvox8"
```
ekohl: I think this should be the default everywhere because we don't support anything else right now. So instead of overriding this for EL10, can you see if we can modify other collections?

Odilhao: I think moving to `defaults/main.yml` is safe, I'll run locally EL9 nightly and 3.85 to see if they work as expected

roles/beaker/tasks/main.yml
```diff
@@ -9,6 +9,28 @@
...
+  when: ansible_facts['distribution_major_version'] == "10"
+
+- name: Align podman's default network name with its reported NetworkMode
+  copy:
+    dest: /etc/containers/containers.conf.d/99-default-network.conf
+    content: |
+      [network]
+      default_network = "bridge"
+    mode: "0644"
+  when: ansible_facts['distribution_major_version'] == "10"
```
ekohl, line 21: How about doing this before podman is started so you don't need to restart?

ekohl (commented):
Looks like this should say as the commit message "support EL 10 in Beaker role". Other than that, this looks ok to me.

roles/beaker/tasks/main.yml
Odilhao: Tried here locally in one devbox with CS10 and it worked, I'll fix with one commit that squash everything.

zjhuntin (approved):

PR #1982, opened 2026-09-03
enable fips on EL10 by using grubby directly

evgeni:
this is not officially supported (fips-mode-setup was removed in EL10, you can only officially enable FIPS mode at installation time), but good enough for our testing

### DIFF
--- a/roles/fips/tasks/main.yml
+++ b/roles/fips/tasks/main.yml
@@ -1,13 +1,20 @@
 ---
-- when: not ansible_fips
+- when: not ansible_facts['fips']
   block:
     - name: Install FIPS package
       package:
         name: crypto-policies-scripts
         state: present
 
-    - name: enable fips mode
+    - name: enable fips mode using fips-mode-setup
       command: /usr/bin/fips-mode-setup --enable
+      when:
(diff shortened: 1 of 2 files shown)


evgeni:
It worked quite fine in https://github.com/theforeman/foremanctl/pull/338

roles/fips/tasks/main.yml
```diff
@@ -1,13 +1,20 @@
...
-    - name: enable fips mode
+    - name: enable fips mode using fips-mode-setup
       command: /usr/bin/fips-mode-setup --enable
+      when:
+        - ansible_facts['distribution_major_version'] != '10'
+
+    - name: enable fips mode using grubby (unsupported)
+      shell: grubby --update-kernel=$(grubby --default-kernel) --args=fips=1
+      when:
+        - ansible_facts['distribution_major_version'] == '10'
```
Gauravtalreja1: I'd suggest using `int` for comparing and making it future proof for future RHELs
```suggestion
        - ansible_facts['distribution_major_version'] | int >= 10
```

ehelms (approved):

PR #1985, opened 2026-09-09
enable SSH pipelining and connection persistence

evgeni:
this should help against the occasional "Timeout (12s) waiting for privilege escalation prompt" on Debian 13 we've been seeing

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -12,3 +12,7 @@ remote_user = vagrant
 host_key_checking = False
 force_color = 1
 remote_tmp = /tmp/ansible-$USER
+
+[ssh_connection]
+pipelining = True
+ssh_args = -C -o ControlMaster=auto -o ControlPersist=180s


ansible.cfg
```diff
@@ -12,3 +12,7 @@ remote_user = vagrant
 host_key_checking = False
 force_color = 1
 remote_tmp = /tmp/ansible-$USER
+
+[ssh_connection]
+pipelining = True
+ssh_args = -o ControlMaster=auto -o ControlPersist=180s
```
ogajduse: Documented default is `"-C -o ControlMaster=auto -o ControlPersist=60s"`

https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/ssh_connection.html#parameter-ssh_args

I want to make sure that you intentionally dropped the `-C` compression.

evgeni: I was mostly copying https://github.com/theforeman/foremanctl/commit/dba6262b49bc75f156624dd5127ea6f8e458cf08

There is probably no big win in using compression, given the connections are local, but can add it back!

ogajduse (approved):

